PR #17, opened 2015-01-07
Fix async calls to console.log.

### LINKED ISSUE #15: REPL log output is not displayed when inside async function
Odd.
It seems the `console.log` is being ignored by the REPL output when we call it inside an async function.  
I took too long to figure out the code was working but just was not displaying the log in the REPL.
In the browser dev tools it works well, i mean, the log appear.

The code i was trying:
```javascript
function p1() {
  return new Promise(function(resolve, reject){
    setTimeout(function() {
      resolve('p1');
    }, 300);
  });
}

async function getVal() {
  var val1 = await p1();
  return val1;
}

(async function() {
  var bla = await getVal();
  console.log(bla); 
}());
```

screenshot from 2015-01-07 11 44 00

eventualbuddha:
Instead of monkey patching `console.log`, we create a console wrapper
and monkey patch that, passing it as an argument to the generated
function so it will always be in scope.

We keep track of the current console wrapper so that any async calls
from previous evaluations will no longer cause output.

Closes #15.

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -63,29 +63,43 @@
 
   var $errorReporter = $('.to5-repl-errors');
   var $consoleReporter = $('.to5-repl-console');
+  var capturingConsole;
 
   var evaluate = function(code) {
     var buffer = [];
     var error;
-    var prevConsoleLog = console.log;
+    var done = false;
 
-    console.log = function() {
+    function write(data) {
+      buffer.push(data);
+      if (done) flush();
+    }
+
+    function flush() {
+      $consoleReporter.text(buffer.join('\n'));
+    }
+
+    capturingConsole = Object.create(console);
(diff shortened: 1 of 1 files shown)


scripts/repl.js
```diff
@@ -63,29 +63,43 @@
...
   var evaluate = function(code) {
     var buffer = [];
     var error;
-    var prevConsoleLog = console.log;
+    var done = false;
+
+    function write(data) {
+      buffer.push(data);
+      if (done) flush();
+    }
```
thejameskyle: Do we want to forward `console.log` to the normal console as well?

eventualbuddha: I like that. Done in a9abfeb.

thejameskyle:
Thanks!

PR #22, opened 2015-01-08
Fix module import syntax

caspervonb:


sebmck:
Thanks! Looks like it also uses the old syntax over at [es6features](https://github.com/lukehoban/es6features) which is what the tour page is based on.

thejameskyle:
ping @LukeHoban

docs/tour.md
```diff
@@ -412,7 +412,7 @@ export default function(x) {
 ```
 ```js
 // app.js
-module math from "lib/mathplusplus";
+import math from "lib/mathplusplus";
```
lukehoban: I believe what is actually needed in this case is this:

```js
import * as math from "lib/mathplusplus"
```

Since the `math` variable is being used to access `pi` and `e` in the later code.

lukehoban:
I made a couple of changes to the examples in the [es6features](https://github.com/lukehoban/es6features) repo to fix this - see https://github.com/lukehoban/es6features/commit/137641e20dce1d8e594d61dc759213f420e2a9a3.

sebmck:
@lukehoban Great, thanks! Ported the changes over.

thejameskyle:
Thanks @lukehoban!

PR #11, opened 2015-01-06
Add make setup

caspervonb:

PR #23, opened 2015-01-08
Add shell invocation usage for make

caspervonb:
See #11


caspervonb:
@thejameskyle  `lib` is the default target, so invocation is simply just `make`.

sebmck:
Thanks!

PR #52, opened 2015-01-13
Corrected amd-module-id => amd-module-ids (plural) flag per 6to5/6to5#190

jayphelps:
The flag was changed to be plural in PR 6to5/6to5#190

### DIFF
--- a/docs/usage/modules.md
+++ b/docs/usage/modules.md
@@ -108,7 +108,7 @@ define(["exports", "foo"], function (exports, _foo) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


sebmck:
Thanks! The docs should probably be changed to reflect the new name of `--module-ids` since `--amd-module-ids` is only in there for backwards compatibility.

jayphelps:
@sebmck ahh okay. Thanks for the tip! I'm jumping in from esnext/traceur so just fixing things as I see them and get acquainted.

sebmck:
@jayphelps Awesome! Let me know if you need any help or run into any other issues.

PR #63, opened 2015-01-18
s/Github/GitHub/

mathiasbynens:


sebmck:
Bit pedantic, but I like it. Thanks!

PR #64, opened 2015-01-18
Clearer title for REPL page

JedWatson:
Just a suggestion.

I was browsing the site and thought REPL isn't the most obvious title for this page (accurate though it is).

Something along the lines of "Try it out" or "Live demo" would be more self explanatory, IMO.

Could also be moved to a separate section of the nav to stand out more, but that's a bigger change so wanted to get feedback on this first.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -68,7 +68,7 @@ navigation:
   url: /docs/compare/
 - title: Caveats
   url: /docs/caveats/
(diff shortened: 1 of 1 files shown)


sebmck:
Ping @thejameskyle

thejameskyle:
I like it

PR #68, opened 2015-01-19
Document Mallet operator

jridgewell:


docs/usage/playground.md
```diff
@@ -90,6 +90,46 @@ var obj = {};
 if (!Object.prototype.hasOwnProperty.call(obj, "x")) obj.x = 2;
 ```
 
+### Mallet assignment operator
+
+The mallet assignment operator allows you to lazily initialize a variable.
+It checks whether the variable is falsey, if it is then the right hand value is set.
+
+While simplisticly equivalent to `a = a || b`, it will never set an already truthy value.
```
thejameskyle: s/simplisticly/simplistically

docs/usage/playground.md
```diff
@@ -90,6 +90,46 @@ var obj = {};
 if (!Object.prototype.hasOwnProperty.call(obj, "x")) obj.x = 2;
 ```
 
+### Mallet assignment operator
+
+The mallet assignment operator allows you to lazily initialize a variable.
+It checks whether the variable is falsey, if it is then the right hand value is set.
```
thejameskyle: s/variable/value

thejameskyle:
Thanks @jridgewell, just a couple of things and this is good to go.

docs/usage/playground.md
jridgewell: That's the last time I write prose in the Github editor...

jridgewell:
Updated.

thejameskyle:
Sweet thanks!

PR #72, opened 2015-01-20
Change `print` to `console.log` in example code

Daniel15:
When ran in the context of a browser, `print()` will call `window.print()` which opens a print dialog. It's very confusing pasting example code into the REPL only to have it open several print dialogs :)

### DIFF
--- a/docs/learn-es6.md
+++ b/docs/learn-es6.md
@@ -261,7 +261,7 @@ for (var n of fibonacci) {
   // truncate the sequence at 1000
   if (n > 1000)
     break;
-  print(n);
+  console.log(n);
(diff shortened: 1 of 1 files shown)


sebmck:
The tour page is based on [es6features](https://github.com/lukehoban/es6features), I don't really think it's a good idea to deviate from it in terms of changes. @thejameskyle thoughts?

Daniel15:
In that case the example should also include a definition for the `print` function so people don't think there's a bug with the REPL where it tries to print the page.

thejameskyle:
Yeah I think this is a good change.

PR #75, opened 2015-01-20
Array destructuring now requires Symbol

goatslacker:
This changed in this commit: https://github.com/6to5/6to5/commit/c509d06bc2d74b58b8edf38597b13bc278896c71


sebmck:
Woops, forgot to update that. Thanks!

Issue #71, opened 2015-01-20
FR: Local storage for REPL.


hemanth:
Was planning to add this feature of saving `input.editor.getValue();` to localstore. 

If `query.code` found display it or else fetch and display the local store data.

What say @sebmck @thejameskyle ?

thejameskyle:
Possibly, I'll look into it

hemanth:
:+1: Eger to help :)

thejameskyle:
@hemanth Go ahead and take this if you want :)

hemanth:
@sparty02 Shall I cheery pick the changes?

sparty02:
@hemanth Go for it!

hemanth:
@sparty02 @thejameskyle @sebmck Done!

hemanth:
This got in. :bowtie:

Issue #66, opened 2015-01-19
REPL: Improve REPL State Management


sparty02:
Currently, the REPL saves its state as an encoded URI in the address bar URL.  This has several drawbacks including things like tricky escaping (i.e. #44) and unfriendly URLs (i.e. #162).

The goal of this enhancement is to leverage an external store for saving REPL state, if desired.  Some requirements:
- URLs for saved REPL sessions should be friendly/short, for easy sharing.
- Persistence should be opt-in (i.e. every REPL session doesn't need to be saved somewhere)
- The persistence of state should be throttled, either by debouncing input or by explicitly asking the user to click a button to 'Save' the REPL state, useful for sharing or returning to at a later time.
- Consider how this ties into local storage persistence (implemented via #77).

_Non requirements/goals_:
- Private persistence (i.e. only certain users can access REPL sessions).  That is way overboard right now.

Some ideas:
- Use the GitHub Gists API to persist REPL input.
- Use the JS Bin API to persist REPL input.

hemanth:
Related [issue](https://github.com/6to5/6to5/issues/423)

sparty02:
relates to #44

sparty02:
@remy I found http://jsbin.com/help/api and http://jsbin.com/help/ajax-requests.  Just want to make sure that JS Bin encourages/tolerates use of it's API for JS persistence.  Or is it more of a 'Pro' option?

sparty02:
Lets stay away from using JS Bin as a storage solution (at least for free).  I think it would be an unnecessary/unfair burden to them.

thejameskyle:
@sparty02 this is currently being worked on: https://github.com/babel/babel.github.io/pull/195

Issue #59, opened 2015-01-15
REPL: Examples


thejameskyle:
Previous there was a dropdown where you could select a bunch of ES6 examples and insert them into the console, although I never finished it and removed it so I could release the new site.

sparty02:
@thejameskyle FYI - I've started implementing examples at 79edbbc8e0fb28d3e5972df5b2920896d71c5ee2

hemanth:
Cool, was noticing few at [here](http://es6fiddle.net/)...

hzoo:
@sparty02 Would you want to make a PR?

hzoo:
@sparty02 I'l probably just make a PR based off your branch - actually would be cool if the data for the invidual plugins pages/learn es2015 page/this were the same though

razat249:
@hzoo I am Rajat Patwa @razat249. I am interested in the GSoC problem REPL dropdown for es6 examples.I want to know how this issue is related to that problem. Are there any other issues related to the REPL project?

DeadlyHollows:
@**hzoo** I am Shivam Bathla @**DeadlyHollows**. Is the issue still open?

xtuc:
@DamianMullins Yes, looks like this issue has no PR yet.

Since this issue has been opened we added "Try in REPL" button on every JavaScript code block. Which includes the [Lean ES2015 page](https://babeljs.io/learn-es2015/) where you can find many examples.

anubhav7495:
I have opened a PR with a initial implementation. Would be grateful if somebody could review it and provide feedbacks.

PR #77, opened 2015-01-21
localStorage support.

hemanth:
@thejameskyle First cut. Let me know if this approach is OK with you. WRT #71

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -107,11 +107,12 @@
    */
   function REPL () {
     var state = UriUtils.parseQuery();
-    this.options = _.assign(new Options(), state);
-
+    
     if (window.localStorage) {
         state = JSON.parse(localStorage.getItem('replState'));
     }
+
+    this.options = _.assign(new Options(), state);
 
     this.input = new Editor('.to5-repl-input .ace_editor').editor;
     this.input.setValue(UriUtils.decode(state.code || ''));
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -111,9 +111,7 @@
     if (window.localStorage) {
         try {
           state = JSON.parse(localStorage.getItem('replState'));
-        } catch(err) {
-          console.error("Can't fetch repl state from local storage.");
-        }
+        } catch(e) {}
     }
 
     this.options = _.assign(new Options(), state);
@@ -227,9 +225,7 @@
     if (window.localStorage) {
         try {
           window.localStorage.setItem('replState', JSON.stringify(state));
-        } catch(err) {
-          console.error("Can't set repl state to local storage.");
-        }
+        } catch(e) {}
     }
   };
 
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -109,6 +109,10 @@
     var state = UriUtils.parseQuery();
     this.options = _.assign(new Options(), state);
 
+    if (window.localStorage) {
+        state = JSON.parse(localStorage.getItem('replState'));
+    }
+
     this.input = new Editor('.to5-repl-input .ace_editor').editor;
     this.input.setValue(UriUtils.decode(state.code || ''));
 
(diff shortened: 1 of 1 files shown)


sebmck:
A check for the existence of localStorage would be good to support browsers that don't have it.

hemanth:
@sebmck Was wondering the [same](http://caniuse.com/#search=localstorage), looks like only Opera mini doesn't have it, yet!

hemanth:
@sebmck Added a `MayBe` construct, what say ;) ?

sebmck:
Bit overkill. I was suggesting more `if (window.localStorage) localStorage.setItem("foo", "bar");`. Your solution wont work either because the `localStorage` reference will throw a `ReferenceError` because it's not declared.

hemanth:
@sebmck  Duh! yeah, I was overwhelmed by mayBe, heh heh, reverting it to a simpler version.

hemanth:
@sebmck Done :cat:

thejameskyle:
@hemanth Could you actually clean up these commits by rebasing your commit onto the latest master?

sparty02:
@hemanth I've found [this wiki](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) to be useful in respect to rebasing pull requests , and squashing if need be

hemanth:
@sparty02 Thanks, i'm aware of it, I fetched and merged your branch, hence the older commits have appeared.

hemanth:
@thejameskyle Done.

scripts/repl.js
```diff
@@ -109,6 +109,10 @@
     var state = UriUtils.parseQuery();
     this.options = _.assign(new Options(), state);
 
+    if (window.localStorage) {
```
sparty02: You may want to move the local storage call up above where options are set on the repl instance (so that options can be persisted too).

hemanth: :+1:

sparty02:
Local storage is an interesting idea, but I wonder if it would be confusing at all if a user came back to the site and there was all of a sudden input (via local storage) in their REPL.  I definitely think persisting options is nice, I'm just concerned about how to effectively persist the code.

I'm almost wondering if this should be beefed up into some kind of "save session for later" feature (which could key into local storage by ID).  This could then be pulled up via a menu.

Note, if it was taken this far, this could tie into #59 (via how to seed the REPL input from a dropdown) and #66 (the idea of 'saving' a session for later in a more robust way)

Just some thoughts......

hemanth:
@sparty02 I would vote to get this rolling first and improvise later? 

I have come across few online REPLs that save the previous code as is and is not confusing, maybe we must roll and wait for feedback from users?

gaearon:
Note `localStorage.setItem` and `localStorage.getItem` throw on iOS in Private Mode.
Not that it's very important to support REPL on iOS, but not nice to crash either.

thejameskyle:
> Note localStorage.setItem and localStorage.getItem throw on iOS in Private Mode.
> Not that it's very important to support REPL on iOS, but not nice to crash either.

It's probably fine to just wrap these in try-catch blocks. @hemanth mind doing that?

gaearon:
>It's probably fine to just wrap these in try-catch blocks. 

This is what I usually do, yeah.

hemanth:
@thejameskyle No other options left, I had created a MayBe monad :D which was an overkill :)

thejameskyle:
I would just leave it as a try-catch

```js
try {
  state = JSON.parse(localStorage.getItem('replState'));
} catch(e) {}
```

hemanth:
@thejameskyle Done.

hemanth:
@thejameskyle @sebmck :cat:

thejameskyle:
Thanks!

hemanth:
Thank you!

PR #78, opened 2015-01-21
Cache bust resources via appended timestamp, fixes #6

### LINKED ISSUE #6: Add cache busting
`?t=timestamp` at the end of all the resources.

sparty02:
@thejameskyle Let me know if this is roughly what you had in mind and/or if there are other resources that need to be considered.  Note that I used the date_to_xmlschema filter because it puts the timestamp in ISO8601 format, which should be URL safe.  The site.time Jekyll site variable should reflect when the Github Pages server spins up a new site after a publish.

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -8,7 +8,7 @@
 
   <meta name="viewport" content="width=device-width, initial-scale=1">
 
-  <link rel="stylesheet" href="{{ "/css/main.css" | prepend: site.baseurl }}">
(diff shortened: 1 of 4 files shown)


thejameskyle:
Mind adding it to the css in `_includes/head.html`?

sparty02:
@thejameskyle Done, plus I added it to the docs template (for docs.js)

thejameskyle:
Cool, thank you.

PR #81, opened 2015-01-21
favicon, closes #16

### LINKED ISSUE #16: favicon
Needs to have a favicon, as this isn't great:

2015-01-07 at 7 54 am

sparty02:


_includes/head.html
```diff
@@ -12,6 +12,9 @@
   <link rel="canonical" href="{{ page.url | replace:'index.html','' | prepend: site.baseurl | prepend: site.url }}">
   <link rel="alternate" type="application/rss+xml" title="{{ site.title }}" href="{{ "/feed.xml" | prepend: site.baseurl | prepend: site.url }}" />
 
+  <link rel="shortcut icon" href="{{ "/images/favicon.ico" | prepend: site.baseurl }}" type="image/x-icon">
+  <link rel="icon" href="{{ "/images/favicon.ico" | prepend: site.baseurl }}" type="image/x-icon">
```
thejameskyle: Why both of these?

sparty02: That's a great question.  I did some research on favicons and there is an obnoxious variety of ways you can import/link these things.  It reminds me of vendor-prefixed css...  :cry:   (see http://realfavicongenerator.net/ as an example).  Maybe the thing to do is to just use:
```
<link rel="icon"
```
and then layer on others if/when the need arises.

sparty02: On a related note, I couldn't figure out how to tell Jekyll to pick up files off the root ('they' recommend putting favicon.ico in the root of the site).....which is why I put it in images.  @thejameskyle do you know?

thejameskyle: @sparty02 you should just be able to put it in the root of the directory.

sparty02:
@thejameskyle I think this should be ready to go now (put favicon in root and cleaned up multiple links in head)

thejameskyle:
Sweet, thanks!

Issue #44, opened 2015-01-12
REPL: Better url escaping


thejameskyle:
`encodeUriComponent` doesn't escape parenthesis (and other characters) so people adding links to markdown end up giving it an incorrect link:

http://6to5.org/repl/#?experimental=true&playground=true&evaluate=true&code=%20%20%20%20function%20*traverse(value)%20%7B%0A%20%20%20%20%20%20if%20(typeof%20value%20

PR #82, opened 2015-01-21
Fix code url on REPL on Firefox

sparty02:


thejameskyle:
@sparty02 Mind giving the commit a better description? Makes it easier to follow.

sparty02:
@thejameskyle Sound better?

thejameskyle:
heh, you don't have to be _so_ verbose, but thanks

sparty02:
@thejameskyle LOL, I really wasn't trying to be snippy about it.... just getting used to contributing more....

thejameskyle:
This actually doesn't appear to have fixed every issue in firefox.

thejameskyle:
@sparty02 It's cool :smile:

sparty02:
@thejameskyle I was actually thinking about this issue recently (in addition to how it relates to #44 that you opened) and I'm wondering if persisting the REPL input to the query fragment is the best approach.  This is kinda why I abstracted out the 'persistState' method in the REPL object so that it would be easier to reason about later.

One thought I had was to just base64 encode the whole thing into a string (vs doing the encodeUriComponent".  I dunno....

thejameskyle:
@sparty02 I would be okay with base64-ing the code, but I think the query fragment works well otherwise.

We'll have to include a base64 implementation since btoa/atob aren't supported < IE10.

sparty02:
@thejameskyle agreed....I'll take a look at this, probably tomorrow.

thejameskyle:
Sure, thank you so much for the work you've been putting into this!

PR #89, opened 2015-01-23
fix null state error on initial visit

sparty02:
@thejameskyle This is a hotfix to the recent localstorage enhancement.  There is an issue when you first come into the REPL with a browser that supports local storage (i.e. it sets the state to null, then state.code throws).


sebmck:
Thanks!

PR #93, opened 2015-01-24
Storage abstraction

sparty02:
@thejameskyle @hemanth  Looking for another pair of :eyes:   This is a simple abstraction to separate concerns from the REPL (not having to know as much about the storage implementation/etc.

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -116,27 +116,9 @@
   }
 
   /*
-   * Soft merge assigns non-undefined values in a source 
-   * object into the destination object.  However, objects are only
-   * assigned if they already exist in the destination object (
-   * otherwise they are dropped);
-   */
-  function softMerge (dest, source) {
-    return Object.keys(dest).reduce(function (dest, key) {
-      var sourceVal = source[key];
-      
-      if (!_.isUndefined(sourceVal)) {
-        dest[key] = sourceVal;
-      }
-      
-      return dest;
-    }, dest);
-  }
-
-  /*
    * 6to5 options for transpilation as used by the REPL
    */
-  function Options (initial) {
+  function Options () {
     var $experimental = $('#option-experimental');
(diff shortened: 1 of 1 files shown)


hemanth:
Cool :+1: I would avoid `new` and use a `IIFE` instead.

scripts/repl.js
```diff
@@ -94,31 +132,29 @@
...
-        try {
-          var storedState = localStorage.getItem('replState');
-          if (storedState) {
-            state = JSON.parse(storedState);  
-          }
-        } catch(e) {}
+  function REPL (storage) {
+    var state = {};
+    if (storage) {
+      _.assign(state, storage.get('replState'));
```
thejameskyle: I think it'd be better if the REPL just created its own storage. Something like:

```js
function REPL () {
  this.storage = new StorageService();
  var state = storage.get('replState') || {};
```

scripts/repl.js
```diff
@@ -227,17 +263,14 @@
 
   REPL.prototype.persistState = function (state) {
     UriUtils.updateQuery(state);
-    if (window.localStorage) {
-        try {
-          window.localStorage.setItem('replState', JSON.stringify(state));
-        } catch(e) {}
-    }
+    storage.set('replState', state);
```
thejameskyle: This is currently depending on the global `storage` object

scripts/repl.js
hemanth: You meant, `this.storage.get` or `var storage` ?

scripts/repl.js
sparty02: Agreed, this should be this.storage

scripts/repl.js
sparty02: Yep, should be this.storage

hemanth: :+1:

hemanth:
:+1:

PR #95, opened 2015-01-25
add nodemon argument swallowing information

cesarandreu:
Spent a while struggling with this last night.
Wrote a comment on an issue [here](https://github.com/6to5/6to5/issues/351#issuecomment-71346319).

### DIFF
--- a/docs/using-6to5.md
+++ b/docs/using-6to5.md
@@ -754,6 +754,6 @@ $ nodemon --exec npm run 6to5-node -- path/to/script.js
 Calling nodemon with 6to5-node may lead to arguments getting parsed incorrectly if you forget to use a double dash. Using npm-scripts helpers prevent this. If you chose to skip using npm-scripts, it can be expressed as:
 
 ```sh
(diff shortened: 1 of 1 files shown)


docs/using-6to5.md
```diff
@@ -737,6 +737,23 @@ $ npm install --global 6to5
...
+
+```sh
+$ nodemon --exec npm run 6to5-node -- path/to/script.js
+```
+
+Calling nodemon with 6to5-node may lead to arguments getting parsed incorrectly if you forget to use a double dash. Using npm-scripts helpers prevent this. If you chose to skip using npm-scripts, it can be expressed as:
+
 ```sh
-$ nodemon --exec 6to5-node --experimental path/to/script.js
+nodemon --exec 6to5-node --experimental --cache --ignore='foo\|bar\|baz' -- path/to/script.js
```
sebmck: Can you add the `$ ` back here?

cesarandreu:
@sebmck Done. Sorry, didn't catch that.

sebmck:
Great, thanks!

PR #124, opened 2015-02-03
Fix markdown syntaxes

okuryu:
* Use shell script syntax for commands
* Use dollar sign for commands


sebmck:
Wonder how those `js` ones snuck in there. Thanks!

okuryu:
Too fast, thanks! :cat2:

PR #128, opened 2015-02-08
Move 6to5-runtime box to after the selfContained header

AluisioASG:
Otherwise a link to https://6to5.org/docs/usage/transformers/#self-contained won't show it.


sebmck:
Whoops, thanks!

PR #130, opened 2015-02-10
Fix invalid link address in the doc/usage/require

SFantasy:
The origin code in the repo would create this link address: http://6to5.org/docs/usage/require/polyfill , and it's 404 page.

So the code in the `[]` should be `/docs/usage/polyfill` or `./polyfill`

### DIFF
--- a/docs/usage/require.md
+++ b/docs/usage/require.md
@@ -30,8 +30,7 @@ require("6to5/register");
 ```
 
 All subsequent files required by node with the extensions `.es6`, `.es`, `.jsx`
-and `.js` will be transformed by 6to5. The polyfill specified in
-[/docs/usage/polyfill](/docs/usage/polyfill) is also automatically required.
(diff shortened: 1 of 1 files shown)


docs/usage/require.md
```diff
@@ -31,7 +31,7 @@ require("6to5/register");
 
 All subsequent files required by node with the extensions `.es6`, `.es`, `.jsx`
 and `.js` will be transformed by 6to5. The polyfill specified in
-[/docs/usage/polyfill](polyfill) is also automatically required.
+[/docs/usage/polyfill](/docs/usage/polyfill) is also automatically required.
```
thejameskyle: could you change this to just:

> The [polyfill]() is also automatically required.

SFantasy: @thejameskyle sure, I will create another Pull Request.

thejameskyle:
:+1:

PR #134, opened 2015-02-15
Update external helpers doc with global injection

goatslacker:
This adds documentation for injecting the `babelHelpers` into window and global.


sebmck:
Looks good! The `require("6to5/runtime");` usage wasn't really documented in 3.x so I think it's fine to leave that out. Can you also add a reference to including `external-helpers.js` from an npm release on your page in a `<script>` tag and it'll work just the same?

goatslacker:
Ok cool this is good.

sebmck:
Great, thank you!

PR #142, opened 2015-02-18
Error messages on access to exports/module/require

hemanth:
As per @sebmck's [suggestion](https://gitter.im/babel/babel?at=54e40cde48b72ed530955d28).

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -4,8 +4,8 @@
   ["module", "exports", "require"].forEach(function(commonVar){
     Object.defineProperty(window, commonVar, { 
       get: function () { 
-      throw new Error(commonVar + " aren’t supported in the browser,\
-        you need a commonjs environment such as node/io, browserify/webpack etc");
+      throw new Error(commonVar + " is not supported in the browser,\
+        you need a commonjs environment such as node.js/io.js, browserify/webpack etc");
       }
     });
   });
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -1,5 +1,15 @@
 (function(babel, $, _, ace, window) {
 
+  /* Throw meaningful errors for getters of commonjs. */
+  ["module", "exports", "require"].forEach(function(commonVar){
+    Object.defineProperty(window, commonVar, { 
+      get: function () { 
+      throw new Error(commonVar + " aren’t supported in the browser,\
+        you need a commonjs environment such as node/io, browserify/webpack etc");
+      }
+    });
+  });
+  
   /*
    * Utils for working with the browser's URI (e.g. the query params)
    */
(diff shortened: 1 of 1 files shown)


sebmck:
:+1: Would be a better idea to use a `var commonVars = ["module", "exports", "require"];` and loop over it instead of duplicating it.

hemanth:
@sebmck Yup, was thinking the same, editing on the fly ;)

scripts/repl.js
```diff
@@ -1,5 +1,15 @@
 (function(babel, $, _, ace, window) {
 
+  /* Throw meaningful errors for getters of commonjs. */
+  ["module", "exports", "require"].forEach(function(commonVar){
+    Object.defineProperty(window, commonVar, { 
+      get: function () { 
+      throw new Error(commonVar + " aren’t supported in the browser,\
```
thejameskyle: s/aren't/isn't

scripts/repl.js
```diff
@@ -1,5 +1,15 @@
...
+        you need a commonjs environment such as node/io, browserify/webpack etc");
```
thejameskyle: node.js/io.js*

hemanth:
@thejameskyle @sebmck done.

scripts/repl.js
```diff
@@ -1,5 +1,15 @@
 (function(babel, $, _, ace, window) {
 
+  /* Throw meaningful errors for getters of commonjs. */
+  ["module", "exports", "require"].forEach(function(commonVar){
+    Object.defineProperty(window, commonVar, { 
+      get: function () { 
+      throw new Error(commonVar + " is not supported in the browser,\
+        you need a commonjs environment such as node.js/io.js, browserify/webpack etc");
```
thejameskyle: Whoops, this line break causes the log to have a lot of extra spaces. @hemanth mind opening a new PR to remove this whitespace?

hemanth: @thejameskyle Oh, it looked fine in the console, do you want the entire message in a single line?  As in with out the `\` 

Oh man it would awesome if we rewrite this in ES6, we could have used template string ;)

thejameskyle: Looks like this for me.

screen shot 2015-02-19 at 10 22 30 pm

I've considered writing it in ES6, but its a small codebase and I don't feel like adding a build step (makes it super easy to just merge) :stuck_out_tongue_winking_eye:

sebmck: @thejameskyle Just add `type="text/babel"` to the script tag and it should work since the browser build of babel is included.

thejameskyle: That's only included on the repl page and it would break CloudFlare's auto-minification I have setup right now. Come fix that when you start at CloudFlare and we'll talk :stuck_out_tongue_winking_eye:

hemanth: heh heh, ok!

New PR sent.

PR #159, opened 2015-02-24
api: Add information about babel-core

TimothyGu:


sebmck:
Thanks @TimothyGu, appreciated!

PR #160, opened 2015-02-25
Add note about the runtime not supporting AMD

patrickarlt:
Looks like this has been asked for several times by me https://github.com/babel/babel/issues/895 and previously in https://github.com/babel/babel/issues/638. Adding a note about the runtime not supporting AMD might help reduce future duplicates of this.


sebmck:
Thanks!

PR #166, opened 2015-03-02
fix the result of `utility.inlineExpressions`

azu:
transformers babel 2015-03-02 16-04-25

- http://babeljs.io/docs/usage/transformers/#utility-inline-expressions

This pull request fixes the result of `utility.inlineExpressions`.

```
$ babel -V
4.6.4
$ cat inline.js
var foo = 5 * 5;
$ babel --optional utility.inlineExpressions inline.js
"use strict";

var foo = 25;
```

### DIFF
--- a/docs/usage/transformers.md
+++ b/docs/usage/transformers.md
@@ -219,7 +219,7 @@ var foo = 5 * 5;
 becomes
 
 ```javascript
-var foo = 10;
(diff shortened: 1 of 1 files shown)


sebmck:
Holy shit wow how did I miss that... Thanks @azu!

PR #173, opened 2015-03-07
fucnsion -> function

benjamingr:
Repl isn't working at the moment because of a SyntaxError


sebmck:
FML, thanks @benjamingr!

PR #175, opened 2015-03-07
Replace JSON.stringify with better pretty-printer. Fixes #172

### LINKED ISSUE #172: Better inspect/format
Via @getify babel/babel#968

> This code doesn't appear to output correctly in babel's REPL:
> 
> ```js
> var s = Symbol("foo");
> 
> var x = {
>   [s]: 42
> };
> 
> console.log( Object.getOwnPropertySymbols(x) );
> // [ null ]
> ```
> 
> Apparently it's because the output is using `JSON.stringify(..)`, which of course cannot represent symbols.
> 
> Can we get an option (another lovely checkbox) to have `console.log(..)` output shown in the REPL raw without this transformation?

thejameskyle:
I couldn't find a good library for this and node/iojs's `util.format` wasn't enough, I should probably turn this into a separate library with tests.

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -2,13 +2,103 @@
 
   /* Throw meaningful errors for getters of commonjs. */
   ["module", "exports", "require"].forEach(function(commonVar){
-    Object.defineProperty(window, commonVar, { 
-      get: function () { 
+    Object.defineProperty(window, commonVar, {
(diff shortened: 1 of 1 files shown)


thejameskyle:
Just published https://www.npmjs.com/package/pretty-format

PR #181, opened 2015-03-15
add docs for double dash and new babel-external-helpers options

neVERberleRfellerER:
Formats section was more detailed but I have decided to shorten it to minimum since everybody can simply type it into console and check the output. Sorry that it took so long.


sebmck:
Thanks!

PR #195, opened 2015-03-22
Add repl short url support for #66

brigand:
Edit: parse.com is shutting down within the year, so this PR doesn't make sense anymore.

**WIP**
 - [x] functionality 
 - [x] styling (useable, I'm not a designer)
 - [ ] update parse client id/key

This adds the ability to use 'short' urls with the repl.  Really they're not very short, but they are unique, persistent, and the length isn't dependant on the input.  The implementation uses Parse.com, but it could be swapped out with gists or similar in the future.

Example URL:

```
/repl/#?short=dOWCPHsLgJ
```

I'm opening this before it's done so @sebmck can create an application in parse.com and give the client id/key.  Once one is saved with that id, you need to go into the data viewer and change the Class Level Permissions for ReplItem... they're very very permissive by default (like having direct database access on the client, permissive).

e.g. if you named it 'babel-repl': https://parse.com/apps/babel-repl/collections#class/ReplItem

Class Level Permissions for ReplItem in parse.com UI

And then in settings on the top nav bar, switch it into production mode and disable client side class creation.

parse.com application settings screen

After that, it'll just work and we'll never hit the free limit.

#66

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -420,14 +420,26 @@
       code: repl.getSource()
     });
 
+    repl.$shortenUrlButton.prop('disabled', true);
+    repl.$shortenUrlButton.css({opacity: 0.2});
+
     repl.shortener.save(state, function (err, id){
+      repl.$shortenUrlButton.prop('disabled', false);
+      repl.$shortenUrlButton.css({opacity: 1});
+
       if (err) {
         console.error('It seems saving isn\'t working for you.  Please report it here: https://github.com/babel/babel.github.io/issues/new');
         console.error(err);
         alert('Could not shorten url, see developer console console.');
         return;
       }
+
       UriUtils.updateQuery({short: id});
+
+      repl.$shortenUrlButton.css({color: '#70ff70'});
+      setTimeout(function (){
+        repl.$shortenUrlButton.css({color: ''});
+      }, 1500);
     });
   }
 
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -1,9 +1,14 @@
 (function(babel, $, _, ace, window) {
 
+  // we load it async, and it's a UMD so it checks for module/exports
+  // it would require automation or dialup internet for this to potentially affect a user
+  var hackToAllowCommonVarWhileLoadingParse = false;
+
   /* Throw meaningful errors for getters of commonjs. */
   ["module", "exports", "require"].forEach(function(commonVar){
     Object.defineProperty(window, commonVar, { 
-      get: function () { 
+      get: function () {
+        if (hackToAllowCommonVarWhileLoadingParse) return undefined;
         throw new Error(commonVar + " is not supported in the browser, you need a commonjs environment such as node.js/io.js, browserify/webpack etc");
       }
     });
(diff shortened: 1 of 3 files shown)


scripts/repl.js
```diff
@@ -57,6 +62,86 @@
   };
 
   /*
+   * Utils for working with parse.com for persistence
+   */
+  function ParseUtils () {}
+
+  var __parseLoadingCallbacks = null;
```
thejameskyle: Is there a reason this can't be added as an instance property of ParseUtils?

scripts/repl.js
```diff
@@ -57,6 +62,86 @@
...
+    this.afterLoad(function(ReplItem){
+      var query = new Parse.Query(ReplItem);
+      query.get(id)
+        .then(function (data) {
+        // give us a plain object without anything parse specific, making
+        // .get's output the same as .save's input
+        var cleanData = data.toJSON();
+        delete cleanData.objectId;
+        delete cleanData.createdAt;
+        delete cleanData.updatedAt;
```
thejameskyle: It's probably fine just to let this fall through.

scripts/repl.js
```diff
@@ -284,7 +371,34 @@
   /*
    * Initialize the REPL
    */
-  var repl = new REPL();
+  var initialQuery = UriUtils.parseQuery();
+  if (initialQuery && initialQuery.short) {
```
thejameskyle: @sebmck Do you think it'd make sense to require all of these to be shortened urls?

scripts/repl.js
```diff
@@ -301,10 +415,33 @@
...
+        alert('Could not shorten url, see developer console console.');
+        return;
+      }
+
+      UriUtils.updateQuery({short: id});
+
+      repl.$shortenUrlButton.css({color: '#70ff70'});
+      setTimeout(function (){
+        repl.$shortenUrlButton.css({color: ''});
+      }, 1500);
```
thejameskyle: What exactly is this timeout for?

brigand: Just the css transition.

scripts/repl.js
brigand: It doesn't require the url to be shortened, it's just handling the case where it is.  Otherwise it falls through to the existing behavior.

scripts/repl.js
brigand: I believe it'll end up in state, which will end up in the URL.  I'll confirm it, though.

scripts/repl.js
brigand: Not currently, but it'd be fragile.

```js
new ParseUtils().load().afterLoad(f);
new ParseUtils().load().afterLoad(g);
```

This loads parse once, both f and g are called when it's loaded.  Changing to to an instance property would load parse twice and maybe break things.  It could be made a property of ParseUtils if that's prefered.

brigand:
Thanks for the comments.  I missed the email notification, so sorry for the late response.

scripts/repl.js
brigand: Yeah, I ended up with: http://localhost:8080/repl/#?code=5%2B5%2B1&evaluate=true&experimental=true&loose=false&playground=false&spec=false&objectId=6kDKy2aEul&createdAt=2015-03-28T07%3A35%3A23.598Z&updatedAt=2015-03-28T07%3A35%3A23.598Z

scripts/repl.js
thejameskyle: I meant instead of the old query params (`?experimental=true&code=foo`)

scripts/repl.js
thejameskyle: Why not create a factory function that returns a singleton? Is there even a case where we'd have to `ParseUtils`?

scripts/repl.js
thejameskyle: Alright np

scripts/repl.js
thejameskyle: cool

scripts/repl.js
brigand: It's always constructed here:

```js
function REPL () {
  this.shortener = new ParseUtils();
```

And conditionally here:
```js
if (initialQuery && initialQuery.short) {
  new ParseUtils().get(initialQuery.short, function (err, data) {
```

It could be done in an iife to create a singleton if that's preferred.  I was just trying to stick to the existing style as much as possible.

scripts/repl.js
brigand: Meaning it should leave the url at /repl/ until they hit save, and only support linking with short urls?

thejameskyle: Yes, I'm leaning towards that being the right thing to do, but I also don't want to break existing URLs.

brigand: Okay, I'll try to do this later today.  It can load the old urls, and just have the url always be `#` or `#?short=...`.

thejameskyle: Cool! Keep it as a separate commit please. :smile:

brigand:
parse.com is shutting down within the year, so this PR doesn't make sense anymore.

PR #205, opened 2015-03-26
Iterators, Generators, and the Reflect API require the Babel polyfill

also:
* Iterators rely on `Symbol.iterator`
* Generators rely on Regenerator
* Reflect is just an API


sebmck:
Thanks!

PR #210, opened 2015-03-30
link to a couple good blog posts and the proposal.

lookfirst:


sebmck:
@thejameskyle is working a resources section which is probably a better place for the blog posts.

thejameskyle:
~~I'm working on a new "Resources" page...~~ god damn it @sebmck I was just typing this out when your comment popped up.

sebmck:
@thejameskyle Don't be so slow then :wink:

thejameskyle:
Anyways yeah, I already have these blog posts on the page if you want to pull those out we can merge just the part that includes the proposal.

lookfirst:
i dislike wading through a list of resources. i'd rather have the contextual information right where it is relevant. but whatever you want...

thejameskyle:
If we want to do this sort of thing in the future we can use jekyll to automatically manage placement of articles. Hardcoding them all over the place would get unwieldy.

PR #213, opened 2015-04-01
replace is with are in plural context

derwolfe:
I may be wrong, but this sentence reads incorrectly in my mind. Also - sorry for the annoying grammar nit PR.

Thanks for babel!


sebmck:
Good catch, thanks!

PR #222, opened 2015-04-03
fix references to es7.objectRestSpread

also:
`es7.objectSpread` was changed to `es7.objectRestSpread` in v3.3.10 (https://github.com/babel/babel/commit/39fe737cb619357d321517f925cc3a3cf0004356), although it was not included in the [CHANGELOG](https://github.com/babel/babel/blob/master/CHANGELOG-6to5.md#3310).


sebmck:
Thanks @also!

PR #229, opened 2015-04-07
Specify compilation and the babel-runtime for ES6 libraries

rajit:
When writing a library, it's clear a polyfill should be used and that the require hook automatically includes the polyfill. The warning that the require hook is not suitable for libraries offers no correct alternatives so this pull request updates the documentation on require hooks to point you to the recommend solution of compiling your library and depending on the babel-runtime.


sebmck:
Awesome, thanks!

PR #231, opened 2015-04-08
Update 2015-03-31-5.0.0.md

dariocravero:
Included some docs around a breaking change while following the ES6 spec on imports being hoisted.


sebmck:
Awesome, thank you @dariocravero!

PR #236, opened 2015-04-10
feat(docs/using-babel.md): add docs about nodemon

hiroqn:
better way to run nodemon


thejameskyle:
I believe it's considered a bad practice to hardcode the path like that. I don't think it needs to be in the docs either.

PR #253, opened 2015-04-25
Remove vertical tab acsii -� - from learn-es6.md

PrototypeAlex:
Markdown doesn't seem to like rendering vertical tabs, as shown below:

screen shot 2015-04-26 at 8 58 18 am

This small change removes all `^K` from the markdown.


sebmck:
Thanks!

PR #256, opened 2015-04-27
Update learn-es6 to clarify 'JavaScript'

zcei:
Just replaced it by `ES5` as the doc title also uses the acronym - thought about adding `(commonly referred to as 'JavaScript')` but imho that's too much for these comments and should be treated as a prerequisite to learn ES6.


sebmck:
Thanks!

Issue #246, opened 2015-04-20
Clarify usage of "only" option


IanVS:
According to https://github.com/babel/babel/issues/1306#issuecomment-94341372 there are a lot of ways that the "only" option can be used:

> It can take an array of globs, a single regex, a single string glob or an array of regexs.

It would be great it the documentation could be a little more clear on this.  As-is, the documentation tells partial stories in different places.  The [options docs](https://babeljs.io/docs/usage/options/) currently state that the `only` option takes an array of globs:

> An array of glob paths to **only** compile. When attempting to compile a non-matching file it's returned verbatim.

But the [require usage](https://babeljs.io/docs/usage/require/) only mentions regex.  

> // Optional only regex - if any filenames don't match this regex then they
// aren't compiled
only: /my_es6_folder/,

greim:
The docs are indeed confusing; I've taken to using this really ugly hack until the day I finally understand how this thing works:

```js
var patt = /.*/
patt.test = function(path){
  var isMatch = /* do some logic here on "path" */
  return isMatch
}
require('babel').register({
  ignore: patt
})
```

PR #257, opened 2015-04-27
Clarifies "only" option

zcei:
I looked up how the `only` option is treated in code to address #246.

### [babel-cli/bin/babel/index.js](https://github.com/babel/babel/blob/67201e969880e0435d7b20200d1e0d2695b5f863/packages/babel-cli/bin/babel/index.js#L112): 

`opts.only = util.arrayify(opts.only, util.regexify);`

### [api/register/node.js](https://github.com/babel/babel/blob/master/src/babel/api/register/node.js#L147)

`if (opts.only != null) onlyRegex = util.regexify(opts.only);`

`util.regexify` epects:

* glob (this can also be an explicit path)
* rexexp
* array of strings to match **exactly** 

Thus the `register` method currently can't handle an array of glob/regexp.

Corner case is currently passing in an array of arrays of strings to the cli, as these are joined by logical or (`|`) and escaped by `lodash`'s `excapeRegExp` instead of `minimatch`'s `makeRe`.
(But how do you pass an array of arrays via command line?)

### DIFF
--- a/docs/usage/options.md
+++ b/docs/usage/options.md
@@ -27,7 +27,7 @@ $ babel --name=value
 | `optional`               | `[]`                 | Array of transformers to [optionally](/docs/usage/transformers#optional) use. Run `babel --help` to see a full list of transformers. Optional transformers displayed inside square brackets. |
 | `nonStandard`            | `true`               | Enable support for JSX and Flow. |
 | `highlightCode`          | `true`               | ANSI highlight syntax error code frames |
-| `only`                   | `null`               | An array of [glob](https://github.com/isaacs/minimatch) paths to **only** compile. When attempting to compile a non-matching file it's returned verbatim. |
+| `only`                   | `null`               | A [glob](https://github.com/isaacs/minimatch), regex, or mixed array of both, matching paths to **only** compile. Can also be an array of arrays containing paths to explicitly match. When attempting to compile a non-matching file it's returned verbatim. |
 | `ignore`                 | `null`               | Opposite to the `only` option. |
 | `jsxPragma`              | `null`               | Custom pragma to use for JSX elements |
 | `auxiliaryComment`       | `null`               | Attach a comment before all helper declarations and auxiliary code. eg. `"istanbul ignore next"` |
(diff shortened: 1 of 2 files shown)


docs/usage/options.md
```diff
@@ -27,7 +27,7 @@ $ babel --name=value
 | `optional`               | `[]`                 | Array of transformers to [optionally](/docs/usage/transformers#optional) use. Run `babel --help` to see a full list of transformers. Optional transformers displayed inside square brackets. |
 | `nonStandard`            | `true`               | Enable support for JSX and Flow. |
 | `highlightCode`          | `true`               | ANSI highlight syntax error code frames |
-| `only`                   | `null`               | An array of [glob](https://github.com/isaacs/minimatch) paths to **only** compile. When attempting to compile a non-matching file it's returned verbatim. |
+| `only`                   | `null`               | A [glob](https://github.com/isaacs/minimatch) / regex or mixed array of both, matching pathes to **only** compile. Can also be an array of arrays containing pathes to explicitly match. When attempting to compile a non-matching file it's returned verbatim. |
```
sebmck: pathes - > paths

zcei: Oops, fixed & squashed

docs/usage/options.md
```diff
@@ -27,7 +27,7 @@ $ babel --name=value
 | `optional`               | `[]`                 | Array of transformers to [optionally](/docs/usage/transformers#optional) use. Run `babel --help` to see a full list of transformers. Optional transformers displayed inside square brackets. |
 | `nonStandard`            | `true`               | Enable support for JSX and Flow. |
 | `highlightCode`          | `true`               | ANSI highlight syntax error code frames |
-| `only`                   | `null`               | An array of [glob](https://github.com/isaacs/minimatch) paths to **only** compile. When attempting to compile a non-matching file it's returned verbatim. |
+| `only`                   | `null`               | A [glob](https://github.com/isaacs/minimatch) / regex or mixed array of both, matching paths to **only** compile. Can also be an array of arrays containing pathes to explicitly match. When attempting to compile a non-matching file it's returned verbatim. |
```
thejameskyle: Can you change this to "A glob, regex, or mixed array of both, [...]" and fix "pathes" to "paths"?

docs/usage/options.md
zcei: Still not sure how I got it managed to write `pathes` all the time.
Thanks for the little language enhancement :+1:

thejameskyle:
Thanks

PR #270, opened 2015-05-05
Fix quotes in Gemfile version of sprockets & sprockets-es6 mentioning

sudodoki:
...to be more consistence with github.com/bbatsov/ruby-style-guide && Gemfile quotes coming with default rails project

### DIFF
--- a/_includes/tools/rails/usage.md
+++ b/_includes/tools/rails/usage.md
@@ -1,7 +1,7 @@
 ```rb
 # Gemfile
-gem "sprockets"
-gem "sprockets-es6"
+gem 'sprockets'
+gem 'sprockets-es6'
 ```
 
 ```rb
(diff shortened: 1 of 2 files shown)


sebmck:
Double quotes are also in **widespread** use and are arguably more/just as common as single. In fact it's what GitHub recommend in their [official style guide](https://github.com/styleguide/ruby). I'd rather maintain consistency with the other code snippets which also use double quotes, especially since there isn't a consensus amongst the Ruby community. Nothing stopping the developer from adapting the snippet to their projects code style :) Thanks for the suggestion!

sudodoki:
@sebmck not going to fight too much on this one, but we're talking about `Gemfile`s,  which are a bit different case because they won't have to many characters that need escaping inside gem names strings and version strings.
Github's styleguide motivation for using double quotes all the time was for specifically not think too much about escaping.
Rails' default Gemfile comes with single quotes, as well as examples on [bundler site](http://bundler.io/).

PR #288, opened 2015-05-14
Fix _interopRequire to match actual code

fishermand46:
The docs showed the body of _interopRequire as being
```javascript
var _interopRequire = function (obj) {
  return obj && (obj["default"] || obj);
};
```
But the code in [https://github.com/babel/babel/blob/v5.3.3/src/babel/transformation/templates/helper-interop-require.js](helper-interop-require.js) shows:
```javascript
function (obj) {
  return obj && obj.__esModule ? obj.default : obj;
}
```

### DIFF
--- a/docs/usage/modules.md
+++ b/docs/usage/modules.md
@@ -144,9 +144,9 @@ export function bar() {
 define(["exports", "foo"], function (exports, _foo) {
   "use strict";
 
-  var _interopRequire = function (obj) {
-    return obj && (obj["default"] || obj);
-  };
+  function _interopRequire(obj) {
+    return obj && obj.__esModule ? obj["default"] : obj;
+  }
 
   exports.bar = bar;
   var foo = _interopRequire(_foo);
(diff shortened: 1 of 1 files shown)


docs/usage/modules.md
```diff
@@ -144,9 +144,9 @@ export function bar() {
 define(["exports", "foo"], function (exports, _foo) {
   "use strict";
 
-  var _interopRequire = function (obj) {
-    return obj && (obj["default"] || obj);
-  };
+  function _interopRequire(obj) {
+    return obj && obj.__esModule ? obj.default : obj;
```
sebmck: `.default` should be `["default"]`.

fishermand46:
Fixed.

docs/usage/modules.md
fishermand46: Fixed.

thejameskyle:
Cool

PR #293, opened 2015-05-16
Correct size in Sets example.

geneticgrabbag:
Should `s.size` not resolve to `3` instead of `2`?


zloirock:
[Example is correct.](https://goo.gl/nWY7Lp)

hzoo:
It's correct since it's adding `hello` two times and Sets have unique values?

geneticgrabbag:
Aargh - you're so right. :)  Apologies for the mixup.

PR #297, opened 2015-05-20
Attempt to clean up website and make Babel's "vision" more clear.

sebmck:
**Not ready to be merged. Homepage still needs to be cleaned up/styled.**

 - Rename "Learn ES6" to "Learn ES2015"
 - Rename "Using Babel" to "Setup"
 - Split "Docs" into "Usage" and "Advanced"
 - Remove a lot of junk on the homepage and make it clear what Babel is.

Please review @thejameskyle.


sparty02:
I apologize in advance if this sounds troll-ish, but some quick feedback...not that you asked for it, so take it with a grain of salt....  :wink: 
* Referring to ES6 as ES2015 is like calling Kleenex - facial tissue, or calling a band-aid - an adhesive bandage, or calling a Q-tip - a cotton swab.  Seems that name is to engrained and calling it ES2015 will just be confusing.  I've seen (as I'm sure you have to) several TC39 members make similar comments.
* I know you've gotten flack for this before, but......JSX/React advertising.  It's the second main thing on the main page and just above the fold.  I know that support has been baked in since the beginning, but there seems to be a struggle in vision between ES compliance (i.e. all of the non TC39 experimental/playground features were removed) and React/JSX support (totally outside of the bounds of TC 39)....but I could be totally off base here.

At any rate, I love your work, just wanted to weigh in so you had some public opinion.

sebmck:
@sparty02 

> Referring to ES6 as ES2015 is like calling Kleenex - facial tissue, or calling a band-aid - an adhesive bandage, or calling a Q-tip - a cotton swab. Seems that name is too engrained in developer land and calling it ES2015 will just be confusing. I've seen (as I'm sure you have to) several TC39 members make similar comments.

No it's not. It's like Kleenex changing their name and then referring to it by their new name... Of course it's ingrained, it's because tooling like Babel has continued to use the ES6 term even though the TC39 have made the ES2015 name official. I'm taking no part in spreading misinformation.

> I know you've gotten flack for this before, but......JSX/React advertising.

Babel supports JSX and Flow and is likely going to become the default compiler for those standards and that community.

> I know that support has been baked in since the beginning, but there seems to be a struggle in vision between ES compliance (i.e. all of the non TC39 experimental/playground features were removed) and React/JSX support (totally outside of the bounds of TC 39)....but I could be totally off base here.

All of the non-TC39 playground features were removed because they had an extremely high chance of colliding (or in fact did, the `@` `this` shorthand would have collided with decorators) with future standards.

I don't think you're very familiar with Babel's "vision" then ("vision" is wrapped in quotes because it's a terrible word and makes it look way more poncey than it actually is). Babel is about encouraging the evolution of JavaScript. Standards compliance has **no bearing** on supporting additional syntax or features. You can be completely standards compliant but tack on your own propietary shit (JScript/IE for examples), hell, all browsers tack on their own experimental APIs, vendor prefixes are a prime example of this.

Apologies if this came off harsh, just frustrated with people making claims about Babel. This homepage change is a step in the right direction IMO in establishing the project long-term and what was discussed in babel/babel#568.

docs/advanced/caveats.md
```diff
@@ -67,4 +68,4 @@ isn't recommended.
 By default, when using modules with Babel a non-enumerable `__esModule` property
 is exported. This is done through the use of `Object.defineProperty` which is
 unsupported in IE8 and below. A workaround for this is to enable
-[loose mode - modules](/docs/usage/loose/#es6-modules).
+[loose mode - modules](/docs/advanced/loose/#es6-modules).
```
sparty02: es6 -> es2015

sparty02:
@sebmck  Honestly, sorry if I struck a nerve... but I do appreciate the response.  I guess just setting the ```nonStandard``` option to false is really what I'm looking for.... I just selfishly wish it was that way by default.

I hear what you are saying on the vision thing (I think I even made some comments in babel/babel#568 that are pretty close to where you ended up with the pluggable transforms).  I'm guessing there are two camps using Babel.  Those using it very pragmatically to use new standards prior to being implemented by browsers (like we are) and those using it to support non-standard extensions....and probably those doing both.  The latter is interesting and valuable for evolution, but not something I want in my targeted for production workflow.

Honestly, my only gripe is that it makes me feel uneasy when non-standard add-ons are in the core of an engine I depend on to do standards compliant transpiling (and turned on by default).  C'est la vie....I just need to get over it.

thejameskyle:
> I'm guessing there are two camps using Babel. Those using it very pragmatically to use new standards prior to being implemented by JS runtimes (like we are) and those using it to support non-standard extensions....and probably those doing both.

You are thinking about this all wrong. Babel is a platform for transpilation first, and a transpiler second. Plugins are the first step towards giving developers a much more powerful set of tools for building javascript tooling.

There is support for ES2015/ES2016/ES3146 for the same reason there is support for JSX/Flow: because this enables people to build modern tooling. The built-in transforms are just a default set of tools that run on this "platform".

This is why comparing Babel to TypeScript, Traceur, or CoffeeScript is rather strange. It is an entirely different beast, and if you ask biased-ole me, it's a much more powerful and compelling beast because of the fact that it's more than just a transpiler.

Also I made a comment on that same thread you mentioned explaining my thoughts https://github.com/babel/babel/issues/568#issuecomment-73654671.

docs/setup.md
```diff
@@ -2,8 +2,9 @@
 layout: page
 title: Using Babel
 description: How to use Babel with your tool of choice
-permalink: /docs/using-babel/
+permalink: /docs/setup/
 redirect_from:
+ - /docs/using-babel/
  - /plugins.html
  - /docs/setup/
```
thejameskyle: You created an infinite redirect loop here. `/docs/setup/ => /docs/setup/` just remove this line ^

thejameskyle:
Made a couple of updates.

This section needs a new image:

screen shot 2015-05-20 at 10 19 30 pm

docs/setup.md
sebmck: lol whoops

sebmck:
@thejameskyle Yeah, that was one of the things I alluded to with the frontpage being a bit wonky.

thejameskyle:
Any ideas of what it should be? I can work on something this evening.

sebmck:
No clue, maybe a blueprint?

thejameskyle:
Why'd you get rid of this section: 
screen shot 2015-05-24 at 3 43 05 pm

sebmck:
It's really markety and just adds visual clutter.

thejameskyle:
Merging

PR #318, opened 2015-05-29
Fix direction of "JS in, JS out" in homepage

pascience:
Who transpiles ES5 to ES6+ ? Not me :)


thejameskyle:
Well, in this case neither is _really_ representative of any version of JavaScript. Also this change was to align with [Babel's "vision"](https://github.com/babel/babel.github.io/pull/297) which is more than ES6. So I'm going to close this.

PR #368, opened 2015-07-13
Include browser.min.js, not babel.min.js.

eventualbuddha:
The babel repo seems to have changed the target dist file for minified browser builds a while ago.


thejameskyle:
Thanks

PR #369, opened 2015-07-13
Use bundler to specify the dependencies for building the site.

eventualbuddha:


thejameskyle:
Does it make sense to use this too? https://github.com/github/pages-gem

eventualbuddha:
Sure, whatever makes sense to use as a dependency that'll get the job done.
How do you build it locally?

thejameskyle:
I just have that gem installed globally

eventualbuddha:
What I mean is, how do you "develop" with this repo? Do you use the `github-pages` gem?

thejameskyle:
No, I run `jekyll serve`. I may have installed it separately, but `github-pages` should install it

eventualbuddha:
Okay. Mainly I wanted a consistent way for people to test out the site. Bundler is generally a good way to do that since you can specify versions. Either way I think we should add something to the README (or CONTRIBUTING) that explains how to get set up. Could be something as simple as:

---

### Setup

Install `jekyll` and the required plugins:

```
# You might need to use sudo.
gem install jekyll
gem install jekyll-redirect-from
```

Then run the server:

```
jekyll serve
```

---

Or, if we go with bundler:

---

### Setup

Install `jekyll` and the required plugins:

```
# You may need to install bundler first.
bundle install
```

Then run the server:

```
bundle exec jekyll serve
```

---

I'm slightly inclined toward the latter since we can help users automatically install dependencies rather than copy-pasting commands from a doc.

thejameskyle:
Hmm, I think either way I'd want it in the Makefile as `make install` and `make serve` which we can include in the documentation in which case the dependency on bundler makes less sense (pun intended).

eventualbuddha:
Okay, I added `make serve` which uses bundler to install the dependencies. The most users will have to do if they don't have bundler is run `[sudo] gem install bundler`.

thejameskyle:
Cool

PR #377, opened 2015-07-16
Fix gulp task sequence. Put concat() after babel()

mattmazzola:
The concat operation should come after the babel transpilation step as seen here:
https://www.npmjs.com/package/gulp-babel

If someone was copying this doc as a gulpfile they could get errors when using modules.


thejameskyle:
Thanks

PR #378, opened 2015-07-18
Fix alias-keys.json reference which no longer exist

marcioj:
While reading the docs I found that the `alias-keys.json` file no longer exist. I guess at some point this file was extracted and the alias definition was moved to each type definition over the `alias` option. So I updated the docs to point for the new location.


hzoo:
Yeah in https://github.com/babel/babel/commit/8d70ab5853b8e43a622d829a212b2d80b2abc435 :+1:

hzoo:
Sorry for the delay - that part of the docs was removed - similar stuff in https://github.com/thejameskyle/babel-plugin-handbook#babel-types though

PR #380, opened 2015-07-21
renamed to `es6.parameters`

chentsulin:
`res6.parameters.rest`, `es6.parameters.default` has been renamed to `es6.parameters`


thejameskyle:
thanks

PR #381, opened 2015-07-21
Fix require for polyfill

cody:
Webpack gave an error: Cannot resolve module 'babel/polyfill'


cody:
I just found out that both ways work. The exiting is better, therefore closing. Babel's npm is confusing.

PR #386, opened 2015-07-24
Add usage with Fly

drKraken:


piecioshka:
@sebmck or @thejameskyle please accept - bigger usage section will increase Babel popularity!

PR #389, opened 2015-07-28
Update polyfill doc according to npm package name

ghoullier:
Update polyfill page according according to npm package name babel-core

### DIFF
--- a/docs/usage/polyfill.md
+++ b/docs/usage/polyfill.md
@@ -12,7 +12,7 @@ permalink: /docs/usage/polyfill/
 </p>
 
 This will emulate a full ES6 environment. This polyfill is automatically loaded
-when using `babel-node` and `babel/register`.
+when using `babel-node` and `babel-core/register`.
 
 ## Usage in Node/Browserify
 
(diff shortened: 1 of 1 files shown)


thejameskyle:
I've never been quite sure how to break down the docs by `babel` and `babel-core`. @sebmck what do you think?

sebmck:
@thejameskyle Yeah, the distinction is really shit. I'm not sure how to distinguish between the two in a clear way in the docs. In the next major it might be worth retiring the `babel` package in favor of `babel-cli` and `babel-core`.

thejameskyle:
Maybe, I also feel like we should tell people to install babel as

```sh
$ npm install --save-dev babel babel-core
```

This dedupes the dependency when installed at once (`babel` depends on it's peer dep), which means that `babel` and one of the other tools like `gulp-babel` will depend on the same copy.

Then the docs can always refer to `babel-core` for JS usage and `babel` (or `babel-cli`) for CLI usage.

sebmck:
Should we really be encouraging that behaviour when npm 3 has tree flattening?

thejameskyle:
Well if people are `require('babel-core')`-ing then it should be an installed dep.

hzoo:
Changed to `babel-polyfill` so closing - thanks anyway!

PR #392, opened 2015-07-30
Clarify caveats: spreading Array doesn't require Array.from.

jmm:
EDIT: current branch: https://github.com/babel/babel.github.io/compare/master...jmm:caveats-spread2

The [Caveats](https://babeljs.io/docs/advanced/caveats/) page has a table of "certain features [that] require certain polyfills." One of the entries says that spread requires `Array.from`. That's not actually the case when spreading an `Array` (even a sparse array), and that seems to be intentional:

> [sebmck commented on Feb 11](https://github.com/babel/babel/issues/754#issuecomment-73873202)
> Not sure how to support this without forcing the polyfill on people who just want to spread/destructure arrays.

> [sebmck commented on Feb 11](https://github.com/babel/babel/issues/757#issuecomment-73939953)
> The point of special casing if it's an array is so the polyfill isn't required for array spread

If I'm reading this stuff right it seems straightforward to me that this should be documented. Why deliberately go to the trouble to make that work only to tell users it won't :)? Currently it says that `Array.from` is required for spread *period*. I read this like any other API documentation: if some usage of spread works without `Array.from`, it's just a side effect of implementation details and subject to not work in any other case or stop working at any time (any release or maybe even when updating deps).

Note: I haven't been able to build the site, so I haven't previewed these changes.

### DIFF
--- a/docs/advanced/caveats.md
+++ b/docs/advanced/caveats.md
@@ -23,7 +23,9 @@ You may alternatively selectively include what you need:
 | Async functions, Generators | [regenerator runtime](https://github.com/facebook/regenerator/blob/master/runtime.js) |
 | Comprehensions              | `Array.from`                                                                          |
 | For Of                      | `Symbol`, `prototype[Symbol.iterator]`                                                |
-| Spread                      | `Array.from`                                                                          |
+| Spread (on non-`Array`)     | `Array.from`                                                                          |
+
+Spreading an actual `Array`, including sparse arrays, does not require `Array.from`.
 
 ## Classes
 


thejameskyle:
> Note: I haven't been able to build the site, so I haven't previewed these changes.

Are you not able to run `make serve` and have it work?

jmm:
> Are you not able to run `make serve` and have it work?

Oh, sorry, I assumed the build process was the same as the last time I tried it. `make serve` actually does work for me (although after running it initially I had to do `bundle exec jekyll serve --host example.com` because I'm not browsing it from localhost.) Thanks @thejameskyle!

This renders fine. FWIW I previously brought this up on slack and **sebmck**'s response was along the lines of "does it need to be documented?". Obviously yes in my opinion.

thejameskyle:
Can you just add like a `*` for the footnote?

jmm:
@thejameskyle Sure, see https://github.com/babel/babel.github.io/compare/master...jmm:caveats-spread2. I thought about doing that originally but decided not to because it'll fall down as soon as something else on the page needs some kind of footnote and for the time being I could put it right underneath. But someone can cross that bridge if and when something else needs a footnote. Maybe then start doing `<sup>[1]</sup>` kind of thing.

hzoo:
Going to close since we removed that page (although I think we should bring it back somewhere)

jmm:
> although I think we should bring it back somewhere

:+1: Not sure why it went away in the first place (I'm assuming some things just got lost in the shuffle).

hzoo:
If you want to just add it back, or I'l do it later

PR #398, opened 2015-08-04
Add babel-node production warning

### LINKED ISSUE #397: babel-node documents should explain the perils of not using kexec
Without the kexec module, you can get into a really ugly situation where the `child_process` dies but its death or error never bubbles up.  For more info see https://github.com/babel/babel/issues/2137.

It would be great if the docs on babel-node explained that it is not aimed for production and that without kexec installed that it has bad behaviour.

https://babeljs.io/docs/usage/cli/#babel-node

thejameskyle:
Resolves #397

cc @sebmck since I stole your words

### DIFF
--- a/docs/usage/cli.md
+++ b/docs/usage/cli.md
@@ -82,6 +82,16 @@ $ babel --out-file script-compiled.js < script.js
 
 ## babel-node
 
+<blockquote class="babel-callout babel-callout-warning">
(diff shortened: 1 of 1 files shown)


sebmck:
:+1: It's phrased kinda weirdly though so feel free to modify it.

thejameskyle:
Cleaned it up slightly

PR #400, opened 2015-08-10
Add an AST inspector to the REPL

ForbesLindesay:
I often find myself wanting to know what the AST looks like for a given bit of code, especially while writing plugins for babel.  This pull request adds a tree view/inspector similar to what you would see in the chrome developer tools when you console.log an object.  It lets you explore the AST before and after transformation.  It also stays open at the same point in the AST as you edit the code.

There are a couple of TODO items outstanding:

1. I couldn't find a way to get at the AST pre-transformation, so ended up editing the generated babel code to expose babylon.  I realise this is a broken hack, what's the correct way of getting at the parser?
2. Currently, when-expanded, the AST Explorers cover up the error reporting space and the console output.  I will fix this properly if you are happy with the general direction this is headed?

Screen Shot

Video Demo: https://www.dropbox.com/s/epe7s93vntalk88/babel-ast.mov?dl=0

_Further thoughts:_

We could use this component in the console output to render objects/arrays as expandable things, rather than just JSON.stringify(ing) them.


thejameskyle:
Hmm this seems like a great idea for the plugins site should I ever get back around to it.

Otherwise, this might be better as a toggle somewhere.

ForbesLindesay:
I'd be happy to add another checkbox at the top.  Something like "Show AST"

thejameskyle:
:+1:

sebmck:
:+1: to "Show AST".

hzoo:
We should figure out how to use http://astexplorer.net/ somehow

hzoo:
Would be nice to combine/use astexplorer somehow @fkling. I think we'll close since most know about it already (we can add a link somewhere too)

PR #410, opened 2015-08-13
added examples for in browser environment (and other cases)

omerts:
Added examples for different ways to add plugins

### DIFF
--- a/docs/advanced/plugins/index.md
+++ b/docs/advanced/plugins/index.md
@@ -41,6 +41,12 @@ require("babel").transform("code", {
 });
 ```
 
+```javascript 
+require("babel").transform("code", {
+  plugins: [require("foo-bar")] // Useful for in browser environment
+});
+```
+
 [CLI](/docs/usage/cli/):
 
 ```sh
@@ -63,6 +69,12 @@ a colon after the plugin name. ie:
 ```javascript
 require("babel").transform("code", {
   plugins: ["foo-bar:after", "bar-foo:before"]
+});
+```
+
+```javascript 
+require("babel").transform("code", {
+  plugins: [{transformer: require("foo-bar"), position: 'after'}]
 });
 ```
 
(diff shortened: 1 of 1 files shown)


docs/advanced/plugins/index.md
```diff
@@ -41,6 +41,12 @@ require("babel").transform("code", {
 });
 ```
 
+```javascript 
+require("babel").transform("code", {
+  plugins: [require("foo-bar")] // Useful for in browser environment
```
thejameskyle: Why is this "Useful for in browser environment"?

omerts:
@thejameskyle Since babel's plugin manager tries to load the plugin during the transformation, and if the transform occurs in the browser, plugin is unavailable. 

PluginManager's code that throws the exception:
```javascript
var loc = _tryResolve2["default"].relative("babel-plugin-" + name) || _tryResolve2["default"].relative(name);
if (loc) {
  var plugin = require(loc);
  return {
    position: position,
    plugin: plugin["default"] || plugin
  };
} else {
  throw new ReferenceError(messages.get("pluginUnknown", name));
}
```
var loc, is always null because:
```javascript
resolve.relative = function (loc) {
  // we're in the browser, probably
  if (typeof Module === "object") return null;
```

Therefore, you have to load the plugin before the transformation, using the code in the example.

thejameskyle:
You're talking about usage with browserify or webpack?

omerts:
@thejameskyle I can't tell you about browserify, becuase I haven't used it. Webpack is my use case.

omerts:
To be more detailed, I am working on a fork of https://github.com/FormidableLabs/component-playground, which gives you a playground for demoing react components. The playground transforms the javascript code as it changes in the browser.

thejameskyle:
I'd rather not have that code comment (`// Useful for in browser environment`)

omerts:
Comment removed

PR #412, opened 2015-08-16
Add missing .md file-extension

michaelzoidl:
Saw the broken link on the [.babelrc](https://github.com/babel/babel.github.io/blob/master/docs/usage/babelrc.md) page, fixed it by adding the missing .md file-extension

### DIFF
--- a/docs/usage/babelrc.md
+++ b/docs/usage/babelrc.md
@@ -5,7 +5,7 @@ description: How to use the babelrc
 permalink: /docs/usage/babelrc/
 ---
 
-The entire range of Babel API [options](/docs/usage/options) are allowed.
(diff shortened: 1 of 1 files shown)


hzoo:
The link for `options` on http://babeljs.io/docs/usage/babelrc/ goes to http://babeljs.io/docs/usage/options/ which should be correct?

michaelzoidl:
Ah well, seems like the link just doesn't work on the github markdown - i wasn't on the babeljs.io pages, look to the github page of usage/babelrc.
Looks like it's important for the Jekyll routing to drop the file-ending (.md)
Then never mind :)

hzoo:
Thanks for looking into it anyway!

PR #420, opened 2015-08-22
Remove `metadataUsedHelpers` flag from External Helpers page

Daniel15:
`metadataUsedHelpers` is now on by default (as per 32f19aff99559e36c3d4fbb946af378a8126b8bf)

Also, the title should be in title case :)


thejameskyle:
Thanks

PR #424, opened 2015-09-01
Clarification, Webpack example, --save-dev amendment, and typo fix

davedx:


docs/usage/runtime.md
```diff
@@ -40,6 +40,14 @@ require("babel").transform("code", { optional: ["runtime"] });
...
 ```
 
+In Webpack:
+
+```js
+loaders: [
+  { test: /\.js$/, loaders: ['babel?optional=runtime']}
+]
+```
+
```
thejameskyle: I'd rather not put web pack in here. If we put every way to add this option it'd be really long.

davedx: Yeah, I did think that after sending the PR. It was useful to me but after looking at it I thought "actually that likes messy".

I'll remove it.

thejameskyle:
thanks

PR #430, opened 2015-09-04
Added a note about non-standard file extensions to the description of the 'nonStandard' option.

MatthewHerbst:
Use case: Many JSX developers mark their files as file.jsx. When using Node.js `require()` such as in a project using `Browserify`, `require()` will only search by .js and .json files for components by default. 

Currently when searching the docs, a user looking for "JSX" will look primarily at the `nonStandard` option since it enables support for JSX. However, many users are not aware that they will have to add a custom file extension to use `require()` to find components such as `require(./Component)` where Component is located in the file Component.jsx. 

Since this is a fairly common usage pattern, I have added a note to the description of the `nonStandard` option that alerts the user to look for additional information.


zertosh:
@MatthewHerbst I think this note makes more sense in the babelify readme than here

MatthewHerbst:
@zertosh The problem is that the Babelify README links to the Options on this page.

zertosh:
@MatthewHerbst yeah, since that's only a concern for babelify users

thejameskyle:
Yup, please move to babelify

PR #432, opened 2015-09-04
Don't initially hide the sub-navigation

silvenon:
The sub-navigation in the sidebar was displaying only on hover, which seemed unnecessary.


thejameskyle:
The reason for this at the time is that some of the pages had longer navigation than the average screen size, making the last few items unclickable as they went off the screen.

silvenon:
Oh, right. I didn't check other pages. Now it makes sense :+1:

PR #439, opened 2015-09-14
Fix brunch config example

es128:
The key for the plugin's config is just `babel`. It appears to have been listed as `ESbabel` because of a find/replace on the original key name `ES6to5`.


thejameskyle:
cool, thanks

PR #442, opened 2015-09-22
Update link to ES7 classProperties proposal

ThibWeb:
Quite a small change! The proposal has moved from https://gist.github.com/jeffmo/054df782c05639da2adb to https://github.com/jeffmo/es-class-properties, as it is indicated inside the gist.


thejameskyle:
Thank you

PR #443, opened 2015-09-22
Update setup instructions for Meteor

benjamn:
ICYMI, Babel is installed by default for all new Meteor apps!


thejameskyle:
Nice

PR #446, opened 2015-09-23
move trailing function commas proposal to stage 2 (docs)

michaelficarra:
Waiting on https://github.com/babel/babel/pull/2427.


hzoo:
Added to http://babeljs.io/docs/plugins/preset-stage-2/ since that page was removed

PR #449, opened 2015-09-28
Enable line wrapping in ACE editor

spicyj:
About half the time I open the Babel REPL it's because I want to check the implementation of one of the helpers, and it's a lot harder to do that if I have to scroll horizontally to see anything. That said, this does look a little more cluttered if you're _not_ interested in the helpers, so I understand if you don't want this.

screenshot of Babel REPL with line wrapping enabled

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -2,14 +2,14 @@
 
   /* Throw meaningful errors for getters of commonjs. */
   ["module", "exports", "require"].forEach(function(commonVar){
-    Object.defineProperty(window, commonVar, { 
+    Object.defineProperty(window, commonVar, {
       configurable: true,
       get: function () {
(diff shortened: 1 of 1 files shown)


hzoo:
Could compromise by adding it as an option somewhere?

thejameskyle:
In an ideal world, we could expand the helpers all the way and use Ace's code folding to collapse them by default

thejameskyle:
With this I'm afraid it might freak people out if they see a mess of code at the top of the generated output.

hzoo:
@spicyj Made a PR with line wrap as an option if we want that https://github.com/babel/babel.github.io/commit/e54f3cfb73ed843f142327f33421db448fa893dd

hzoo:
K gonna close in favor of the option if we still want to add it

PR #450, opened 2015-09-30
Update example to match JSX documentation

davidlormor:
According to the JSX documentation

> Since JSX is JavaScript, identifiers such as `class` and `for` are discouraged as XML attribute names. Instead, React DOM components expect DOM property names like `className` and `htmlFor`, respectively.

Edited docs to reflect current JSX standards.


thejameskyle:
thanks

PR #460, opened 2015-10-12
react.inlineElements global polyfill dependency

STRML:
`optimisation.react.inlineElements` breaks without the global polyfill.

See https://github.com/babel/babel/issues/2517 for more information.

### DIFF
--- a/docs/plugins/transform-react-inline-elements.md
+++ b/docs/plugins/transform-react-inline-elements.md
@@ -8,6 +8,16 @@ package: babel-plugin-transform-react-inline-elements
 
 Converts JSX elements to object literals like `{type: 'div', props: ...}` instead of calls to `React.createElement`.
 
+Note that this requires the global polyfill due to React's use of ES6 `Symbol` when validating React Elements.
+If `Symbol` is not present in the browser, your application will fail to render, as Babel will polyfill `Symbol`
+but React will not have access to that polyfill.
+
+Therefore, you must require the global polyfill at the entry point to your application:
+
+```js
+require("babel-polyfill");
+```
+
 This transform **should be enabled only in production** (e.g., just before minifying your code) because although they improve runtime performance, they make warning messages more cryptic and skip important checks that happen in development mode, including propTypes.
 
 ## Example


STRML:
Alternatively, we could suggest `require('babel-runtime/core-js/symbol');` which is not nearly as heavy. In my application, the full global polyfill added 130kB to the build, but the symbol shim only 8kB, and both fixed the issue.

STRML:
@sebmck Rebased this to work on v6.

Including the entire polyfill feels hacky / huge. In v5, I was using:

```js
global.Symbol = require('babel-runtime/core-js/symbol').default;
```

I haven't updated to v6 yet so I'm not sure what the equivalent would be. `require('babel-polyfill/node_modules/core-js/symbol').default` seems wrong.

thejameskyle:
> Not sure what's going on now, all issues have been removed in babel/babel.

We're in the middle of migrating to [Phabricator](https://phabricator.babeljs.io)

thejameskyle:
> We could, yeah. It's hugely important to notice this caveat because your app will break in production only, not in dev.

And a callout will get more attention

hzoo:
Added the callout

oliviertassinari:
I believe that it's no longer true since the *inlining* approach changed. I'm gonna submit a PR to remove this section from the documentation.

STRML:
@oliviertassinari Could you elaborate? Last I checked this is still necessary - on an older browser, the polyfill will be available via babel-runtime to userland code, meaning React elements are tagged with a polyfilled `Symbol.for('react.element')`, but since no polyfill is available to React, it tags with `0xeac7`.

See https://phabricator.babeljs.io/T2517 for more.

oliviertassinari:
@STRML This [commit](https://github.com/babel/babel/commit/d1b8db1532d82b5d533732412b1bb9b9b6987251) refactored the  implementation of the plugin. Thanks for sharing the issue. I think you are right :+1:.

STRML:
Right - that gave us a decent speedup but, unfortunately, does not fix the polyfill issue. 

It's too bad that this problem exists, because it has two terrible properties:

1. It doesn't happen in development, only production, and
2. It only causes issues on older browsers, i.e. the types that your devs probably aren't using as often as they should.

PR #461, opened 2015-10-12
add undocumented blacklist and optional to repl

kolodny:


kolodny:
`UriUtils.parseQuery` will happily parse whatever is in the qs of the hash, all this does is pass it along to `babel.transform` https://github.com/kolodny/babel.github.io/blob/17d28d796956fcef4a04d6bb68a4bbbcf69428a2/scripts/repl.js#L30

PR #468, opened 2015-10-19
Inform about babel-node limitations.

clentfort:
Include a callout to inform users that ES6-style module loading is not supported in `babel-node`.


loganfsmyth:
Maybe we want to say "may not function as expected" or something? It does work sometimes, but you'd need all your code to be on one line. or explicitly reassign it yourself.

PR #471, opened 2015-10-21
Add clarifications about supported ES6 methods to caveats

grrowl:
Babel can't distinguish between `arr.includes` and `str.includes` at compile time, so we added clarifications around why these methods might be missing (can't be inferred statically at compile-time) and a workaround (use the polyfill).

Happy for feedback on the wording here, I think this info should be included more obviously in more places, especially since the high-level proposition to developers is "Use next generation JavaScript, today." and we tout these methods (only supported with additional polyfill) in [our overview of ES2016 features](http://babeljs.io/docs/learn-es2015/#math-number-string-object-apis)

### DIFF
--- a/docs/advanced/caveats.md
+++ b/docs/advanced/caveats.md
@@ -28,7 +28,9 @@ You may alternatively selectively include what you need:
 ## Classes
 
 Built-in classes such as `Date`, `Array`, `DOM` etc cannot be properly subclassed
-due to limitations in ES5.
+due to limitations in ES5. This means, because Babel can't distinguish between 
+`(arr).includes()` and `(str).includes()` statically, you'll have to use [the
+polyfill](/docs/usage/polyfill/) to use these methods.
 
 ## ES5
 


grrowl:
That said, including the following lines have no effect on defining `Array.prototype.includes` or `[].includes`:

```
require('babel-runtime/core-js/array/includes');
```
or
```
require('babel-core/polyfill');

loganfsmyth:
@grrowl It's not clear why this clarification would belong in a section of the page that is talking about subclassing builtins.

Babel is a syntax converter, it entirely relies on polyfills to provide new non-syntactic features. Your comment about `require('babel-core/polyfill');` is incorrect, as that will indeed globally polyfill `Array.prototype.includes` for you. Part of using Babel is understanding what polyfills to use and when. This is not done for you automatically unless you ask by enabling `runtime` and by doing so, you are essentially telling Babel that you don't expect any prototype methods to be used.

As you have pointed out, there is a note about this on the `runtime` page, which is the right place for it since `runtime` is the only time Babel makes any attempt to analyse the code and inject polyfill calls. If you feel like that note should be expanded, feel free to do that though.

grrowl:
Thanks for the clarification — my test certainly validates what you said as I've got runtime and core hooked up. The issue is mentioned in babel-runtime, which is the source of the problem, so you're totally right. Thanks for your time!

PR #479, opened 2015-10-28
Update module docs with new UMD output

flashbackzoo:


hzoo:
This page was removed - ref https://github.com/babel/babel.github.io/issues/485

PR #490, opened 2015-10-30
fix doc: transform-async-to-generator

jessy1092:
Thanks babel. The amazing tool!! Just help to fix some v6 document.


thejameskyle:
Thanks

PR #492, opened 2015-10-30
Fix the Options description for presets and plugins

oligot:
presets description referenced plugins instead of presets
plugins description referenced presets instead of plugins


thejameskyle:
Thanks

oligot:
This was really fast :wink:

PR #505, opened 2015-10-30
add note about transform plugins auto requiring syntax plugin

hzoo:


thejameskyle:
Want to make it a callout instead?

hzoo:
like this?

thejameskyle:
Thanks

PR #513, opened 2015-10-31
add some docs about plugin options

hzoo:
Ref https://github.com/babel/babel/issues/2710


docs/plugins.md
```diff
@@ -131,3 +131,17 @@ These plugins apply transformations to your code.
...
+
+```js
+{
+  "plugins": [
+    ["transform-async-to-module-method", {
+      module: 'bluebird',
+      method: 'coroutine'
+    }]
+  ]
+}
```
thejameskyle: Want to add the CLI form?

thejameskyle: never mind i thought that was added

PR #516, opened 2015-11-01
fix usage/browser docs

hzoo:
Not sure what you need to add to the `.babelrc` - can update instead of that's correct


thejameskyle:
Thanks

Issue #508, opened 2015-10-31
Link each plugin's doc page to its source code


callumlocke:
It would be helpful if each doc page linked to its source code within the babel repo.

For example http://babeljs.io/docs/plugins/preset-es2015/ should include a link to https://github.com/babel/babel/tree/master/packages/babel-preset-es2015

thejameskyle:
Sure, wanna make a PR?

callumlocke:
done. also with npm links, mainly because it looks better.

tomByrer:
@callumlocke close?

PR #524, opened 2015-11-01
Add Github and npm links to package docs

callumlocke:
See #508.

I've added a `package` property that can be set in the front matter of any docs page, for specifying the name of the relevant npm package (if any).

In the layout this property is used to add GitHub and npm links in the header.

How this looks on a page with a description:

image

How it looks on other pages:

image

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -8,6 +8,9 @@ layout: default
       <div class="col-md-12">
         <h1>{{page.title}}</h1>
         <p>{{page.description}}</p>
+        {% if page.package %}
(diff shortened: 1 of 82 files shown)


thejameskyle:
Hmm, it's not so obvious that these are links. Want to make them yellow with underlines?

callumlocke:
Underlines aren't used for links anywhere else on the site, except on hover.

I originally had the links as yellow, but it clashed with the yellow headlines which are not links.

Are you sure you want me to make them yellow and underlined? I'll do it if you like, but personally I think it would make it look ugly and I don't think it's necessary. It seems fairly obvious they're links just from their placement as two little words in a bullet-separated list (same style as in the footer), both of which are well known external sites.

thejameskyle:
> Are you sure you want me to make them yellow and underlined?

yes

callumlocke:
ok done, looks like this now:

image

thejameskyle:
Thanks!

hzoo:
Looks good!

PR #526, opened 2015-11-02
Adds hackerYou logo

Rchristiani:


images/users/hackeryou.svg
```diff
@@ -0,0 +1,21 @@
...
+        </g>
+    </g>
+</svg>
```
thejameskyle: Can you run this file through `svgo`?

thejameskyle: Also, the dimensions need to be `500x200 (2.5x1)`

Rchristiani:
I was just in the middle of doing that! I think this fixes those issues.

thejameskyle:
Thanks http://babeljs.io/users/#hackerYou

PR #531, opened 2015-11-02
New User: KongGe

taberh:


thejameskyle:
Thanks http://babeljs.io/users/#KongGe

PR #551, opened 2015-11-09
Add link to blog post about function bind syntax

spadin:
I was having a difficult time figuring out what features some plugins add. I think a "Learn more" section with either a code example or a link would be helpful so people can read more about the feature. Would it be okay if I start adding this section to a few of the plugin's documentation that I know about?

### DIFF
--- a/docs/plugins/transform-decorators.md
+++ b/docs/plugins/transform-decorators.md
@@ -21,3 +21,7 @@ Add the following line to your `.babelrc` file:
   "plugins": ["transform-decorators"]
 }
 ```
+
+## References
+
+* [Proposal: Javascript Decorators](https://github.com/wycats/javascript-decorators/blob/master/README.md)
--- a/docs/plugins/transform-function-bind.md
+++ b/docs/plugins/transform-function-bind.md
@@ -21,3 +21,7 @@ Add the following line to your `.babelrc` file:
   "plugins": ["transform-function-bind"]
 }
 ```
+
+## References
+
+* [Babel Blog: Function Bind Syntax](/blog/2015/05/14/function-bind/)
--- a/docs/plugins/transform-class-properties.md
+++ b/docs/plugins/transform-class-properties.md
@@ -21,3 +21,7 @@ Add the following line to your `.babelrc` file:
   "plugins": ["transform-class-properties"]
 }
 ```
+
+## References
+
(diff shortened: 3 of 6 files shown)


docs/plugins/transform-function-bind.md
```diff
@@ -21,3 +21,7 @@ Add the following line to your `.babelrc` file:
   "plugins": ["transform-function-bind"]
 }
 ```
+
+## Learn more
+
+Learn more about [function bind syntax](http://babeljs.io/blog/2015/05/14/function-bind/).
```
thejameskyle: - `Learn more => References`
- Make it just a bulleted list of links
- `Learn more about function bind syntax => Babel Blog: Function Bind Syntax`
- Can you use a relative `/blog/...` url?

thejameskyle:
> Would it be okay if I start adding this section to a few of the plugin's documentation that I know about?

Sure! Could you just make the changes I commented above and do that everywhere?

It'd also be good to link stuff like MDN/spec/proposals wherever possible

spadin:
Thanks for the comments. I've updated this page. I'll add similar links and headings to the other plugins that I understand. Would you prefer one PR with several changes or one PR for every page updated?

hzoo:
Putting it in this one sounds fine to me

spadin:
Sounds good. I'll update this PR over the next couple of days. I'll leave a note when I think it's ready for another review. Thanks!

hzoo:
@spadin You can probably just use the links from https://github.com/tc39/ecma262#current-proposals or maybe link to that too

spadin:
Sweet. Thanks for the link. I'll use those links wherever I can.

docs/plugins/transform-decorators.md
```diff
@@ -21,3 +21,7 @@ Add the following line to your `.babelrc` file:
   "plugins": ["transform-decorators"]
 }
 ```
+
+## References
+
+* [Javascript Decorators](https://github.com/wycats/javascript-decorators/blob/master/README.md)
```
thejameskyle: Can you prefix these proposals with `Proposal:`

In general it'd be good to say where they are from first (ie. `MDN: Classes`, `Spec: Classes`, `Proposal: Decorators` etc.)

spadin: Good point. Will do.

spadin:
If you like these change, could we merge it? Then I can take some time this weekend to look for other pages I can update. Just so we don't leave this PR open for too long.

PR #568, opened 2015-11-13
Babel setup: Clarified the need of enabling a plugin in .babelrc

dominicumbeer:
I had some trouble configuring Babel in my project the first time. It wasn't totally clear for me that enabling of a plugin is mandatory.

I thought we can make it more clear if we add a small example to it. What do you think?


dominicumbeer:
Other changes have been merged which address my concerns.

PR #573, opened 2015-11-17
Update usage.md

PeterDaveHello:
Should use `babel-browser` package instead of `babel-core` for `browser.js`

### DIFF
--- a/_includes/tools/babel_in_browser/usage.md
+++ b/_includes/tools/babel_in_browser/usage.md
@@ -1,7 +1,7 @@
 Add the `type="text/babel"` attribute to your `<script>` tags. For example:
 
 ```html
-<script src="node_modules/babel-core/browser.js"></script>
+<script src="node_modules/babel-browser/browser.js"></script>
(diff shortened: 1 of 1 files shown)


hzoo:
Thanks although it's removed now - https://babeljs.io/docs/usage/browser/ so maybe we should just remove that section

thejameskyle:
Yeah, let's kill it.

thejameskyle:
Oh, you did already

hzoo:
@thejameskyle But I didn't remove https://babeljs.io/docs/setup/#babel_in_browser yet which is what I meant :smile:

_includes/tools/babel_in_browser/usage.md
```diff
@@ -1,7 +1,7 @@
 Add the `type="text/babel"` attribute to your `<script>` tags. For example:
```
VukDukic: T

PeterDaveHello:
:+1:

PR #574, opened 2015-11-17
Add mention of ReactJS.NET

Daniel15:
Add mention of ReactJS.NET for C# and MSBuild support. Eventually I'd like to split ReactJS.NET into a smaller project that isn't actually explicitly for React, but this will do for now. This part of ReactJS.NET isn't actually specific to React in any way, it's just wrapping Babel in the V8 or Chakra/MSIE engine (similar to the Ruby integration)

### DIFF
--- a/_data/tools.yml
+++ b/_data/tools.yml
@@ -14,6 +14,7 @@
     gulp: Gulp
     jspm: jspm
     make: Make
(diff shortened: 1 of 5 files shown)


thejameskyle:
Thanks

PR #579, opened 2015-11-17
update homepage

hzoo:
Ref https://github.com/babel/babel.github.io/issues/578

Do we want to link to the plugins?

Or we can mention using presets or in a new section


aamirafridi:
I think you end the sentence with ... Using plugins and presets (see details here)
Link to plugins, presents and details should like to an example or the list of changes Made in babel 6

Regards,
Aamir

thejameskyle:
Oh, links would be nice too

PR #584, opened 2015-11-19
New user: blendle

jtangelder:


thejameskyle:
For the image, can you resize to be 500x200 (2.5x1) and run it through `svgo`?

Also this needs a rebase.

PR #587, opened 2015-11-19
New User: TES Global

aamirafridi:


aamirafridi:
@thejameskyle logo size is corrected and I ran it through svgo. Hope its ok to merge this now :|

PR #601, opened 2015-11-25
Updated require hook usage to use babel-register

noamokman:
Just updated the [Require Hook](http://babeljs.io/docs/usage/require/) usage page.
Should now be `babel-register` instead of `babel-core/register`


thejameskyle:
Thanks

PR #603, opened 2015-11-26
Package.json script `--stage 0` to `--stage-0`

caalberts:
Should this be `--stage-0` instead of `--stage 0`?


thejameskyle:
`--stage` doesn't even exist anymore, this should be specifying presets.

PR #604, opened 2015-11-26
Delete babel in browser setup

thejameskyle:
Trying to delete these file from the GitHub UI


thejameskyle:
Nice, you can do this all in the GitHub UI

hzoo:
Very cool

PR #609, opened 2015-11-28
Move the "no default transformations in v6" note to the usage section.

pygy:
See https://phabricator.babeljs.io/T6711

tl;dr: I'm trying out Babel and wanted to set up a one shot shell session. I thus skipped the `.babelrc` section (no need for a permanent config file at this point) and ended up surprised that vanilla Babel is just a JS syntax validator.

Update: accidently a whole word (still absent in the commit message title, hopefully you won't mind).

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -37,6 +37,13 @@ custom_js_with_timestamps:
 
       {% include tools/items.md name="usage" %}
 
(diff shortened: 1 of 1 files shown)


pygy:
Actually, it may be better to add that note to the CLI page... Working on another PR...

PR #610, opened 2015-11-28
 	Add a note about the lack of transformations in v6, give a presets example.

pygy:
Sorry for the initially blank message, I hit return instead of backspace while editing the title... So:

See https://phabricator.babeljs.io/T6711

In their current state, the docs are not very helpful for newcomers.

The only mention of the lack of transformations post v6 is in the `.babelrc` section, and you must navigate too many pages to get Babel to perform what you'd expect it to do (i.e. more than syntax validation).

This may make #609 superfluous.

### DIFF
--- a/docs/usage/cli.md
+++ b/docs/usage/cli.md
@@ -20,6 +20,14 @@ available from the command line.
 $ npm install --global babel-cli
 ```
 
+Out of the box, Babel (since v6) doesn't perform any transformation, it only validates the syntax. You may want to install [plugins](/docs/plugins) or more likely whole [presets](/docs/plugins/#presets) before going further. For the sake of the example, we'll use the [es2015](/docs/plugins/preset-es2015/) preset.
+
+```sh
+$ npm install babel-preset-es2015
+```
+
+For Babel to use the preset you must either pass it in the command line as `--presets es2015` (omitted in the rest of the page for readability) or add it to the [`.babelrc`](/docs/usage/babelrc/) file, in the `"presets"` section.
+
 ## babel
 
 #### Compile Files
--- a/docs/plugins/preset-es2015.md
+++ b/docs/plugins/preset-es2015.md
@@ -37,7 +37,7 @@ $ npm install babel-preset-es2015
 
 ## Usage
 
-Add the following line to your `.babelrc` file:
+Pass `--presets es2015` as a command line option or add the following line to your `.babelrc` file:
 
 ```json
 {


docs/usage/cli.md
```diff
@@ -20,6 +20,14 @@ available from the command line.
 $ npm install --global babel-cli
 ```
 
+Out of the box, Babel (since v6) doesn't perform any transformation, it only validates the syntax. You may want to install [plugins](/docs/plugins) or more likely whole [presets](/docs/plugins/#presets) before going further. For the sake of the example, we'll use the [es2015](/docs/plugins/preset-es2015/) preset.
+
+```sh
+$ npm install babel-preset-es2015
```
thejameskyle: `--save-dev`

pygy: I can add it if you really want, but I don't think it's necessary, for the same reason you install babel-cli globally rather as a dev-dependency (or even gulp-babel). This is for the sake of trying things on the command line.

thejameskyle: Babel presets cannot be installed globally. Also you should be installing `babel-cli` locally rather than globally.

pygy: For the presets, locally, but not saved, yes, that's what I'm proposing.

The file whose diff we're discussing instructs to install babel-cli globally. This is actually visible in the snippet [above](https://github.com/babel/babel.github.io/pull/610/files#diff-2f91ae7edae087838435be4adf7bc557L20).

thejameskyle: >  instructs to install `babel-cli` globally

Then we should rewrite this to explain that you should install it locally and use it from npm scripts.

thejameskyle: Also we should still `--save-dev`, please do that

pygy: Serious users already know about npm scripts. Adding these parameters add noise and clobber the flow of the article.

The article is a tutorial for using Babel from the command line. For dabbling. To get a feel of how the thing works. It's pretty good as it is.

All examples show interactive commands and hint at output, and it's great as is. 

If babel really shouldn't be used from the CLI at all, you may want to rename `babel-cli` `babel-npm` and refuse to run outside of `npm` scripts.

pygy:
Hopefully you're in a bad day...

thejameskyle:
I'm not going to go back and forth on this for two hours when I'm already working on rewriting all of this.

pygy:
Hopefully for the best.

I hoped that my perspective as a new Babel user[†] would be useful to you. I'm basically the target audience of these pages, advanced users don't need them.

The tutorial in its current form allows to follow along on the command line, interactively discovering the `cli` tool, and it's rather good IMO.

A NPM scripts-based workflow would be much more cumbersome.

----
†. I've actually submitted a Babel patch in the past, but it was for a bug encountered by the user of a lib of mine that exports a default IIFE, among other module options, I hadn't used Babel personally.

thejameskyle:
> I hoped that my perspective as a new Babel user[†] would be useful to you. I'm basically the target audience of these pages, advanced users don't need them.

Of course it's useful, but I gone through hundreds of individual people's problems, and not installing things locally is a huge issue (especially after 6.0 was released). This documentation needs to teach users the best practices.

> The tutorial in its current form allows to follow along on the command line, interactively discovering the cli tool, and it's rather good IMO.
> 
> A NPM scripts-based workflow would be much more cumbersome.

Many users will stop reading documentation as soon as they get something running. If you get them doing that too early, many will never read the parts that explain what they actually should be doing. So even though it might be a few more steps to follow initially, it pays of in the long run because everyone ends up doing things the right way.

---

I'm sorry I didn't sit down and explain this all right away, but I go through hundreds of issues and it becomes exhausting to have to explain myself to every single person. In the future please just trust that as the maintainer of a project I know what I'm talking about.

And please don't go saying mean things about me on twitter just because I wasn't willing to drop 20 minutes of my time to write out a lengthy response on a relatively inconsequential issue.

This is why people burn out on projects.

pygy:
> And please don't go saying mean things about me on twitter...

Not very proud of that one, sorry. The exchange was painful to me and I needed to vent, but I should have refrained, and I apologize.

People burn out because:

1. They are overworked, and it looks like you may be reaching capacity. Running such a large and growing project as a two men show must be exausting.

2. They don't get the respect they think they're due. In that regard, you reaped what you sowed. Bossing around volunteer contributors is not how you earn respect. If what you did at the begining of this thread is representative, it's not surprising you're running into point 1. Possibly a vicious cycle.

I know you're busy, this is a big project, that's why I didn't object, at least initially, to the terseness of your first request, even though I found it off-putting. That's not why I objected, though. While I agree that the page needs an overhaul, adding `-D` to the then current version made it worse and I don't wanted my name associated with that.

Architecturally-speaking, trimming down the core and pushing functionality to plugins makes a lot of sense, but applying the same treatment to `babel-cli` turned it into a tool not meant for standalone interactive use, which is a bit damning. At best now it can annoy the ["Standard" JS](http://standardjs.com/#rules) folks, but that's it (-;

So you're left with the option of littering the current directory (what I suggested), which is not very good. What makes it worse is that the tutorial doesn't give any context on where to run the commands. Saving the dependencies adds more garbage, which is why I objected to it.

The alternative is to build a full npm project with suitable options pre-installed, as a playground. It raises the barrier of entry, though, and makes the workflow more cumbersome. You may want to create a `babel-playground` repo, to be cloned for that usage.

To ease said workflow, I just realized that you could run a shell in the `npm run-script` environment. Here's a first attempt at doing in a cross platform way. Maybe there's a way to do it already, but I couldn't find one.

https://github.com/pygy/npm-run-shell

Obviously WIP, not yet tested anywhere beside my Mac laptop.

----

But the core of the issue is elsewere. `babel-cli`,  there's a interactivity promise in that name that's currently not fulfilled.

For it to make sense again you'd either need to rename it or to add a reasonable default preset, like `ES$currentYear`.

PR #613, opened 2015-12-02
Upgrade REPL to Babel 6

Daniel15:
Upgrading the REPL to use Babel 6. This is a work in progress, but I'm posting it to get any initial feedback.

Live demo: http://babeldev.dan.cx/repl/

 - [x] Load Babel from NPM rather than cloning it in the build script
 - [x] Use Webpack or Browserify to build the REPL + Babel
 - [x] Update Makefile
 - [x] Make options (presets, etc) configurable
 - [x] Optimize Webpack config
 - [x] Script to push to `gh-pages`
 - [x] Use latest Babel 6.x.x rather than hard-coding a specific version
 - [x] Display Babel version at top right of REPL (replace the current "work in progress" notice)


Daniel15:
Okay, this is pretty much ready to review, with the exception of a script to push it to Github Pages. In the past I've used [gulp-gh-pages](https://github.com/shinnn/gulp-gh-pages) for that (eg. that's what I use for [react-magic](https://github.com/reactjs/react-magic)) which I can do here too. Are you okay with Gulp or would you just prefer a simple shell script?

hzoo:
Whatever is simpler probably? I think gulp is fine since we are using it in babel now as well.

Daniel15:
Cool, I'll do that tonight. Currently I'm just using `rsync` to push it to http://babeldev.dan.cx/repl/ for testing.

hzoo:
BTW great effort - will make debugging issues/PR that much easier

sparty02:
Awesome work on this @Daniel15 !    Did you purposefully pin down the babel versions (vs using latest)?

thejameskyle:
It would be best to put the build into Babel itself when deploys happen

Daniel15:
@sparty02 

> Did you purposefully pin down the babel versions (vs using latest)?

Yes, in order to get a repeatable build. Generally version ranges should only be used in shared libraries, not in top-level apps. Apps themselves should always be using specific version numbers (ideally `npm shrinkwrap`ped I guess) so you're always on a known good version.

@thejameskyle 

> It would be best to put the build into Babel itself when deploys happen

The idea with the site is to deploy it every time a change is made, through a post-commit hook or similar. I don't think Babel's build should trigger the site to be built, as updates to the website don't necessarily line up with updates to Babel itself (that is, pages on the site are totally separate from Babel itself).

Daniel15:
@thejameskyle - Although I guess the idea of automatically updating Babel + deploying the site  whenever a new Babel version is released is a good one, so the REPL is always up-to-date with the latest bug fixes. We can do that separately to this update though, as that doesn't exist right now with the Babel 5 REPL either :)

sparty02:
> I don't think Babel's build should trigger the site to be built, as updates to the website don't necessarily line up with updates to Babel itself (that is, pages on the site are totally separate from Babel itself)

This is debatable.  You could argue that it woudl be ideal for the REPL to always reflect ```babel@latest```.

sparty02:
jinx!

Daniel15:
> You could argue that it woudl be ideal for the REPL to always reflect babel@latest.

Yeah, I think I've been convinced that that's a good idea! I'll change it :stuck_out_tongue: 

At least using the latest 6.x.x version - I don't think we'd want it to magically switch to Babel 7 whenever that's released, given there'd probably be breaking API changes in a major release.

sparty02:
> At least using the latest 6.x.x version - I don't think we'd want it to magically switch to Babel 7 whenever that's released, given there'd probably be breaking API changes in a major release.

That makes sense to me.  So, babel major release --> REPL updates --> REPL reflects babel latest major again

Daniel15:
> it may be nice to display the version of babel that was resolved when building the REPL, in the top corner of the REPL site.

Good idea, I'll add that too. I think we already have the version available somewhere, it's just a matter of displaying it.

I won't have time to complete this tonight, but I'll update it over the weekend :+1:

thejameskyle:
I don't want the website to have a build system at all, most thing are merged from the GitHub UI.

The only thing that requires a build is the REPL, which only needs to be updated when Babel itself updates. So it should live there

Daniel15:
Having a build step doesn't disallow you from merging from the Github UI.
The build can be done as a post commit hook.

The REPL also needs to be updated if any changes are done to its JavaScript
file on the site.

Daniel15:
Okay, updated based on the feedback. I upgraded all the Babel packages to the latest version, and changed it to `^6.3.13` so it accepts newer versions too. The REPL now shows the loaded Babel version at the top right.

I added a proof-of-concept of deploying the site using TravisCI. Basically, pulls the repo, builds the site, then pushes to the `gh-pages` branch, like most other sites on Github use (the React site does something very similar, although I'm not sure if it's automated yet). It works pretty well, and merging pages from the Github UI works fine since TravisCI runs the build on commit. The config would need tweaking for the real Babel site, I just wanted to confirm that my idea would work as expected.

My proof-of-concept built site on Github Pages is at http://daniel15.github.io/babel-site/repl/, although it doesn't render properly since the site expects to be at the root (and thus all the CSS and JS links are wrong :stuck_out_tongue:). However, I manually inspected the JavaScript files to verify that they were compiled correctly. The working demo is still on my own server at http://babeldev.dan.cx/repl/ :smile:

thejameskyle:
> The REPL also needs to be updated if any changes are done to its JavaScript
file on the site.

It's far more common that the REPL needs to be updated when new Babel releases go out. So that needs to be considered as well.

Seb and I have also talked about a much different UI for this. We want to be able to configure Babel with any combination of the built in presets or plugins, which would be similar to exposing a `.babelrc` file in the REPL UI.

That being said, I'm very happy with this UI in the interim.

Daniel15:
> It's far more common that the REPL needs to be updated when new Babel releases go out. So that needs to be considered as well.

Yeah, for sure. I imagine that's also an issue today - someone needs to manually push the site again when a new Babel release goes out. In the meantime, if it were to use TravisCI or some other similar system (like I did in this diff), you could just click a button to redeploy the site. Since Babel comes from NPM now, that would automatically pull the latest version of Babel. I think that's better than what exists today, right? :smile: 

>  We want to be able to configure Babel with any combination of the built in presets or plugins, which would be similar to exposing a .babelrc file in the REPL UI.

Yeah I had the idea of making something that's fully customisable (including plugins), and potentially even allowing custom plugins to be added. I just didn't want to do anything too ambitious and figured it's better to start with a smaller scope (mostly sticking with what already exists) and incrementally improve things. Personally, the UI that exists today suffices for my needs, which is mostly what I was aiming for :stuck_out_tongue:

hzoo:
Do we want to include babel-polyfill/runtime?

kolodny:
Would it be possible to include a mechanism to not break the old http://babeljs.io/repl/ links? Maybe version the url like http://babeljs.io/repl-6.3.21/ or something similar

Daniel15:
@hzoo 

> Do we want to include babel-polyfill/runtime?

That's probably a good idea :) I'm not going to have much free time in the next few weeks, but feel free to clone my branch and add the polyfill.

@kolodny

> Would it be possible to include a mechanism to not break the old http://babeljs.io/repl/ links?

Is there anything in particular that breaks with the links? I guess I'd need to convert the old settings (eg. "experimental", etc.) to the equivalent Babel 6 presets, but apart from that the old links should work fine. Do you have an example of an old link that no longer works?

kolodny:
Nope, nothing jumps out at me. I was just fearful about every old repl link breaking when the new repl is up. As long as the presets match, and load the same way on a new state of the page, it should be ok

thejameskyle:
I'm okay with the urls breaking, it's not like the output is going to be the same between versions and we're not going to version the REPL for multiple versions of Babel

loganfsmyth:
Some things will certainly break for features that Babel 6 no longer supports like TCO and await*, but I do feel like we should make at least some small effort to map the old flags over to the new ones if we can. It doesn't have to be perfect though.

kentcdodds:
Thanks for pointing me to this @hzoo. I was concerned that the updated repl would break all my examples in my [ES6 workshop slides](http://slides.com/kentcdodds/es6-workshop#/4), but it seems to work [just fine](http://babeldev.dan.cx/repl/#?experimental=true&evaluate=true&loose=false&spec=false&playground=true&code=function%20f()%20%7B%0A%20%20%7B%0A%20%20%20%20let%20x%3B%0A%20%20%20%20%7B%0A%20%20%20%20%20%20%2F%2F%20okay%2C%20block%20scoped%20name%0A%20%20%20%20%20%20const%20x%20%3D%20%22sneaky%22%3B%0A%20%20%20%20%20%20%2F%2F%20error%2C%20const%0A%20%20%20%20%20%20%2F%2F%20x%20%3D%20%22foo%22%3B%0A%20%20%20%20%7D%0A%20%20%20%20%2F%2F%20error%2C%20already%20declared%20in%20block%0A%20%20%20%20%2F%2F%20let%20x%20%3D%20%22inner%22%3B%0A%20%20%7D%0A%7D%0A%0Af()%3B) :D :D Thanks!

scripts/repl.js
```diff
@@ -117,6 +122,51 @@
...
+    Object.keys(presets).forEach(presetName => {
+      var $group = document.createElement('div');
+      $group.className = 'form-group';
+
+      var $label = document.createElement('label');
+      $label.htmlFor = 'option-' + presetName;
+      $group.appendChild($label);
+
+      var $input = document.createElement('input');
+      $input.name = 'preset';
```
thr666: Giselle martino

paulfalgout:
This'd be a great addition. Surely a babel 6 repl at this point is more useful to the community than keeping the babel 5 stuff working?

hzoo:
It's been a while - what we do we need to get this merged?

kittens:
I'm not concerned with breaking old links, backwards compat shouldn't really be a priority for this PR. Someone can do it afterwards if they want. @Daniel15 would you mind rebasing? Sorry this has sat for so long. This looks good though!

Daniel15:
Sure, I'll try rebasing it later tonight. A while back I created [babel-standalone](https://github.com/Daniel15/babel-standalone) to replace babel-browser, using some of the ideas from this PR. Perhaps I should update this PR to pull in babel-standalone from NPM instead of directly bundling Babel itself.

@hzoo and @kittens I sent you an email about this a while back for a bit more context on the changes. I'll copy it here:

> Hello!

> I've been [working on upgrading the REPL on the Babel site to use Babel 6](https://github.com/babel/babel.github.io/pull/613). I just wanted to check whether you're okay with the changes I've done so far, and figured I'd start an email thread for more focused discussion :)

> Here's a demo of the changes: http://babeldev.dan.cx/repl/

> An overview of the changes I've made:
> * Latest 6.x.x version of Babel is pulled from NPM, rather than being checked in as a minified blob in the site repo. Much cleaner :D
> * REPL JS is built using Webpack
> * Removed the "Experimental", "Loose mode" and "High compliancy" checkboxes and replaced them with checkboxes to select the presets to use. I wanted to allow selection of individual plugins but didn't get around to it, and wanted to just keep the scope small for now, in case anyone wants to completely revamp the REPL.
> * Added a TravisCI configuration to build and publish the site, so it can be automatically built on every commit.
>    * This could/should also be done when the Babel repo changes. It could be added as a webhook.

> This is similar to how many many other Github Pages sites are deployed, in order to work around the pretty large limitations in Github Pages' core functionality. The React site is very similar, and in fact it looks like the React site was also changed to use TravisCI to push the site rather than having to build it manually as was the case in the past. Using Webpack or Browserify or whatever to build the REPL also allows it to easily be extended in the future, for example if you ever want to rewrite it using React (which would use a similar build step, since Github Pages doesn't support JSX out-of-the-box). Unfortunately this does mean that build artifacts still need to be checked in, but at least they can be hidden away on a generated branch.

> One thing I noticed is that the Babel site is an organisation Github Pages repo rather than a regular Github Pages repo, which means the site lives in the master branch rather than the gh-pages branch (thanks, Github). I think this means the "raw" site would need to live in some other branch (like dev or something) which could be set to the default branch, and the compiled site would go in master. I'm not quite sure how to update my pull request such that it goes into a different branch, though.

> Let me know if you've got any questions, feel free to reply to this email or message me through Facebook or Google Hangouts. I'd love to get your feedback and see what else I could do to help get this merged :D

> Thanks!

Daniel15:
Actually I might be able to just use Babel-standalone through CDNJS and totally avoid the build step, which would simplify things quite a bit. I'll play around with it and see :)

Daniel15:
Okay, I just updated the pull request @hzoo @kittens. I'm now pulling `babel-standalone` from CDNJS, which made it significantly simpler :tada: . Of course, we could always switch back to the site building its own version of Babel, but that can come in the future :smile:

paulfalgout:
Thanks @Daniel15 etc for jumping on this so quick!

Daniel15:
Free time comes and goes, I was lucky to have lots of free time today :stuck_out_tongue:

thejameskyle:
Tried it out locally and it works great so lets just ship it

Daniel15:
Yesss it's working :D http://babeljs.io/repl/

kittens:
Thanks so much for this! Sorry this took so long to get addressed. It's
super appreciated.

On Sunday, 6 March 2016, Daniel Lo Nigro <notifications@github.com> wrote:

> Yesss it's working :D http://babeljs.io/repl/
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/babel/babel.github.io/pull/613#issuecomment-192859737>
> .
>


-- 
Sebastian McKenzie

hzoo:
Thanks Daniel! :100:

PR #631, opened 2015-12-09
Explain how babelrc works with require hook

callumlocke:
based on [/packages/babel-register/src/node.js#L54-L57](https://github.com/babel/babel/blob/e96f3544f6a82345a469fe55b3b05c7f54536f11/packages/babel-register/src/node.js#L54-L57)


thejameskyle:
Thanks

PR #644, opened 2015-12-16
optimize images

marcobiedermann:
Running all images through [ImageOptim](https://imageoptim.com/) saves 1.1 MB out of 3.6 MB without losing any quality.

babel-images


thejameskyle:
Nice, thanks

PR #645, opened 2015-12-16
add will-change property to sticky header

marcobiedermann:
Position: fixed adds a lot of repaint on scroll to the page. Adding `will-change: transform` will promote it to it's own layer.

babel-sticky-header


thejameskyle:
Cool

PR #649, opened 2015-12-18
Add docs for the extend option

baer:
Add docs for the extend option added here: https://github.com/babel/babel/commit/c8901cf98df898c3839135ad7dbfbeab43abdb08
And implemented here: [option-manager.js#L212-L220](https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/file/options/option-manager.js#L212-L220)

I'm excited to have this feature! We were feeling quite stuck until I found this :smile:


chaseadamsio:
:+1: this is def.useful.

PR #657, opened 2015-12-23
New User: Kogan.com

choonkending:
- [x] Tested locally
- [x] Ran through `svgo`


choonkending:
@thejameskyle Thanks!

Issue #391, opened 2015-07-30
New User: SparkPost


AimeeKnight:
Please fill the following out:

- Name: SparkPost
- Homepage url: https://www.sparkpost.com/
- Brand Guidelines/Licensing: NA
- Logo: 
![sparkpost](https://cloud.githubusercontent.com/assets/2165184/8986837/13270de4-36ac-11e5-8e9f-d2e9dd70103c.png)

thejameskyle:
Nice, wanna open a PR for it? [Here's some examples](https://github.com/babel/babel.github.io/pulls?utf8=✓&q=is%3Apr+is%3Aclosed+%22New+User%3A%22)

AimeeKnight:
@thejameskyle Yep. I'll follow up with a PR.

PR #661, opened 2015-12-28
New user: SparkPost

AimeeKnight:
Follow up from #391


thejameskyle:
Can you run this through `svgo`?

AimeeKnight:
@thejameskyle no problem. Should be all set.

PR #668, opened 2016-01-05
Warn users that npm >= 3.3.0 is required to use babel 6.x

baer:
Babel 6's new tool `babel-doctor` [requires users to be using npm >= 3.3](https://github.com/babel/babel/blob/e75f2ef5bb340e71ec413709eaa1311beed2a834/packages/babel-cli/src/babel-doctor/rules/npm-3.js). It is not clear in any docs or blog posts whether or not Babel has officially dropped support for npm 2.x but if this is true it should be very prominently displayed in the docs.

Thanks for all the great work!

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -57,7 +57,11 @@ custom_js_with_timestamps:
 
       <p>Great! You've configured Babel but you haven't made it actually <em>do</em> anything. Create a <a href="/docs/usage/babelrc">.babelrc</a> config in your project root and enable some <a href="/docs/plugins">plugins</a>.</p>
       <p>Assuming you installed the <a href="https://babeljs.io/docs/plugins/preset-es2015/">ES2015 Preset</a>, you can do this with <pre>echo '{ "presets": ["es2015"] }' > .babelrc</pre></p>
-      
+
     </div>
+
+    <p>
+      <strong>Note</strong>: Running a Babel 6.x project using npm 2.x can cause performance problems because of the way npm 2.x installs dependencies. This problem can be eliminated by either switching to npm 3.x or running npm 2.x with the <a href="https://docs.npmjs.com/cli/dedupe">dedupe</a> flag. To check what version of npm you have run <pre>npm --version</pre>
+    <p>
   </div>
 </div>


thejameskyle:
We don't require npm 3, but npm 3 improves the performance dramatically

hzoo:
I think we can merge it but its true we don't require it (maybe just change it to say that its recommended?). We do a check for it in babel-doctor as well https://github.com/babel/babel/blob/master/packages/babel-cli/src/babel-doctor/rules/npm-3.js

baer:
Well, the big question for me is whether or not Babel 6.x will support npm@2 specific issues. If it works for now but new issues will be marked "wont fix" then the message should be stronger than recommend. What do you think?

hzoo:
Hmm not sure, but I suppose we won't support it then. npm 2 with `npm dedupe` should be ok as well but yeah we recommend npm 3. People are noticing that the es2015 preset is like 200mb but that's with npm 2 without dedupe; with npm 3 is like 200kb.

thejameskyle:
If something was actually broken in Babel only for npm 2 we would fix it. Obviously we can't fix the problems in npm 2 itself.

hzoo:
Ok yeah the warning is mostly about the performance of npm 2 - it should still work fine though

baer:
If @thejameskyle is speaking on behalf of the project then the message should probably be the following:

*Note*: Running a Babel@6 project using npm@2 can cause performance problems because of the way `npm@2` install's dependencies. This problem can be eliminated by either switching to npm@3 or running npm@2 with the dedupe flag. To check what version of npm you have run `npm --version`

How does that sound?

hzoo:
If we want we can do a short sentence explaining it but we don't have to (or link to somewhere that mentions what dedupe does)
Yeah that sounds fine. If you want you can mention the command (`npm dedupe`) instead.

baer:
I'll update this PR and link to npm dedupe in the message.

baer:
@hzoo - Updated the markup :)

docs/setup.md
```diff
@@ -56,7 +56,11 @@ custom_js_with_timestamps:
       <h2><span class="step-no">4</span> Create <code>.babelrc</code> configuration file</h2>
 
       <p>Great! You've configured Babel but you haven't made it actually <em>do</em> anything. Create a <a href="/docs/usage/babelrc">.babelrc</a> config in your project root and enable some <a href="/docs/plugins">plugins</a>.</p>
-      
+
     </div>
+
+    <p>
+      <strong>Note:</strong>: Running a Babel 6.x project using npm 2.x can cause performance problems because of the way npm 2.x install's dependencies. This problem can be eliminated by either switching to npm 3.x or running npm 2.x with the <a href="https://docs.npmjs.com/cli/dedupe">dedupe</a> flag. To check what version of npm you have run `npm --version`
```
hzoo: Typos: there are 2 colons (`:`). I think `install's dependencies` should be `installs dependencies` right?

Since this is html you will have to use `<pre>` instead of backticks for `npm --version`

baer:
Fixed

hzoo:
Hm it says there's merge conflicts?

baer:
:/ I just rebased onto the latest master - go go gadget green button. Sorry bout that I should have checked.

hzoo:
ok thanks!

PR #678, opened 2016-01-11
Documentation for babel-plugin-transform-runtime

trodrigues:
Based on https://github.com/babel/babel.github.io/blob/862b43db93e48762671267034a50c30c00e433e2/docs/usage/runtime.md

Not sure if the documentation from Babel 5 is still 100% accurate, but from what I could understand it seems to be.

Related to https://github.com/babel/babel.github.io/issues/674 and https://github.com/babel/babel.github.io/issues/491

Also added this to the main repo at https://github.com/babel/babel/pull/3258

### DIFF
--- a/docs/plugins/transform-runtime.md
+++ b/docs/plugins/transform-runtime.md
@@ -6,6 +6,8 @@ permalink: /docs/plugins/transform-runtime/
 package: babel-plugin-transform-runtime
 ---
 
+# babel-plugin-transform-runtime
+
 Externalise references to helpers and builtins, automatically polyfilling your code without polluting globals.
 
 ## Why?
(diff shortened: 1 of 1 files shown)


trodrigues:
Updated code examples according to @zloirock feedback on https://github.com/babel/babel/pull/3258

trodrigues:
My [other PR](https://github.com/babel/babel/pull/3258) where I added these same docs to the plugin's readme got merged, and I see that some docs were added in the meanwhile to http://babeljs.io/docs/plugins/transform-runtime/

Should I try and merge this PR with the current documentation on the website or would it be more desirable to keep it short?

hzoo:
If we're merging that, then yeah that's fine (go ahead). I think we need to figure out how to sync the readmes from the repo to here so we don't have to update both.

hzoo:
Merged manually in https://github.com/babel/babel.github.io/commit/be27b1e57194a239d07b77ee913f1ef04e0fe7b9

trodrigues:
Ah cool, I was gonna do that now but I see you've already done it. Thanks :)

PR #681, opened 2016-01-14
Inline install instructions setup

hzoo:
Just adding the cli commands in setup if that makes it clear

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -40,7 +40,7 @@ custom_js_with_timestamps:
       <blockquote class="babel-callout babel-callout-info">
         <h4>Note</h4>
         <p>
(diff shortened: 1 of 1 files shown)


jmm:
Typo `babel-preset-2015 => babel-preset-es2015`. Also, I don't think we should be encouraging people to `npm install babel-preset-es2015` -- they should at least be using the options to save it as a dep (though I never do that -- I just edit `package.json`).

hzoo:
Cool thanks for the catch haha

jmm:
np. Hopefully people can figure out for themselves if they need `--save|--save-dev`.

hzoo:
Yeah hard to say but I assume most of the time people would do `--save-dev`. If they are making a library/dependency on it then they should know `--save`. Otherwise I would just do `npm i babel-preset-es2015 -D` haha

PR #684, opened 2016-01-15
Add message about usage of babel-register in production

drewhamlett:
according to  https://phabricator.babeljs.io/T6940

### DIFF
--- a/_includes/tools/babel_register/usage.md
+++ b/_includes/tools/babel_register/usage.md
@@ -11,6 +11,13 @@ All subsequent files required by node with the extensions `.es6`, `.es`, `.jsx` 
   </p>
 </blockquote>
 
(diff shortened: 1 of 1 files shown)


hzoo:
Hmm I thought we had this already but it looks like it was for [babel-node](http://babeljs.io/docs/usage/cli/#babel-node). Thanks! I guess we could recommend they compile to a separate directory here but this is fine too

drewhamlett:
Agree, we need additional instruction on what to do instead.  I actually didn't know this either.  I  have ran babel-register(then 6to5/register) in production for a few server side Node projects for over a year.  Didn't seem to run into any issues except maybe a little higher memory usage.  I've switched to using separate directory since though.

PR #688, opened 2016-01-17
Get rid of REPL keyboard hijacking annoyances.

bevacqua:
I use the Babel REPL website a lot, and I also use <kbd>⌘</kbd> + <kbd>L</kbd> a lot to focus the browser address bar. In the Babel REPL site I can't because the editor hijacks this shortcut, as well as the browser Find command under <kbd>⌘</kbd> + <kbd>F</kbd>.

Neither of these features make sense for a REPL that's most typically used to play around with small snippets of code.


hzoo:
Sounds fine to me? @thejameskyle

PR #691, opened 2016-01-19
Fix math in Modules example

tiemevanveen:
This line: 

```js
alert("2π = " + exp(pi, e));
```
* Result = `23.14070064895278` == e^π
* `2π`  should be `e^π`
* `exp` expects only one argument

### DIFF
--- a/docs/learn-es6.md
+++ b/docs/learn-es6.md
@@ -414,7 +414,7 @@ export default function(x) {
 ```js
 // app.js
 import exp, {pi, e} from "lib/mathplusplus";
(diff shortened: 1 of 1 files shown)


tiemevanveen:
Btw, why is `alert` used here and not `console.log` like the other samples? 
Because all module-related-examples always use `alert`?

stevemao:
Should probably just use expressions in the example instead of `alert` or `console.log`.

thejameskyle:
It can be `console.log` if you want to update it

PR #692, opened 2016-01-19
Object.assign example Point -> shape

tiemevanveen:
Shouldn't `Point ` be a `shape` or a `rect` or something? And with a lowercase letter? 
Or am I missing the point (pun not intended, haha) here?


loganfsmyth:
The `origin` property in this example is meant to be a static property on the `Point` constructor, so this should make sense.

tiemevanveen:
Okay. So just to be clear. This `Object.assign` adds a static property `origin` to `Point` and that also happens to be a `Point`?

loganfsmyth:
Yes, so it's essentially doing `Point.origin = new Point(0, 0);` so that when doing math you can use the static `Point.origin` point as the origin.

tiemevanveen:
Check. Thanks for explaining.

PR #699, opened 2016-01-24
Fix external helpers plugin package link

TheDeveloperXYZ:
This fixes missing Github link and leads to the correct and newest external helpers plugin in npm.


hzoo:
Thanks

PR #700, opened 2016-01-26
add docs for transform-react-jsx-source

hzoo:


docs/plugins/transform-react-jsx-source.md
```diff
@@ -0,0 +1,44 @@
...
+
+```javascript
+// this/file.js
+<sometag />
+```
+
+**Out**
+
+```javascript
+var _jsxFileName = "this/file.js";
```
frantic: IMHO it's worth noting that this path will be absolute

hzoo: Yeah good catch

FDiskas: Can this be removed? I'm developing some react module and this is the pain to see my name in the dist files.

PR #705, opened 2016-02-02
Update react preset doc to include displayName transform

thomasboyt:
The react preset [includes babel-plugin-transform-react-display-name](https://github.com/babel/babel/blob/master/packages/babel-preset-react/index.js#L7) but this wasn't reflected in the doc page.


hzoo:
Thanks! good catch

PR #706, opened 2016-02-04
Minor gramma fix

stevemao:
anything => something
Make it sound more natural.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -8,8 +8,7 @@ redirect_from:
  - /docs/usage/modules/
 ---
 
-Out of the box Babel doesn't do anything. In order to actually do anything to your code
-you need to enable plugins. Don't know where to start? Check out some of the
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -8,7 +8,7 @@ redirect_from:
  - /docs/usage/modules/
 ---
 
-Out of the box Babel doesn't do anything. In order to actually do anything to your code
+Out of the box Babel doesn't do anything. In order to actually do something to your code
```
hzoo: Maybe we should just remove that and just do something like `Out of the box Babel doesn't do anything; you need to enable plugins to transform your code.`

stevemao: `use plugins` instead of `enable plugins`? "enable"  sounds like it's already there in babel-core but "use" means to use any plugin on npm.

hzoo: Ok lets go with that

thejameskyle: :+1:

stevemao:
Done :)

hzoo:
ok thanks

PR #713, opened 2016-02-07
Go to blob rather than edit page

kittens:
Otherwise you get an unfriendly message saying you need to fork the repo and you don't even get to see the file.


vjeux:
:+1:

PR #722, opened 2016-02-13
Make code block render in babel-register usage docs

psalaets:
# Changes

- Added newline above es6 example

# Why

Newline needed so it renders as code block and not as inline code.

It seems that [Github's markdown rendering](https://github.com/babel/babel.github.io/blob/master/_includes/tools/babel_register/usage.md) works without this newline but jekyll doesn't in the [live site](http://babeljs.io/docs/setup/#babel_register).


hzoo:
Thanks for catching that!

PR #724, opened 2016-02-14
Add new user: Transloadit

kvz:


thejameskyle:
Thanks http://babeljs.io/users/#Transloadit

kvz:
:sparkling_heart:

PR #728, opened 2016-02-17
Correct the helpers path in Jasmine

ymhuang0808:
According to the [example project](https://github.com/piecioshka/test-jasmine-babel/blob/master/test/jasmine.json) and [Jasmine code](https://github.com/jasmine/jasmine-npm/blob/793b7bb4f3ba4f420c7183b825ec477ebcf6eb52/lib/jasmine.js#L82), the root path of the helpers is not `spec/support` but `spec/`.

### DIFF
--- a/_includes/tools/jasmine/usage.md
+++ b/_includes/tools/jasmine/usage.md
@@ -3,7 +3,7 @@ In your `spec/support/jasmine.json` file make the following changes:
 ```json
 {
   "helpers": [
-    "../../node_modules/babel-register/lib/node.js"
+    "../node_modules/babel-register/lib/node.js"
   ]
 }
 ```


joeldenning:
This pull request really helped me, I was really struggling to set up the jasmine babel. However, for me, instead of `../node_modules/babel-register/lib/node.js` I actually had to do `../node_modules/babel-core/register.js` to get babel to actually work. Not sure why, but if that's the case for everyone else maybe this pull request should be updated to reflect that?

hzoo:
Ah thanks a lot for the fix!

Regarding babel-core/register vs babel-register - that is weird since all it does is run babel-register now.

https://github.com/babel/babel/blob/master/packages/babel-core/register.js

joeldenning:
Turns out that when I was trying the `node_modules/babel-register/lib/node.js`, I had `../..` instead of `../`, which is why it wasn't working. Although, it might be better to still do the `babel-core/register.js`, so that you don't have to install babel-register manually.

PR #729, opened 2016-02-17
Stack tools on setup page

pyrotechnick:
This change causes the tools, contained in [button groups](http://getbootstrap.com/components/#btn-groups), on the [setup page](http://babeljs.io/docs/setup/) to stack horizontally on tablet and desktop devices; saving precious vertical space.

| Small devices - Tablets (≥768px) | Medium devices - Desktops (≥992px) |
|---------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|
| Small devices Tablets (≥768px) | Medium devices Desktops (≥992px) |

The behaviour on [extra small devices](http://getbootstrap.com/css/#grid) remains the same.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -15,16 +15,20 @@ custom_js_with_timestamps:
   <div class="step-wizard">
     <div class="step">
       <h2><span class="step-no">1</span> Choose your tool</h2>
+      
+      <div class="row">
+        {% for tool in site.data.tools %}
+          <div class="col-xs-12 col-sm-6 col-md-4">
+            <h5>{{tool.name}}</h5>
 
-      {% for tool in site.data.tools %}
-        <h5>{{tool.name}}</h5>
-
-        <div class="btn-group">
-          {% for item in tool.items %}
-            <div class="btn btn-default" data-name="{{item[0]}}">{{item[1]}}</div>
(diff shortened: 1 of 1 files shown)


hzoo:
Maybe a picture would be useful here?

pyrotechnick:
That was fast!

I've amended the PR with screenshots.

The `btn-group`s look far better with `btn-group-justified` added but I'm not sufficiently versed in Jekyll/Liquid to pull off the likes of [_.chunk](https://lodash.com/docs#chunk).

hzoo:
Ok thanks. @thejameskyle?

kittens:
This seems really confusing to me. There's a lot of noise and it further reinforces some of @thejameskyle's original complaints of the design that it looks like you have to select an item from each button set rather than them all being a part of the same radio.

thejameskyle:
I still think this should be a giant `<select>` like this: (although using a selectbox library that makes it look prettier)

screen shot 2016-03-05 at 10 31 36 am

Been sitting in https://github.com/babel/babel.github.io/tree/redesign forever

hzoo:
Thanks for the PR but I think we're going to close for now

PR #734, opened 2016-02-21
Clarify polyfill dependency

bclinkinbeard:
I was initially confused by the existing language, thinking I needed the polyfill any time I used `babel-register`. This attempts to clarify things, based on my rudimentary understanding.


hzoo:
Thanks :+1:

PR #736, opened 2016-02-22
Outdated info

lifeBalance:


hzoo:
Actually this is the updated one. see https://github.com/babel/babel.github.io/pull/601

Thanks anyway!

lifeBalance:
Well I was building a isomorphic app, and you know, to bootstrap the react app I installed and required `babel-register` in my `server.js`; starting the server was followed by `SyntaxError: Unexpected token <`.
So resorting to the line `require("babel-core/register");` before requiring the component itself did the trick.

hzoo:
It should be the same for either (it does the same thing)

lifeBalance:
They weren't doing the same in my case. I did find the answer in this [issue](https://github.com/gulpjs/gulp/issues/1359) so I decided to help out with the docs. If you want you can check [my code here](https://github.com/lifeBalance/todo-isomorphic) 
Or I can try to reproduce again the error...

PR #738, opened 2016-02-24
Invalid URL

bpneal:
URL contains a trailing slash which results in a 404


hzoo:
Hmm wonder if it was working before - thanks!

PR #746, opened 2016-03-05
Link to Dan's repl

hzoo:
(at least until we bring it back or merge)

cc @thejameskyle


kittens:
Why can't we just merge #613?

Daniel15:
I just updated #613, now it's in a state that it can be safely merged :) We should probably do that and just close this PR.

PR #750, opened 2016-03-06
add docs for dup keys plugin

hzoo:
Wasn't added before

@AgentME


AgentME:
Thanks! Looks good to me.

hzoo:
Cool!

PR #752, opened 2016-03-09
make repl work in ES5

bergus:
I guess we want the REPL to work in ES5 environments?
So if we want to write it in ES6, it should be transpiled :-)

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -149,13 +149,13 @@
     });
 
     return {
-      get() {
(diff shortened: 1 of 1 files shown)


thejameskyle:
heh

Daniel15:
Oops! Good catch, thanks @bergus. I was originally using Babel to compile the REPL code, but in the end I stopped doing that. Looks like I forgot to update this part of the code.

PR #761, opened 2016-03-17
Update the docs of transform-react-inline-elements

oliviertassinari:
The [output changed](https://github.com/babel/babel/blob/3232c30f61c2d321f7813ec406d8b28392bcadc0/packages/babel-plugin-transform-react-inline-elements/test/fixtures/inline-elements/component-with-props/expected.js) since this example was written.


hzoo:
Alright thanks!

PR #772, opened 2016-03-30
An more realistic example on homepage

hzoo:
I know we may not want to go for this but figured I could do it for a suggestion. (The idea is to explain the simplest way to use it without clicking on anything else)

Current:
screen shot 2016-03-29 at 11 02 56 pm

Proposed (or something similar):
screen shot 2016-03-30 at 12 35 02 pm
screen shot 2016-03-30 at 12 35 10 pm

@jmm @loganfsmyth

### DIFF
--- a/index.html
+++ b/index.html
@@ -29,20 +29,31 @@ third_party_js:
   <div class="row featurette text-center">
     <h2 class="text-center">Babel transforms your JavaScript</h2>
 
+    <h3 class="lead text-center">
+{% highlight javascript %}
+npm install --save-dev babel-cli babel-preset-es2015
+./node_modules/.bin/babel index.js --presets es2015
+{% endhighlight %}
+    </h3>
+
     <div class="col-lg-6">
       <h3>You put JavaScript in</h3>
-      <div class="lead">
+      <div class="lead text-left">
 {% highlight javascript %}
-myJavaScript("foobar");
+// original index.js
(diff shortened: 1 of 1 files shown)


weslleyaraujo:
Nice improvement!
How about using those lines `npm install..` /  `./node_modules/.bin...` in the same block?

I think it would be more readable :)

index.html
```diff
@@ -29,20 +29,33 @@
   <div class="row featurette text-center">
     <h2 class="text-center">Babel transforms your JavaScript</h2>
 
+    <h3 class="lead text-center">
+      {% highlight javascript %}
+      npm install --save-dev babel-cli babel-preset-es2015
+      {% endhighlight %}
+      {% highlight javascript %}
+      ./node_modules/.bin/babel index.js --presets=es2015
```
loganfsmyth: Do we want to encourage usage  of the `=` form or the `--presets es2015` form?

hzoo:
I actually moved it into 2 boxes since I thought it was better but ok we can try that too.

index.html
hzoo: I have no idea haha.. What looks clearer?

If we wanted conciseness I would of done `npm i -D babel-{cli, preset-es2015}`

hzoo:
Updated with new pictures

weslleyaraujo:
looks awesome :+1:

PR #776, opened 2016-04-02
removed make:build as bootstrap is being tracked

sarbbottam:
As [`bootstrap` is being tracked and checked in](https://github.com/babel/babel.github.io/tree/master/_sass/bootstrap), I guess `make:build` is not required, thus ...

* removed package.json:scripts.build
* added package.json:scripts.serve
* removed the bootstrap-sass from package.json:devDependencies
* in fact removed package.json:devDependencies

Please close the PR if it sound inappropriate. Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "description": "",
   "main": "",
   "scripts": {
+    "start": "make start",
     "build": "make build",
     "test": "echo \"Error: no test specified\" && exit 1"
   },


package.json
```diff
@@ -5,12 +5,9 @@
   "description": "",
   "main": "",
   "scripts": {
-    "build": "make build",
+    "serve": "make serve",
```
hzoo: we could do `start` so we can use to `npm start` instead

hzoo:
cc @thejameskyle

package.json
thejameskyle: yeah makes sense

thejameskyle:
I kinda want to keep it just so bootstrap can be updated easier

sarbbottam:
@thejameskyle @hzoo could you take another look?

hzoo:
Thanks

PR #778, opened 2016-04-03
add more plugins + link them to plugin pages, fix typo, misc

hzoo:
Add some more links?

screen shot 2016-04-02 at 10 50 12 pm

screen shot 2016-04-02 at 10 59 43 pm

I kind of want to switch the babel blueprint picture out with just an example plugin like on astexplorer?

### DIFF
--- a/index.html
+++ b/index.html
@@ -144,7 +144,7 @@ echo '{ "presets": ["es2015"] }' > .babelrc
 // create a babel config with your presets
 echo '{ "presets": ["es2015", "react"] }' > .babelrc
 {% endhighlight %}
-      <p>Learn about <a href="https://facebook.github.io/jsx/">JSX</a> and <a href="http://flowtype.org/">Flow</a>
+      <p>Learn about <a href="https://facebook.github.io/react/docs/jsx-in-depth.html">JSX</a> and <a href="http://flowtype.org/">Flow</a>
       </p>
     </div>
   </div>
--- a/index.html
+++ b/index.html
@@ -142,7 +142,7 @@ echo '{ "presets": ["es2015"] }' > .babelrc
       <p>Babel can convert JSX syntax and strip out type annotations. Check out our <a href="http://babeljs.io/docs/plugins/preset-react/">React preset</a> to get started. Use it together with the <a href="https://github.com/babel/babel-sublime">babel-sublime</a> package to bring syntax highlighting to a whole new level.</p>
 {% highlight javascript %}
 // create a babel config with your presets
-echo '{ "presets": ["es2015"] }' > .babelrc
+echo '{ "presets": ["es2015", "react"] }' > .babelrc
 {% endhighlight %}
       <p>Learn about <a href="https://facebook.github.io/jsx/">JSX</a> and <a href="http://flowtype.org/">Flow</a>
       </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -129,8 +138,14 @@ <h3 class="lead text-center">
     </div>
 
     <div class="col-md-7">
-      <h2>JSX and React</h2>
-      <p>Babel has support for JSX and Flow. Check out our <a href="http://babeljs.io/docs/plugins/preset-react/">React preset</a> to get started. Use it together with the <a href="https://github.com/babel/babel-sublime">babel-sublime</a> package to bring syntax highlighting to a whole new level.</p>
+      <h2>JSX and Flow</h2>
+      <p>Babel can convert JSX syntax and strip out type annotations. Check out our <a href="http://babeljs.io/docs/plugins/preset-react/">React preset</a> to get started. Use it together with the <a href="https://github.com/babel/babel-sublime">babel-sublime</a> package to bring syntax highlighting to a whole new level.</p>
+{% highlight javascript %}
+// create a babel config with your presets
+echo '{ "presets": ["es2015"] }' > .babelrc
```
jmm: This is meant to be one of the following, I think?:

```console
echo '{ "presets": ["es2015", "react"] }' > .babelrc
echo '{ "presets": ["react"] }' > .babelrc
```

hzoo: Yes! I just copied it from the other one..

index.html
```diff
@@ -129,8 +138,14 @@ <h3 class="lead text-center">
...
+{% endhighlight %}
+      <p>Learn about <a href="https://facebook.github.io/jsx/">JSX</a> and <a href="http://flowtype.org/">Flow</a>
```
jmm: Even though the React docs are a mess, I'm thinking it'd be better to link JSX to https://facebook.github.io/react/docs/displaying-data.html#jsx-syntax or https://facebook.github.io/react/docs/jsx-in-depth.html, since the spec one here is more geared toward implementers. Unfortunate though that it does a better job of just showing an example right off the bat than the React docs.

hzoo: Ok using https://facebook.github.io/react/docs/jsx-in-depth.html.

Can't wait to use the squash feature

jmm:
LGTM with a couple of comments. I think replacing the blueprint graphic with a simple example plugin would be fine.

index.html
jmm: :+1:

thejameskyle:
It would be nice if we could update the homepage to effectively be a tutorial. The sections on es2015 and react/flow could just be steps in that setup process.

screen shot 2016-04-03 at 12 09 41 pm

I don't like the whole `./node_modules/.bin/babel` and `echo '{ "presets": ["es2015"] }' > .babelrc` things because they aren't taking the opportunity to explain what is going on

hzoo:
Tutorial sounds pretty awesome! We could do an interactive thing on how to setup a .babelrc right on the homepage?

From [twitter](https://twitter.com/secoif/status/716435469018472448) - "everything could be an embedded repl" sounds interesting as well.

hzoo:
Current updates:

screen shot 2016-04-03 at 4 29 23 pm
screen shot 2016-04-03 at 4 29 36 pm
screen shot 2016-04-03 at 4 29 42 pm
screen shot 2016-04-03 at 4 29 50 pm

thejameskyle:
awesome!

PR #781, opened 2016-04-03
scroll to the installation step when a tool is selected

sarbbottam:
I find it annoying to scroll to reach the installation/usage step when I select a tool, I tried to address that in this PR, scroll to the installation step when a tool is selected, please refer the behavior in the attached gif.

babel-setup


hzoo:
Oh ok cool! Thanks for the PR + the gif!

thejameskyle:
Nice

PR #783, opened 2016-04-03
added instruction to install ruby as prerequisites

### LINKED ISSUE #782: Add instructions on local setup
Add some instructions on local setup if stuff isn't installed

https://github.com/babel/babel.github.io/issues/258#issuecomment-111142413
> $ # install ruby
$ gem install bundler # if ruby hasn't been installed
$ bundle install # if you haven't already
$ gem install jekyll-redirect-from
$ jekyll s

sarbbottam:
fixed #782 

@hzoo could you take a look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,5 +5,6 @@ This is the source for the [babeljs.io](https://babeljs.io) website, powered by 
 This repo also contains babel's [source documentation](https://github.com/babel/babel.github.io/tree/master/docs).
 
 ### Setup
-
-To start the web server on your machine, check out this repo and run `make serve`.
+* [Please install ruby](https://www.ruby-lang.org/en/documentation/installation/), if not already installed.
+  * Use `which ruby` or `ruby --version`, to verify if ruby is installed.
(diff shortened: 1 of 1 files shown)


hzoo:
Looks good?

So if bundler isn't installed it we print a message. I'm assuming if you install ruby it comes with `gem`.

And we may need to use `make build`?

sarbbottam:
>I'm assuming if you install `ruby` it comes with `gem`.

I am not sure, `ruby` and `gem` came pre-installed in mac. However, we could add an instruction about [`gem`](https://rubygems.org/pages/download).

>And we may need to use `make build`?

hmm, not sure, since `ruby` and `gem` is a system/user level install and not local to the repo.

sarbbottam:
Also, [current `make build` seems to be redundant](https://github.com/babel/babel.github.io/pull/776).

hzoo:
Right, I was referencing make build in the makefile, but I see you made https://github.com/babel/babel.github.io/pull/776

PR #785, opened 2016-04-06
Update transform-react-constant-elements.md

oliviertassinari:
Simplify the examples to be more explicit.


hzoo:
Looks good. Maybe we should make a few examples - with es6 class and with the stateless functional components? Thoughts?

hzoo:
We can do it in another PR (or I'l do it)

PR #786, opened 2016-04-06
Update transform-react-inline-elements.md

oliviertassinari:
Simplify the examples to be more explicit.


hzoo:
Thanks!

hzoo:
And we could mention the deopts here as well?

oliviertassinari:
@hzoo Looks like you have already added the *deopt*, awesome :+1:.

PR #791, opened 2016-04-07
Fix formatting in webpack usage example "babel-loader"} ~> "babel-loader" }

alexnormand:
Fix formatting in webpack usage example "babel-loader"} ~> "babel-loader" }


hzoo:
Thanks

PR #802, opened 2016-04-15
Integrate Disqus on doc pages.

STRML:
After some discussion on Slack I thought it would be really nice, given that Babel doesn't have a wiki and there's ton of useful information doc readers would want, but is unfortunately scattered all over Stackoverflow, Phabricator, PRs, Medium articles & blog posts. Perhaps this would help centralize it a bit.

A useful analogue is how the comments sections on some projects (PHP, Redis, etc.) are sometimes the most useful sources of information.

Awkward if/else on 21-25 is to get around Liquid's lack of ternary operators.

I registered `babeljs.disqus.com` - happy to transfer it to whomever. I noticed `babel.disqus.com` was taken but I didn't want to take the risk that it was taken by someone unrelated to the project.

### DIFF
--- a/_layouts/docs.html
+++ b/_layouts/docs.html
@@ -16,4 +16,24 @@ custom_js_with_timestamps:
       </div>
     </div>
   </div>
+  <div id="disqus_thread"></div>
+  <script>
+    {% if page.package %}
+    {% assign identifier = page.package %}
+    {% else %}
+    {% assign identifier = page.permalink %}
+    {% endif %}
+    var disqus_config = function () {
+      var origin = window.location.protocol + "//" + window.location.hostname +
+        (window.location.port ? ':' + window.location.port : '');
+      this.page.url = origin + "{{page.permalink}}";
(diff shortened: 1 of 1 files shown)


STRML:
screen shot 2016-04-15 at 3 29 35 pm

STRML:
Alternatively we could integrate the existing Discourse setup pretty easily.

This is [how it works](https://meta.discourse.org/t/embedding-discourse-comments-via-javascript/31963):

1. We embed JS on each page,
2. The first time Discourse sees it, it creates a thread for the page.
3. Users can view comments right on the page. To post a comment, unfortunately, you need to go to the actual forum thread - but it could start driving a lot more traffic to Discourse (and away from GitHub/Phabricator/Slack) which would be great.

hzoo:
That would be interesting as well..

STRML:
@hzoo see #875

STRML:
We dropped this for Discourse in #875.

PR #803, opened 2016-04-17
update polyfill install command to use save-dev option

shogunsea:
just to be consistent with `babel-cli` and `babel-preset-es2015` install command


hzoo:
Thanks!

PR #808, opened 2016-04-20
Added section about lookup behavior.

jeffijoe:


hzoo:
Thanks 😄

jmm:
Thanks @jeffijoe. We also need to document how [this behavior is affected by `options.filename`](https://github.com/babel/babel.github.io/issues/539) for API calls. It'd probably also be good to mention `options.babelrc` on this page.

PR #812, opened 2016-04-28
Update AVA documentation

maur8ino:
Hello,

I've updated the AVA documentation, pointing the "more information" link to the AVA recipe "Configuring Babel".

@sindresorhus the AVA default Babel configuration, still includes the `"espower"` and `"transform-runtime"`plugins? Because I didn't see the being mentioned anywhere in the recipe.

Thanks!

### DIFF
--- a/_includes/tools/ava/usage.md
+++ b/_includes/tools/ava/usage.md
@@ -2,22 +2,8 @@ AVA ships with ES2015 support built-in using Babel 6 under the hood, so
 you can write test using ES2015 syntax right away.
 
 The AVA default Babel configuration includes the `"es2015"` and `"stage-2"`
-presets and the `"espower"` and `"transform-runtime"` plugins, corresponing to:
-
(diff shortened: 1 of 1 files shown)


sindresorhus:
See: https://github.com/sindresorhus/ava/blob/master/docs/recipes/babelrc.md#notes

maur8ino:
so, if I'm not mistaken, other those two, the default configuration adds also `"ava-throws-helper"` plugin, right?

sindresorhus:
Yes, but not sure why you would need to know that. AVA adds those regardless of custom Babel config.

maur8ino:
@sindresorhus ah ok, I haven't noticed that "**_always_**" in https://github.com/sindresorhus/ava/blob/master/docs/recipes/babelrc.md#notes hehe 😄

danez:
Thanks.

PR #813, opened 2016-04-28
replace all http URLs with https equivalents

graingert:


graingert:
Note this also fixes the 404 error pages notworking on https://babeljs.io/

adamchainz:
*equivalents

graingert:
Woops

hzoo:
the website links should probably be relative so you don't have to type it all out - thanks

PR #816, opened 2016-05-02
Switch to kramdown

hzoo:
Fixes https://github.com/babel/babel.github.io/issues/709

Looks like markdown is fine...

screen shot 2016-05-01 at 10 50 37 pm

updated deps, changed to kramdown


Gemfile
```diff
@@ -2,3 +2,4 @@ source 'https://rubygems.org'
 
 gem 'jekyll'
 gem 'jekyll-redirect-from'
+gem 'jekyll-paginate'
```
hzoo: updated to jekyll 3? had to add this since it's not part of it by default

queenfoxylady: No more negative influences please grown ups should be adults

queenfoxylady:
FYI that's why no one should have broke into my where houses for my inventory lists because it wasnt organized at all 3 places with different uses of material and now everything's all mixed up..games and books go together and contract and sha go together and phone numbers and businesses and new projects ext ect ect

PR #817, opened 2016-05-04
Update mocha setup instruction

frantic1048:
Setup instruction for Mocha is a little outdated. Here's an update for it.

### DIFF
--- a/_includes/tools/mocha/install.md
+++ b/_includes/tools/mocha/install.md
@@ -1,3 +1,3 @@
 ```sh
-$ npm install --save-dev babel-register
+$ npm install --save-dev babel-core
 ```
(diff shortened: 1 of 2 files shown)


loganfsmyth:
Unfortunately, this is the opposite of what we recommend. `babel-core/register` and `babel-register` point at the same thing, but `babel-register` is the newer and recommended approach.

frantic1048:
Sorry for my careless. I was setting up babel for mocha yesterday, I have installed `babel` and tried `babel/register` first, which not work properly. After that, I installed `babel-register` and removed `babel` and caused an error. Finally, I tried `babel-core/register`, everything goes well. So I misunderstood the instruction was outdated.

I configured a new project with `babel-register` just now. Everything works well. The error I met should cause by my removing `babel` step.

loganfsmyth:
No problem, it happens. Glad you got it working.

PR #819, opened 2016-05-05
Add trasformFromAst documentation

sheerun:


hzoo:
Thanks, we should add a description too

PR #820, opened 2016-05-09
correct command

hootlex:
quotes should be removed since the output must be a JSON object.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -56,7 +56,13 @@ custom_js_with_timestamps:
       <h2><span class="step-no">4</span> Create <code>.babelrc</code> configuration file</h2>
 
       <p>Great! You've configured Babel but you haven't made it actually <em>do</em> anything. Create a <a href="/docs/usage/babelrc">.babelrc</a> config in your project root and enable some <a href="/docs/plugins">plugins</a>.</p>
-      <p>Assuming you installed the <a href="https://babeljs.io/docs/plugins/preset-es2015/">ES2015 Preset</a>, you can do this with <pre>echo { "presets": ["es2015"] } > .babelrc</pre></p>
+      <p>Assuming you have installed the <a href="https://babeljs.io/docs/plugins/preset-es2015/">ES2015 Preset</a>, in order to enable it your <code>.babelrc</code> file should look like this:
+      
+      <code>
+      {
+          presets: [["es2015"]]
+      }
+      </code>
 
       <p>
         <strong>Note</strong>: Running a Babel 6.x project using npm 2.x can cause performance problems because of the way npm 2.x installs dependencies. This problem can be eliminated by either switching to npm 3.x or running npm 2.x with the <a href="https://docs.npmjs.com/cli/dedupe">dedupe</a> flag. To check what version of npm you have run <pre>npm --version</pre>
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -56,17 +56,17 @@ custom_js_with_timestamps:
       <h2><span class="step-no">4</span> Create <code>.babelrc</code> configuration file</h2>
 
       <p>Great! You've configured Babel but you haven't made it actually <em>do</em> anything. Create a <a href="/docs/usage/babelrc">.babelrc</a> config in your project root and enable some <a href="/docs/plugins">plugins</a>.</p>
-      <p>Assuming you have installed the <a href="https://babeljs.io/docs/plugins/preset-es2015/">ES2015 Preset</a>, in order to enable it you have to define it in your <code>.babelrc</code> file, like this:
+      <p>Assuming you have installed the <a href="https://babeljs.io/docs/plugins/preset-es2015/">ES2015 Preset</a>, in order to enable it you have to define it in your <code>.babelrc</code> file, like this:</p>
       
-      <code>
-      {
-          presets: ["es2015"]
-      }
-      </code>
+      {% highlight json %}
+{
+  "presets": ["es2015"]
+}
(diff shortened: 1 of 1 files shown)


loganfsmyth:
The quotes are around the argument for the `echo` command, they don't go in the file.

hootlex:
They actually do. 
Running the command will give you this file
.babelrc

loganfsmyth:
What type of system are you running?

hootlex:
Hmm, on my mac it works as you said. The quotes don't affect the output file.
On my windows pc, using cmder it has the abode output.

I believe removing the quotes worth it since it doesn't have any conflicts on both systems.

loganfsmyth:
Unfortunately, we'd likely have the opposite problem in that case. `bash` appears to be fine without quotes, but there's no guarantee that others will. `zsh` for instance throws an error for me without the quotes.

hootlex:
That's weird, I am using `zsh` too and it seems to works fine with or without the quotes.
I've also tested it on my vagrant machine using Ubuntu 14.04 with `bash` and it works too.

I don't know what to say man :S

jmm:
Let's just get rid of that. That's what happened in #778 and I agree that there's no real point to showing it that way. It got changed to:
> Create a `.babelrc` file in your project (or use your package.json)
```js
{
  presets: []
}
```

(I'd say it'd be a bit better for that to be JSON though.)

Someone who knows you can do that and wants to create the file that way can go ahead, but I doubt it's typical, let alone necessary, so let's just eliminate shell nuances as an issue.

hootlex:
Yeah, this way it is more clear and there will be no conflicts.

jmm:
@hootlex Ok, cool, would you like to PR it? I'd say just make it similar to how it is on the index page ("Create a `.babelrc` file...") now (although with JSON, not just a JS object literal). If you want to do it I'd say just do it in this PR (and now that there's squash merging you can just push an additional commit).

hootlex:
Sure, I'll push a commit in a while.

hootlex:
I didn't manage to prettify the code using '```js', should I create a new file and include with tools.md?

I embedded the json inside `<code>` tags for now.

docs/setup.md
```diff
@@ -56,7 +56,13 @@ custom_js_with_timestamps:
       <h2><span class="step-no">4</span> Create <code>.babelrc</code> configuration file</h2>
 
       <p>Great! You've configured Babel but you haven't made it actually <em>do</em> anything. Create a <a href="/docs/usage/babelrc">.babelrc</a> config in your project root and enable some <a href="/docs/plugins">plugins</a>.</p>
-      <p>Assuming you installed the <a href="https://babeljs.io/docs/plugins/preset-es2015/">ES2015 Preset</a>, you can do this with <pre>echo '{ "presets": ["es2015"] }' > .babelrc</pre></p>
+      <p>Assuming you have installed the <a href="https://babeljs.io/docs/plugins/preset-es2015/">ES2015 Preset</a>, in order to enable it you have to define it in your <code>.babelrc</code> file, like this:
+      
+      <code>
+      {
+          presets: [["es2015"]]
```
hzoo: `"presets": ["es2015"] ` - only one []

hootlex:
sure, it was a typo

jmm:
Thanks @hootlex.

1. 
  > should I create a new file and include with items.md?

  Sorry, I'm not exactly sure what you mean. But it looks like this is the way to do a syntax highlighted block:

  ```
        {% highlight json %}
  {
    "presets": ["es2015"]
  }
        {% endhighlight %}
  ```

  (Note that the code itself is dedented.)

1. Please make the code JSON (as above). Reason: https://github.com/babel/babel.github.io/issues/455.

1. There's a missing `</p>` (line 66 or 60) that's resulting in escaped HTML rendering:

  ```html
      </div>
    </div>
  </div>
  ```

1. There's a whitespace error (trailing spaces) on line 60.

1. This isn't in your PR, it was already there, but there's a `<p>` on line 69 that should be a `</p>`. Just noticed it while reviewing this and noting it in case you want to fix it when making updates, or for one of us to fix.

1. Again, it looks like this was pre-existing, so you don't necessarily have to fix it, but the source formatting of the HTML on lines 58 & 59 doesn't quite match the rest of the file.

Please let us know if you'll have the chance to make any of these updates. If not we can do a fixup commit. Thanks!

hootlex:
1. In `setup.md` some codeblocks are included like this `{% include tools/items.md name="install" %}`. Their highlight is really nice when rendered. :)

No problem, I will fix these tags too.

jmm:
Thanks @hootlex! FYI, there was still that trailing white space error on line 60, but I took care of it (and some pre-existing issues with the source formatting) in a fixup commit. Thanks for fixing that `<p>` tag problem that was already there.

> In `setup.md` some codeblocks are included like this `{% include tools/items.md name="install" %}`. Their highlight is really nice when rendered. :)

Ah ok, thanks for explaining that to me. I'm by no means an expert on the jekyll etc. and didn't know about that.

I merged this manually in https://github.com/babel/babel.github.io/commit/5275dc999c827d23cb55625c4ab1dd2d49469a2b.

With the way the source is formatted it renders with an extra blank line at the end of the `{% highlight %}...{% endhighlight %}` code block. It can be eliminated by fiddling with the white space before `{% endhighlight %}`, but then the source formatting is ugly. If anyone happens to know a way to deal with that (an attribute, different syntax like `-%}`, ?) please let me know.

hootlex:
@jmm no problem :)

That /p was hard to see.

PR #822, opened 2016-05-13
Update polyfill.md

lolluslollus:
Slightly clearer description. It would have saved me a couple of hours.


hzoo:
It does what you added below in the code snippet so I'm not sure it's needed?

loganfsmyth:
I assume this is saying that the code snippet doesn't make it clear enough what's supposed to be added, which I could see.

PR #828, opened 2016-05-27
document systemGlobal option

guybedford:
As discussed in https://github.com/babel/babel/pull/3482#issuecomment-219545753.

### DIFF
--- a/docs/plugins/transform-es2015-modules-systemjs.md
+++ b/docs/plugins/transform-es2015-modules-systemjs.md
@@ -51,7 +51,7 @@ Add the following line to your `.babelrc` file:
 {
   "plugins": [
     ["transform-es2015-modules-systemjs", {
-      // outputs `customSystemName.register(...`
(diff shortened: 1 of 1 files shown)


docs/plugins/transform-es2015-modules-systemjs.md
```diff
@@ -41,8 +41,19 @@ $ npm install babel-plugin-transform-es2015-modules-systemjs
...
+// without options
 {
   "plugins": ["transform-es2015-modules-systemjs"]
 }
+
+// with options
+{
+  "plugins": [
+    ["transform-es2015-modules-systemjs", {
+      // outputs `customSystemName.register(...`
```
hzoo: Maybe we want to make sure people understand that it replaces `System.register`? This is fine too

guybedford: Sure I've added a clarification here.

hzoo:
thanks again!

PR #830, opened 2016-05-28
Add some more information about plugins

hzoo:
such as how its the 2nd step in babel's 3 steps for more context

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -8,13 +8,13 @@ redirect_from:
  - /docs/usage/modules/
 ---
 
-Babel is a compiler. At a high level, it has 3 stages that it runs code in: parsing code, transforming code, and code generation (like many other compliers).
+Babel is a compiler. At a high level, it has 3 stages that it runs code in: parsing, transforming, and generation (like many other compilers).
 
 > For an awesome/simple tutorial on compliers, check out [the-super-tiny-complier](https://github.com/thejameskyle/the-super-tiny-compiler), which also explains how Babel itself works on a high level.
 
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -8,7 +8,13 @@ redirect_from:
  - /docs/usage/modules/
 ---
 
-Out of the box Babel doesn't do anything. You need to use plugins to transform your code.
+Babel is a compiler. At a high level, it has 3 stages that it runs code in: parsing code, transforming code, and code generation (like many other compliers).
```
jmm: typo: compliers. Personally I don't think the repetition of *code* is necessary (e.g. could be "parsing, transforming, and generation"), but I'm not making an issue of it if you prefer it that way :smiley:

docs/plugins.md
```diff
@@ -8,7 +8,13 @@ redirect_from:
...
+
+> For an awesome/simple tutorial on compliers, check out [the-super-tiny-complier](https://github.com/thejameskyle/the-super-tiny-compiler), which also explains how Babel itself works on a high level.
+
+Now, out of the box Babel doesn't do anything. It basically acts like `const babel = code => code;` by parsing the code and then generating the same code back out again.
+
+You will need to add some plugins (the 2nd stage that transforms code) for babel to do anything.
```
jmm: I would suggest maybe saying something like this:
> You will need to add some plugins <del>(the 2nd stage that transforms code)</del> for babel to do anything (<ins>they affect stage 2, transformation</ins>).

What do you think?

It's probably ideal to consistently capitalize as *Babel* or *babel*.

hzoo:
K updated

PR #832, opened 2016-05-29
Link to a built version of the Babel 5 docs rather than the repo

hzoo:
- Forked @developit's version and fixed the urls https://developit.github.io/babel-legacy-docs/ (the relative links are broken)

http://henryzoo.com/babel.github.io/

screen shot 2016-05-29 at 7 26 05 pm

annoying that its changing to my cname though. or can pr their version and link to that


developit:
@hzoo Awesome, thanks!

PR #844, opened 2016-06-07
External helpers doc: this → self

compact:
In the source, the helpers are assigned to `self`, not `this`.
https://github.com/babel/babel/blob/5b89849f432cfc034860f495b5001039123e7754/packages/babel-helpers/src/helpers.js#L297


hzoo:
thanks!

PR #849, opened 2016-06-11
Promote trailing-function-commas to stage 3

jacobrask:
See matching pull request https://github.com/babel/babel/pull/3522


jacobrask:
This should probably be merged in now that the change was released in v6.11.0.

hzoo:
Thanks!

PR #850, opened 2016-06-12
fix checkbox creation

bergus:
Opera had problems with this. If you create a new `<input>` element, its type defaults to  `text`. You can set a `value` now, but when you alter the type afterwards to `checkbox` the value will get replaced by the string `"on"`. Which is not a known babel preset :-/
This fix just sets the `.type`  before the `.value` to make it work again.


thejameskyle:
Thanks

PR #852, opened 2016-06-18
Clarify that callback options cannot be expressed in .babelrc files

rmacklin:
Based on https://phabricator.babeljs.io/T2173, I think it's a good idea to explicitly state this in the docs


hzoo:
Sure, unless you use a js file as the config instead of .babelrc. We can also try to support .babelrc.js

PR #860, opened 2016-06-28
Add usage example for ignoring files

theSmaw:
Adds an example of how to use the `--ignore` option on the cli, as mentioned in https://github.com/babel/babel.github.io/issues/286

Also removes `#` from heading that seemed inconsistent with others.

### DIFF
--- a/docs/usage/cli.md
+++ b/docs/usage/cli.md
@@ -18,7 +18,7 @@ package: babel-cli
 
 ## babel
 
-#### Compile Files
+### Compile Files
 
 Compile the file `script.js` and **output to stdout**.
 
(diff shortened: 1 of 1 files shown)


docs/usage/cli.md
```diff
@@ -68,6 +68,14 @@ Compile the entire `src` directory and output it to the one concatenated file.
 $ babel src --out-file script-compiled.js
 ```
 
+### Ignore files
+
+Ignore spec and test files
+
+```sh
+$ babel src --out-dir lib --ignore spec.js,test.js
```
hzoo: Thanks, although it does seem weird in to me in this example because the tests are usually in a different folder than `src` so it wouldn't need to be ignored anyway?

theSmaw: I think a number of projects these days (including ours), include the .spec files adjacent to the file it is testing).

Happy to update to a different type of file to ignore?

hzoo: Nope I was just thinking about what I've been doing, makes sense!

hzoo:
Thanks 🍰

theSmaw:
Thank you!

PR #862, opened 2016-07-07
Add a note on merging env options

jesstelford:
Had to dig into the code to find out that it uses lodash's `mergeWith` for merging the options. Figured I might not be the only one searching for that clarification :)


hzoo:
Thanks!

PR #871, opened 2016-07-18
Upgrade lodash from 2.4.1 to 4.13.1 & change from // to https

peterpme:
Paul Irish says that `//` is now an anti pattern.

Primarily, I came here to update lodash in the REPL

Thanks!

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -26,7 +26,7 @@ gems:
 
 default_js:
   - "//cdnjs.cloudflare.com/ajax/libs/jquery/2.1.3/jquery.min.js"
-  - "//cdnjs.cloudflare.com/ajax/libs/lodash.js/2.4.1/lodash.min.js"
+  - "//cdnjs.cloudflare.com/ajax/libs/lodash.js/4.16.3/lodash.min.js"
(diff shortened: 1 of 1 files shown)


hzoo:
Thanks, we don't have tests for the repl so I think we would need to check all calls in the relevant js files to make sure everything still works?

peterpme:
Hey,

Thanks for the quick reply!

There are 19 counts of `_.` within this repo.

- 8 of them come from `babel.js` and include methods like `isFunctionDeclaration` which I'm not familiar with.
- 6 of them come from `docs.js`
  - `memoize`
  - `each`
- 5 of them come from `repl.js`
  - `assign`
  - `debounce`

The remaining are still supported by lodash today.

Thanks!

peterpme:
@hzoo any chance we can still make this happen? Thanks!

hzoo:
Yeah sorry 😛 , just have to lookup changes

peterpme:
Its worth noting the compiled babel.js file included has lodash 3 as a dependency

peterpme:
:cry: @hzoo if there's anything I can do to help you please let me know!

hzoo:
ok should be ok if you want to fix merge conflicts (as long as the repl works)

peterpme:
@hzoo ready when you are! Thanks again

PR #883, opened 2016-08-05
Move decorators, transform-class-properties from stage-1 to stage-2 preset docs

kevinSuttle:
See: https://github.com/babel/babel/pull/3613

TL; DR:
> TC39 advanced decorators to stage 2, w/ crucial open questions re:
> function, param, & property decorators needing good answers for
> progress.

- @BrendanEich

### DIFF
--- a/docs/plugins/preset-stage-1.md
+++ b/docs/plugins/preset-stage-1.md
@@ -10,7 +10,6 @@ This preset includes the following plugins:
 
 - [transform-class-constructor-call](/docs/plugins/transform-class-constructor-call) (Deprecated)
 - [transform-class-properties](/docs/plugins/transform-class-properties)
-- <del>[transform-decorators](/docs/plugins/transform-decorators)</del> – *disabled pending proposal update*
 - [transform-export-extensions](/docs/plugins/transform-export-extensions)
 
 And all plugins from presets:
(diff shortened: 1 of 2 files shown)


docs/plugins/preset-stage-2.md
```diff
@@ -9,6 +9,7 @@ package: babel-preset-stage-2
 This preset includes the following plugins:
 
 - [transform-object-rest-spread](/docs/plugins/transform-object-rest-spread)
+- [transform-decorators](/docs/plugins/transform-decorators)
```
hzoo: I think we can keep the message from before since the transform still isn't updated

I added https://github.com/babel/babel/pull/3626/files as a warning for the default one

docs/plugins/preset-stage-1.md
```diff
@@ -10,7 +10,6 @@ This preset includes the following plugins:
 
 - [transform-class-constructor-call](/docs/plugins/transform-class-constructor-call) (Deprecated)
 - [transform-class-properties](/docs/plugins/transform-class-properties)
```
hzoo: You can move this one as well

kevinSuttle:
Will do!

docs/plugins/preset-stage-2.md
kevinSuttle: You mean the message should now be in stage 2?

hzoo: Yeah the plugin implementation didn't change so we should keep the message (although yes the spec changed)

kevinSuttle: Ok. Should it still be formatted as strikethrough?

hzoo:
Great

kevinSuttle:
Thanks!

PR #885, opened 2016-08-08
Fix incorrect CLI comments flag option docs

hatched:


hatched:
After further testing the cli argument and lib argument are different. `comments` on the lib and `no-comments` on the cli. This should probably be normalised to one or the other and then the docs updated (or not).

hzoo:
Yeah we can mention the `no` in the cli docs

PR #887, opened 2016-08-13
fix misinformation

cztomsik:
babel-preset-es2016 does not include es2015


hzoo:
Thanks!

But I realize it actually says that below,

> If you need to compile ES6/ES2015, use the [es2015 preset](/docs/plugins/preset-es2015/)`

PR #888, opened 2016-08-14
Update docs to reflect preset-react update

bryanbierce:
This is a change to the documentation of the react preset to reflect the change in this open [PR to the preset](https://github.com/babel/babel/pull/3654) which adds the transform-class-properties plugin. This is only a valid change if that PR is merged.


bryanbierce:
Closing following the [associated change](https://github.com/babel/babel/pull/3654) being closed with a [better solution](https://github.com/babel/babel/pull/3655)

PR #892, opened 2016-08-20
Plugin/Preset Ordering

hzoo:


danez:
Are plugins applied before or after presets?

hzoo:
Yeah plugin before presets, and ordering is based on vistors used in the plugins (I guess this would make sense because we are doing the merging of visitors for the single pass) @taion

taion:
Another thing maybe worth noting is that if you do something like https://github.com/facebook/jest/blob/v13.2.3/packages/babel-jest/src/index.js#L17-L22, the presets (and plugins) specified there will all run before anything specified in `.babelrc`.

This came up in https://github.com/facebook/jest/issues/1387#issuecomment-238708769.

docs/plugins.md
```diff
@@ -171,6 +171,34 @@ These plugins allow Babel to parse specific types of syntax.
...
+
+Preset Ordering is reversed (last to first).
+
+> Yes this is confusing, see [babel/notes #2](https://github.com/babel/notes/blob/master/2016-08/august-01.md#potential-api-changes-for-traversal).
+
+> I believe the reason why (for backwards compatability) is that most users had listed "es2015" first and "stage-0" second.
+> And stage-0 would run before es2015.
+
+```js
+"plugins": [
```
ykzts: Did you mean "`presets`"?

PR #896, opened 2016-08-25
Fix indentation on plugins page

jmeas:
See https://github.com/babel/babel.github.io/pull/895/files#r76295818


jmeas:
Thanks! :v:

PR #897, opened 2016-08-25
Restructure the plugins/presets

danez:
This updates the structure to reflect ES2016 and ES2017

I removed the plugins from the preset section and instead created new sections in the plugins section for ES2016 and ES2017.

Also changed the syntax plugins to be separated into active/deprecated and the ones that are now active by default.


hzoo:
👍 awesome

PR #899, opened 2016-08-27
Update REPL to contain new presets

Daniel15:
Updates the REPL to contain the new presets (es2016, es2017, latest). Since the list of presets was getting pretty long, I moved it to be a dropdown list:

Live demo: http://babeldev.dan.cx/repl/

I had to hack around Bootstrap a bit. Its dropdowns close whenever an item within them is clicked, so I needed to add an event handler to avoid that happening. Additionally, using `<label>` tags messes up the dropdown styling, so I just used an `<a>` and handle that in the event handler. #yolo

### DIFF
--- a/_sass/pages/_repl.scss
+++ b/_sass/pages/_repl.scss
@@ -86,3 +86,16 @@
 #babel-repl-presets {
   display: inline-block;
 }
+
+.babel-repl-preset-container {
+  position: relative;
+
+  .dropdown-toggle {
+    cursor: pointer;
+  }
+}
+
+#babel-repl-preset-dropdown {
+  margin-left: 10px;
+  margin-top: -3px;
+}
--- a/repl.html
+++ b/repl.html
@@ -20,8 +20,14 @@ third_party_js:
         </label>
       </div>
 
-      Presets:
-      <div id="babel-repl-presets"></div>
+      <div class="form-group button-group babel-repl-preset-container">
+        <span class="dropdown-toggle" data-toggle="dropdown">
+          Presets: <span id="babel-repl-selected-presets"></span>
+          <span class="caret"></span>
+        </span>
+        <ul class="dropdown-menu" id="babel-repl-preset-dropdown"></ul>
+      </div>
+
       <label for="option-lineWrap">
         <input id="option-lineWrap" type="checkbox">
         Line Wrap
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -1,7 +1,18 @@
 (function(babel, $, _, ace, window) {
   'use strict';
 
-  var presets = ['es2015', 'es2015-loose', 'react', 'stage-0', 'stage-1', 'stage-2', 'stage-3'];
+  var presets = [
+    'es2015',
+    'es2015-loose',
+    'es2016',
+    'es2017',
+    'latest',
+    'react',
+    'stage-0',
+    'stage-1',
+    'stage-2',
+    'stage-3'
+  ];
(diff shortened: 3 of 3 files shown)


Daniel15:
cc @hzoo 

I haven't changed the default from `es2015,stage-2,react` - Do we want to include any of the new presets in the defaults?

Daniel15:
Also I wanted to show a pencil icon rather than the caret (to make it more obvious that you click the list to edit), but it seems like the Babel site doesn't load Glyphicons properly. The CSS is available, but the font file just throws a 404.

hzoo:
#yolo is cool

We could include latest (or 15,16,17), stage-2, react as the default. 

aside: not sure how we want to show preset options

Daniel15:
Yeah we'll probably need a fancier UI for preset options... I figured this is a good interim solution as more presets get added.

hzoo:
Yay thanks @Daniel15 , will have to look into the icon stuff

Daniel15:
This is what I was trying to do:
```
<span class="glyphicon glyphicon-pencil" aria-hidden="true"></span>
```

Which should work as per https://getbootstrap.com/components/

Daniel15:
This is what I was trying to do:
```
<span class="glyphicon glyphicon-pencil" aria-hidden="true"></span>
```

Which should work as per https://getbootstrap.com/components/

But I got this:
```
[2016-08-27 11:56:31] ERROR `/node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.woff' not found.
[2016-08-27 11:56:31] ERROR `/node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.ttf' not found.
```

I don't have that folder. I do have `node_modules\bootstrap-sass\assets\fonts\bootstrap\` though. Maybe some file path somewhere is just messed up.

brigand:
It might be good to add a description of each, involving the stability of each preset. For stage presets, some shortened version of [the process document](https://tc39.github.io/process-document/), what loose mode is, and for es2016 and es2017 the features included in them.

hzoo:
maybe https://github.com/babel/babel.github.io/blob/master/_sass/bootstrap/_glyphicons.scss#L11-L18

```
@font-face {
  font-family: 'Glyphicons Halflings';
  src: url("../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.eot");
  src: url("../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.eot?#iefix") format("embedded-opentype"), url("../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.woff") format("woff"), url("../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.ttf") format("truetype"), url("../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.svg#glyphicons_halflingsregular") format("svg"); }
```

Daniel15:
> It might be good to add a description of each, involving the stability of each preset

That's a good idea! Since the site is already loading Bootstrap and its JavaScript, we could just use [Bootstrap tooltips](https://getbootstrap.com/javascript/#tooltips), or add a help link that opens the documentation for that preset. 

@hzoo - That looks like it. I don't think `node_modules` is deployed to the live site, right? So even if the file path was correct, it wouldn't actually work on the live site. Probably just have to copy the font files across then modify the variable: https://github.com/babel/babel.github.io/blob/845d86b6c08dbc066859ad9611b554f606801d41/_sass/_variables.scss#L83

PR #900, opened 2016-08-27
Add Babili to REPL

Daniel15:
Adds Babili to the REPL. It is lazy-loaded, only if the "Babili" checkbox is checked.

Ideally #899 should be merged before this, as it collapses the preset list into a dropdown leaving more room for extra checkboxes like this. Both PRs are separate and can be merged individually though.

This is how it looks with both #899 and this PR:

### DIFF
--- a/repl.html
+++ b/repl.html
@@ -26,6 +26,13 @@ third_party_js:
         <input id="option-lineWrap" type="checkbox">
         Line Wrap
       </label>
+
+      <div class="form-group">
+        <label for="option-babili">
+          <input id="option-babili" type="checkbox">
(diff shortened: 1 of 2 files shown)


hzoo:
Awesome! should we move minify to the left of "line wrap"? not sure

hzoo:
🎉 😄

hzoo:
Hmm, maybe we should have an option to keep whitespace/newlines for readability? (pass minified: false to the babelrc)

hzoo:
Yeah maybe a "whitespace" option for that

PR #903, opened 2016-08-29
New User: NodeXperts

nodexpertsdev:
- Added NodeXperts to users.yml
- Added optimized nodexperts.svg image


danez:
merged with #986

PR #908, opened 2016-08-30
Fix incorrect example in `preset-es2017.md`

preco21:
`preset-es2017` does not have `transform-exponentiation-operator`.
Seems it should be fixed another example like `async/await`.


loganfsmyth:
Thanks!

PR #911, opened 2016-08-31
Update learn-es6.md

xiangzhuyuan:
Change some comment string to make it more easier to understand.


hzoo:
Sure, I think i'd change it a bit differently

PR #912, opened 2016-09-01
proxy console.clear() and console.error() to the real console

jayphelps:
I use the babel repl many times a day, and it's always been a pain that `console.clear()` and `console.error()` did not proxy to the _real_ console.

This obviously solves that.

So the common usage of adding `console.clear()` to the top of your code between REPL iterations will now clear the real console, as you would expect in JSBin, etc.

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -356,23 +356,10 @@
       if (done) flush();
     }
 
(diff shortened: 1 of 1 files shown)


hzoo:
Thanks @jayphelps!

PR #915, opened 2016-09-02
Fix lettercase for FAQ path in navbar

jelder:
Clicking the FAQ link in the navbar currently results in a 404 page because `_data/navigation.yml` has the url in upper case.


hzoo:
Thanks

PR #916, opened 2016-09-02
proxy other console methods to their corresponding real methods

jayphelps:
image

This also removes a duplicate call to the real console that was introduced in https://github.com/babel/babel.github.io/commit/8925d2259195f1c13ddd43080252b52ba6460ef1#diff-abe23564b3ea7bb6b8680951a5f778d3R362


hzoo:
👍

PR #924, opened 2016-09-26
Update syntax-class-properties.md

conartist6:
Add an example so that the purpose of this plugin may be more readily identified.


hzoo:
Thanks!

This is the syntax plugin readme so if we do this the pr would be for the transform plugin

We do have an example on the website link (unfortunately the readme and the site isn't synced)- http://babeljs.io/docs/plugins/transform-class-properties/

conartist6:
Oops, didn't notice I was on the syntax plugin. Would copying these into (or linking to them from) the README.md files be something which might be considered?

hzoo:
Thanks!

PR #931, opened 2016-09-30
caveats: update regenerator-runtime link

paulirish:
https://github.com/facebook/regenerator/blob/master/runtime.js indicates it's moved.


hzoo:
:+1: thanks!

PR #933, opened 2016-10-03
Change to "Set up Babel" from "Setup Babel"

wub:
In this case we want to use it as a verb. "Babel Setup" would make more sense if "setup" was kept.


hzoo:
Oh ok nice catch! Looked at https://www.visualthesaurus.com/cm/wc/is-it-set-up-set-up-or-setup/ and makes sense.

wub:
Tackling the big issues as always 😜

hzoo:
Feel free to check out https://github.com/babel/babel.github.io/labels/help%20wanted and on babel

Issue #932, opened 2016-10-03
stage-2 incorrectly lists object-rest-spread


olalonde:
See https://github.com/babel/babel/issues/4645

hzoo:
Yeah I made https://github.com/babel/babel.github.io/issues/928 - do you want to make a PR @olalonde? I can guide you through it

olalonde:
Thanks for offering but no time ;(

hzoo:
Looks like https://github.com/babel/babel.github.io/commit/dfc2aa090dcf2606a2d87480b49544e779d4dc60

hzoo:
https://github.com/babel/babel.github.io/pull/934

PR #934, opened 2016-10-04
fix stage-2 incorrectly lists object-rest-spread #932

kopijunkie:
updated the stage-2, stage-3 docs for object-rest-spread


hzoo:
Thanks!

PR #937, opened 2016-10-09
Updated documentation for transform-es2015-shorthand-properties

ellenaua:
Mentioned that transform-es2015-shorthand-properties transforms both properties and methods and added examples.


docs/plugins/transform-es2015-shorthand-properties.md
```diff
@@ -6,6 +6,34 @@ permalink: /docs/plugins/transform-es2015-shorthand-properties/
 package: babel-plugin-transform-es2015-shorthand-properties
 ---
 
+Compile ES2015 shorthand properties and methods to ES5
+
+ES2015
```
hzoo: We've been using In/Out in other plugins but this is good too

hzoo:
Thanks!!

PR #950, opened 2016-10-21
Fix markdown formatting

abhisekp:
Add a newline between Example Usage and usage code.

Ref: http://babeljs.io/docs/plugins/transform-async-generator-functions/


danez:
Thanks

PR #952, opened 2016-10-23
Link to Example Node Server w/ Babel

martyychang:
[The docs](https://babeljs.io/docs/usage/cli/) currently just warn people against using `babel-node` in production.

> You should not be using `babel-node` in production.

While this may make sense to Node.js veterans, for newbies like myself it would help greatly to see an example or reference implementation of using Babel in a production deployment. Since there's already an "official" [example-node-server](https://github.com/babel/example-node-server) repo, mentioning this repo as a reference in the same warning block I believe will help more people get started with using Babel correctly.

### DIFF
--- a/docs/usage/cli.md
+++ b/docs/usage/cli.md
@@ -98,7 +98,7 @@ There are many more options available in the babel CLI, see [options](/docs/usag
     experience a startup performance penalty as the entire app needs to be compiled on the fly.
   </p>
   <p>
-    Check out the [example Node.js server with Babel](https://github.com/babel/example-node-server)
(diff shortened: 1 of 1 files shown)


STRML:
Thanks @martyychang, this is a good change.

It appears most Markdown parsers will bail on that HTML, so could you make the link a standard anchor tag instead?

martyychang:
Hi @STRML thanks for the feedback. I just changed the link to use an `a` tag.

STRML:
:shipit: Thanks for the contribution!

PR #959, opened 2016-10-26
Update cli.md

dburrows:
Every (rare) time I need to ignore .babelrc and add some presets/plugins for a custom CLI build I come to this page and it's not clear at all how to do it, so I have to go trawling Stack Overflow to find out. I've added examples for these options.


hzoo:
Thanks!

loganfsmyth:
It looks like there may have been a mixup here, `--no-babelrc` is an option for `babel`, but this has added it to the list of options for `babel-node`.

dburrows:
Aha, sorry, missed the section nesting. I've opened a new PR to fix this #962

PR #960, opened 2016-11-01
move setup/try buttons

hzoo:
Idea: moving the buttons to the first part of the page? (had to change the color too)

screen shot 2016-11-01 at 12 24 23 am

the Setup color is on hover/active

### DIFF
--- a/_sass/components/_buttons.scss
+++ b/_sass/components/_buttons.scss
@@ -50,3 +50,18 @@
     }
   }
 }
+
+.jumbotron {
+  .btn-featured {
+    background: $babel-dark-gray;
+    color: $babel-yellow;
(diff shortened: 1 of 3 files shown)


danez:
I like that

thejameskyle:
These buttons are not accessible. You need to make it more clear that they are buttons

hzoo:
@thejameskyle should the colors be inverted or should it just have an outline?

thejameskyle:
An outline would be enough

hzoo:
screen shot 2016-11-01 at 1 08 12 pm

Cool simple enough?

thejameskyle:
Yeah that should be fine

PR #965, opened 2016-11-08
Remove outdated Disqus from posts; add Discourse

STRML:
Disqus was still referring to 6to5.


STRML:
Merging this in; the Disqus code is completely broken so we may as well go ahead with removing it.

PR #968, opened 2016-11-09
Init Resources Page

hzoo:
Ref https://github.com/babel/babel.github.io/issues/27

From @thejameskyle (looks awesome?)

screen shot 2016-11-09 at 12 23 57 pm

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -25,5 +25,7 @@
     url: /docs/usage/browser/
 - title: Try it out
   url: /repl/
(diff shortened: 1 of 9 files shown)


STRML:
:o Looks awesome!

hzoo:
I know! - was looking at an old branch (2yrs ago)

I think we should definetely do it - maybe update the links

thejameskyle:
Yeah, this is still a good idea I think. Just going to take a lot of maintenance

PR #971, opened 2016-11-11
Add browser usage

hzoo:
Fixes https://github.com/babel/babel.github.io/issues/170


danez:
Should we maybe then remove this?
https://babeljs.io/docs/usage/browser/

PR #975, opened 2016-11-13
Missing let, square function actually doubles

deepakshrma:
1. Missing let throws, number is not defined
2. square is equal to _(number*number)_ not _(number*2 )_

### DIFF
--- a/docs/learn-es6.md
+++ b/docs/learn-es6.md
@@ -73,8 +73,8 @@ var bob = {
 function square() {
   let example = () => {
     let numbers = [];
-    for (number of arguments) {
-      numbers.push(number * 2);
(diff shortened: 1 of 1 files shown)


docs/learn-es6.md
```diff
@@ -73,8 +73,8 @@ var bob = {
 function square() {
   let example = () => {
     let numbers = [];
-    for (number of arguments) {
-      numbers.push(number * 2);
+    for (let number of arguments) {
+      numbers.push(number * number);
```
danez: Can you please also adjust the comment with the return values to match this change?

hzoo:
Awesome!

PR #438, opened 2015-09-11
Add Algolia search

pixelastic:
As [rapidly discussed](https://twitter.com/pixelastic/status/640938034041593856) on Twitter, here is my proposal for adding search to the babeljs website.

Screencast

You can [test it live here](https://babeljs.algolia.com/).

This uses our [jekyll plugin](https://github.com/algolia/algoliasearch-jekyll) for pushing data to your [Algolia](https://www.algolia.com/) index, and our [autocomplete.js](https://github.com/algolia/autocomplete.js) library (inspired by typeahead) for displaying the results. All the ranking configuration is available in the `_config.yml` (which mirrors options available in our [REST API](https://www.algolia.com/doc/rest)), and in the `search.rb` file.

We will of course give you full access to your account, and you'll be available to edit all the settings as you see fit (just send me a PM with the email address you would like associated with) . We just ask that you keep the Algolia logo in the results.

Tell me what you think.

PR #977, opened 2016-11-15
add Docsearch

maxiloc:
@hzoo, Following #438,  and the discussion you had by email with Remy from Algolia, here is the docsearch implementation

yyksljifwg


hzoo:
Awesome!

Also was wondering if anything needs to be done to get keyboard shortcuts working (up and down arrows)

hzoo:
Hm we have a lot of the same settings so if I type API I get 
screen shot 2016-11-15 at 10 30 53 am

rather than anything under usage

screen shot 2016-11-15 at 10 30 16 am

maxiloc:
@hzoo The up and down arrow are handled by default, but the yellow highlight was a bit to transparent. It should be better now.
There was a mistake in the docsearch config and the usage pages were not picked up correctly. It should be ok now.

Tell me if it's good for you.

hzoo:
yeah it should be everyhting in https://github.com/babel/babel.github.io/tree/master/docs

looks better now
screen shot 2016-11-15 at 11 00 34 am

hzoo:
@maxiloc how does the config work or is that on your end? Was wondering how to get better results if we find there are issues. I'm cool to merge this now though

hzoo:
Since it looks like the top level h1 isn't picked up. Like if I search polyfill I might expect that specific page to show up http://babeljs.io/docs/usage/polyfill

maxiloc:
You can have a look at the config here: https://github.com/algolia/docsearch-configs/blob/master/configs/babeljs.json. It's a set of simple rules that are used to crawl the website. You can make PRs if you need or just ping us when you see something that is not optimal and we'll fix it.

We then crawl the site automatically with https://github.com/algolia/docsearch-scraper

hzoo:
Oh ok thanks a lot!

I was wondering given you added 
```
  "start_urls": [
    "http://babeljs.io/docs/",
    "http://babeljs.io/docs/usage/api"
  ],
```

whether we need to add the rest https://github.com/babel/babel.github.io/tree/master/docs/usage

maxiloc:
I tweaked a bit the config the polyfill should appear first now

maxiloc:
No we don't need to add it all urls manually. We have http://babeljs.io/docs/usage/api because http://babeljs.io/docs/ is a 404 but we need it to accept all http://babeljs.io/docs/* urls

Issue #647, opened 2015-12-17
Offline Support


marcobiedermann:
I think it would be great if babeljs.io would be offline first. The other day I was without internet and could not read through the docs. I've seen that the website is running on a SSL connection, which is necessary for ServiceWorkers to run.

thejameskyle:
PRs welcome but I'm probably not going to spend the time doing this

j-f1:
[DevDocs](https://devdocs.io) just added [a Babel section](https://devdocs.io/babel/) scraped from `https://babeljs.io/docs/*`. You can go to [devdocs.io/offline](https://devdocs.io/offline/) to save it for offline access.

PR #978, opened 2016-11-15
Offline support with service worker

xtuc:
First naive implementation to support offline mode.

Refers to issue #647

Some points:
* [x] Search is broken (search will work only for already searched items because Algolia doesn't support offline mode)
* [x] Handle trailing slashes in URLs (cache first strategy won't get invalid URL)
* [x] Remove worker development logs
* [x] Add offline indicator ?
* [ ] If the ressource is not in the cache, show an offline page ?

Note that it is necessary to know the files to be cached in advance. That's why I need Jekyll templating in `service-worker.js`.

@thejameskyle @marcobiedermann What do you think ?


boopathi:
Why not use [sw-toolbox](https://github.com/GoogleChrome/sw-toolbox) ? ... which provides a declarative way of handling routes

service-worker.js
```diff
@@ -0,0 +1,19 @@
+---
+---
+
+importScripts('/scripts/sw-toolbox.js');
+
+var swOptions = {
+  debug: true
+}
+
```
boopathi: It'll be good to use a versioned cache name.

```js
const VERSION = 1;

toolbox.cache.name = "Babel-Cache-" + VERSION;
```

service-worker.js
```diff
@@ -0,0 +1,19 @@
...
+var preCachedRessources = [
+  '{{ "/css/main.css" | prepend: site.baseurl }}?t={{ site.time | date_to_xmlschema }}"',
+  {% for page in site.pages %}
+  '{{ page.url }}',
+  {% endfor %}
+];
+
+toolbox.precache(preCachedRessources);
+
+toolbox.router.get('/(.*)', toolbox.cacheFirst, swOptions);
```
boopathi: I see requests to these domains as well,

```js
toolbox.router.get("/*", toolbox.cacheFirst, { origin: "cdnjs.cloudflare.com" });
toolbox.router.get("/*", toolbox.cacheFirst, { origin: "cdn.jsdelivr.net" });
toolbox.router.get("/*", toolbox.cacheFirst, { origin: "unpkg.com" }); // for repl
```

service-worker.js
```diff
@@ -0,0 +1,19 @@
...
+
+var swOptions = {
+  debug: true
+}
+
+var preCachedRessources = [
+  '{{ "/css/main.css" | prepend: site.baseurl }}?t={{ site.time | date_to_xmlschema }}"',
+  {% for page in site.pages %}
+  '{{ page.url }}',
+  {% endfor %}
```
boopathi: I'm not sure if the following is valid. I see this to be valid at-least for page transitions from and to REPL.

The Service Worker shouldn't change from pageA to pageB if both the pages are under the scope of the same service worker. This is because, service worker updates based on the byte diff, and during a page change, a new service worker will install (if the resources are different -> i.e new `service-worker.js`).

service-worker.js
xtuc: Yes, I will add cache versioning.

What do you think about a max age ? Like 7 days ?

I don't know if limiting entries is a good idea. I pre-cache already all pages.

service-worker.js
xtuc: Ok, I will add these.

I'm currently searching a way to respond to Algolia queries. It seems to not support offline mode. For pre-caching I already use a list of all pages, it should be easy to search in it.

boopathi: Algolia queries are `POST` queries - `toolbox.router.post(...)`

service-worker.js
xtuc: Since the strategy is `cacheFirst` I removed the file `main.css` which has a time in it.

xtuc: Since the strategy is `cacheFirst` I removed the file `main.css` which has a time in it.

xtuc:
The offline indicator looks like this :

offline-babel

Might not be the most beautiful way to do.

xtuc:
Or the grey version:

offline-grey

xtuc:
Label version:

offline-label

existentialism:
Nice work!

I feel like the notification should stand out enough to draw attention, but not be distracting. Maybe black (fits color scheme and contrasts with the yellow header)?

image

boopathi:
I tried a variant of grey version - I like the label. My opinion is that removing/changing yellow might give the feeling of offline. 

screenshot 2016-11-23 22 20 51

addyosmani:
This is all really awesome to see come together. Fwiw, we're seeing a few sites choose the grayed out pattern for indicating offline (coupled with a top or bottom bar with the 'offline' labelling):



Some alternatives we've suggested elsewhere include showing a toast when switching online state or just having something prominent in the UI that can be toggled depending on when the user has a network connection again.

ikhattab:
using `glyphicon glyphicon-off` and background color of `#f55555` and we can fade color to green and fade out the offline div when the user is online

screen shot 2016-11-24 at 12 11 53 am

service-worker.js
addyosmani: TL for sw-toolbox and sw-precache here. Great to see you using it :)

If you're going to set `maxEntries` to 20, it means that after the 21st entry is cached, the least-recently used entry would be automatically deleted. Unless you're storing a number of really large assets, I'd consider whether this is really that valuable. I don't think the Babel site would benefit from setting this all that much.

I do think there's use in imposing a maxAgeSeconds age. @boopathi can probably comment to the length of time, but a week doesn't sound like a bad ballpark.

marcobiedermann:
@xtuc Awesome, great work you did there. I quite like that there would be a little indicator on the top which shows if the user is offline.
Looking forward getting this into production 😄  🎉

service-worker.js
addyosmani: Depending on how you want to handle post requests, you could do `toolbox.router.post('/endpoint', toolbox.networkOnly)`. This means you're not just returning a cached response but are instead stating that requests made to endpoint need to be network only.

service-worker.js
```diff
@@ -15,4 +15,8 @@ var preCachedRessources = [
 
 toolbox.precache(preCachedRessources);
 
-toolbox.router.get('/(.*)', toolbox.cacheFirst, swOptions);
+toolbox.router.get('/*', toolbox.cacheFirst, swOptions);
+toolbox.router.get('/*', toolbox.cacheFirst, { origin: "cdnjs.cloudflare.com" });
+toolbox.router.get('/*', toolbox.cacheFirst, { origin: "cdn.jsdelivr.net" });
+toolbox.router.get('/*', toolbox.cacheFirst, { origin: "unpkg.com" }); // for repl
+toolbox.router.post('/*', toolbox.cacheFirst, { origin: "algolia.net" }); // Cache Algolia search response
```
addyosmani: I mentioned this in your earlier commits, but is the goal here to always return a cached post response from Algolia or to only have this work when the network is available?

mbj36:
'offline fallback' for the last point would do the same. We have to make a offline page for it. We have to add service worker 'fetch' event in index.js .

Any idea what should offline page looks like ?

jeffposnick:
Apologies for the drive-by feedback. I work on `sw-toolbox`/`sw-precache`, and I've spent some time thinking about best practices for using service workers with templated sites, so I wanted to throw in my two cents. 

I'll leave some inline feedback on this PR, and I've started a series of posts laying out some best practices for this scenario, with part 1 at https://jeffy.info/2016/11/02/offline-first-for-your-templated-site-part-1.html

service-worker.js
```diff
@@ -0,0 +1,22 @@
+---
+---
+
+importScripts('/scripts/sw-toolbox.js');
+
+const VERSION = 1;
```
jeffposnick: How will you make sure that `VERSION` is incremented whenever any existing content anywhere on the site is updated? If `VERSION` doesn't get incremented, then the previously cached content for a given URL will never get refreshed, since there's a `cacheFirst` policy being used.

service-worker.js
```diff
@@ -0,0 +1,22 @@
...
+var preCachedRessources = [
+  {% for page in site.pages %}
+  '{{ page.url }}',
+  {% endfor %}
+];
+
+toolbox.precache(preCachedRessources);
+
+toolbox.router.get('/*', toolbox.cacheFirst);
+toolbox.router.get('/*', toolbox.cacheFirst, { origin: "cdnjs.cloudflare.com" });
```
jeffposnick: You'd probably want to configure a different cache name here the runtime CDN caches, so that they're reused indefinitely and don't become irrelevant whenever you change the default `toolbox.cache.name` value above.

If you're using `cacheFirst` strategy for them, please make sure that all of the URLs that they refer to are versioned. E.g. https://unpkg.com/react@15.3.1/dist/react.min.js includes `15.3.1` in the URL, so it's fine to use `cacheFirst`. But if you request URLs like https://unpkg.com/react/dist/react.min.js, once it's cached for the first time, the initial version will be reused indefinitely (at least until the cache name changes).

service-worker.js
```diff
@@ -0,0 +1,22 @@
...
+toolbox.router.get('/*', toolbox.cacheFirst, { origin: "cdn.jsdelivr.net" });
+toolbox.router.get('/*', toolbox.cacheFirst, { origin: "unpkg.com" }); // for repl
+toolbox.router.post('/*', toolbox.cacheFirst, { origin: "algolia.net" }); // Cache Algolia search response
```
jeffposnick: I don't know exactly what `algolia.net` is used for specifically, but if it's search-related, then `cacheFirst` doesn't sound like a good strategy to use. `networkFirst` would normally be more appropriate, since you'd want fresh search results, but could fall back to stale results if the network is unavailable.

service-worker.js
```diff
@@ -0,0 +1,22 @@
...
+
+var preCachedRessources = [
+  {% for page in site.pages %}
+  '{{ page.url }}',
+  {% endfor %}
+];
+
+toolbox.precache(preCachedRessources);
+
+toolbox.router.get('/*', toolbox.cacheFirst);
```
jeffposnick: This is super broad. Are you *sure* you're okay with every URL under your origin being served `cacheFirst`? And aside from the caching strategy, are you sure that any time there's any request made for a resource on your site, you want that added to the cache? If there are large images or other large media, then that's not necessarily a good approach.

service-worker.js
```diff
@@ -0,0 +1,22 @@
...
+
+importScripts('/scripts/sw-toolbox.js');
+
+const VERSION = 1;
+
+toolbox.cache.name = "Babel-Cache-" + VERSION;
+
+var preCachedRessources = [
+  {% for page in site.pages %}
+  '{{ page.url }}',
```
jeffposnick: If you're precaching entire HTML documents, then you're going to end up populating user's caches with a lot of duplicated data. I.e. if your HTML shares common headers/footers (because they're all using some underlying layouts) then the structure of those headers and footers will be take up space on your user's disk when they don't actually add any extra value.

The ideal implementation would just cache the layout elements once, then cache the underlying content once, and then perform the templating logic in the service worker to combine them at runtime.

I've got a very experimental example of this implemented at https://github.com/jeffposnick/jeffposnick.github.io/tree/work/src, but I don't know that it's stable enough to recommend using for a high-traffic site.

service-worker.js
```diff
@@ -0,0 +1,22 @@
+---
+---
+
+importScripts('/scripts/sw-toolbox.js');
+
+const VERSION = 1;
+
+toolbox.cache.name = "Babel-Cache-" + VERSION;
```
addyosmani: 👍  to Jeff's comment. You'll want to iterate through your cache entries and `cache.delete` doing something along these lines:

```js
// Where urlsToCacheKeys is a map of your items to cache
self.addEventListener('activate', function(event) {
  let setOfExpectedUrls = new Set(urlsToCacheKeys.values());
  event.waitUntil(
    caches.open(cacheName).then(function(cache) {
      return cache.keys().then(function(existingRequests) {
        return Promise.all(
          existingRequests.map(function(existingRequest) {
            if (!setOfExpectedUrls.has(existingRequest.url)) {
              return cache.delete(existingRequest);
            }
          })
        );
      });
    }).then(function() {
      return self.clients.claim();
    })
  );
});
```

service-worker.js
hzoo: algolia.net is for the search functionality 

screen shot 2016-11-15 at 11 00 34 am

jeffposnick:
After looking through the code, my high-level comment is that it's unfortunately tricky to effectively implement this type of runtime caching: you have to balance a few things:

- making sure that every user who visits your site once doesn't end up downloading a large amount of data in the background, which might be expired before they ever get a chance to make use of it
- the flexibility to make small changes to specific pages and only causing those changed URLs to be purged and refreshed
- the desire for the speed of cache-first with the need to ensure that updates to previously cached content are exposed to the user

I don't think it would be safe to move ahead with the current implementation, for the reasons outlined in the inline comments. If you did want to go with an approach that relied on `sw-toolbox`, then switching to a `networkFirst` strategy for the URLs on your site would end up being safer in general, but it would still not be as ideal as it could be. (I'd be happy to try out an approach that I've been working on at https://github.com/jeffposnick/jeffposnick.github.io/tree/work/src, but it's not quite ready for production at this point.)

service-worker.js
xtuc: I wanted to keep search even once offline. The goal here is to cache some responses from Algolia because it does not support offline searching. I'm not sure about this since the user will be able to find what he already searched online.

service-worker.js
xtuc: Yes sure. We could use the `site.date` from Jekyll. But this way the entier cache will be invalided even if the content hasn't changed.

service-worker.js
boopathi: @addyosmani I guess not. it should be networkFirst.

@xtuc `cacheFirst`, `networkFirst`, `fastest` - all these ultimately hit the cache in some way.

service-worker.js
xtuc: Yes good point. This is part of good practice I guess.

service-worker.js
xtuc: Yes, I'm now thinking that idea wasn't that good. (https://github.com/babel/babel.github.io/pull/978#discussion_r89448045)

service-worker.js
xtuc: I wanted an offline documentation. This caching strategy might indeed not be the best approach.

service-worker.js
xtuc: Rather than cleaning manually the cache, I think we need to limit the cache using `cache.maxEntries` or `cache.maxAgeSeconds`.
The old cached website(s) will be cleaned up given these configurations.

service-worker.js
boopathi: We could do networkOnly, but @xtuc wanted that the user be able to search even when offline - as mostly it is just static data updated via commits. So, it could be networkFirst with a separate cache name and maxEntries.

xtuc:
@mbj36: There is a 404 page : http://babeljs.io/404 but I wanted to create a dedicated "Sorry, you are offline" page.

@jeffposnick: Yes, good explanation of the problem. Your comments pointed out the bad points. I will update them. Thanks.

service-worker.js
```diff
@@ -0,0 +1,23 @@
+---
+---
+
+importScripts('/scripts/sw-toolbox.js');
+
+const VERSION = '{{ site.time }}';
+
+toolbox.cache.name = "Babel-Cache-" + VERSION;
+toolbox.cache.maxEntries = 150;
+toolbox.cache.maxAgeSeconds = 604800;
```
xtuc: @addyosmani @boopathi @jeffposnick I limited the cache to 150 entries and one week. Every time the website will be update the whole cache will be invalided.

service-worker.js
xtuc: I checked, resources from CDN (Cloudflare and jsdelivr) have a version in the URL.

service-worker.js
vigneshshanmugam: Does that mean the user will see the stale data for a week even if the resources change?

xtuc: I don't think so.

 ```js
const VERSION = '{{ site.time }}';
```

That ensure the cache to be invalided after a deployement of the website.

A stale website in cache won't show up but will probably be cached a week.

File: scripts/index.js [hunk 12]
```diff
@@ -33,7 +33,20 @@ $(document).ready(function() {
...
+    var offlineIndicator = $("#offline-indicator");
+
+    window.addEventListener('online',  function() {
+      offlineIndicator.addClass("hidden");
+    });
+
+    window.addEventListener('offline',  function() {
+      offlineIndicator.removeClass("hidden");
+    });
+
```
SerkanSipahi: @addyosmani flipkart is awesome !

service-worker.js
vigneshshanmugam: oops yeah missed that part.. Thanks!

service-worker.js
```diff
@@ -17,7 +21,7 @@ var preCachedRessources = [
 
 toolbox.precache(preCachedRessources);
 
-toolbox.router.get('/*', toolbox.cacheFirst);
-toolbox.router.get('/*', toolbox.cacheFirst, { origin: "cdnjs.cloudflare.com" });
-toolbox.router.get('/*', toolbox.cacheFirst, { origin: "cdn.jsdelivr.net" });
+toolbox.router.get('/*', toolbox.cacheFirst, contentCacheOptions);
+toolbox.router.get('/*', toolbox.cacheFirst, { origin: "cdnjs.cloudflare.com", name: "cdn" });
```
boopathi: this should be `{ cache: { name: "cdn" } }`

service-worker.js
```diff
@@ -17,7 +21,7 @@ var preCachedRessources = [
 
 toolbox.precache(preCachedRessources);
 
-toolbox.router.get('/*', toolbox.cacheFirst);
-toolbox.router.get('/*', toolbox.cacheFirst, { origin: "cdnjs.cloudflare.com" });
-toolbox.router.get('/*', toolbox.cacheFirst, { origin: "cdn.jsdelivr.net" });
+toolbox.router.get('/*', toolbox.cacheFirst, contentCacheOptions);
```
boopathi: `networkFirst` or `fastest` - cacheFirst is never going to hit the network after first time.

xtuc:
A other try : 

offline-v

scripts/service-worker.init.js
```diff
@@ -0,0 +1,21 @@
+if ('serviceWorker' in navigator) {
+
+  navigator.serviceWorker.register('/service-worker.js').then(function() {
+    var offlineIndicator = $("#offline-indicator");
+
+    window.addEventListener('online',  function() {
+      offlineIndicator.addClass("hidden");
```
mathiasbynens: nit: quotes are inconsistent (single quotes are used elsewhere)

service-worker.js
```diff
@@ -0,0 +1,18 @@
+importScripts('/scripts/sw-toolbox.js');
+
+var VERSION = '{{ site.time }}';
+
+var contentCacheOptions = {
+  name: "Babel-Cache-" + VERSION,
```
mathiasbynens: same in this file

scripts/service-worker.init.js
xtuc: Yes, thanks. I updated both files with single quotes.

xtuc:
@boopathi @marcobiedermann @mathiasbynens @jeffposnick what are you thought about this ?

jeffposnick:
The current implementation, with `toolbox.fastest` for (most of) the requests means that you'll have a much better upgrade story, and you'll avoid the issues I've previously pointed out where content would be cached once and potentially reused indefinitely. I don't see any specific shooting-yourself-in-the-foot issues.

What I'd recommend is that you try doing some sanity checks in a local deployment by keeping an eye on Chrome DevTool's [Cache Storage display](https://developers.google.com/web/tools/chrome-devtools/progressive-web-apps#caches) in the Applications panel, and the [Network panel](https://developers.google.com/web/tools/chrome-devtools/network-performance/resource-loading).

The Cache Storage display will give you a list of URLs that are being cached, and you can verify that you're not accidentally caching anything that you don't intend to. (Your `/*` route is very broad, and it's possible that there are requests that the site is making which you realize don't need to be cached.)

xtuc:
Thanks a lot @jeffposnick for your explanations.

LGTM, @hzoo @boopathi can we merge it ?

hereisnaman:
@jeffposnick @xtuc Hello I am new here and trying to contribute. 

I was thinking of if we can have different named caches for different kind of assets. Like for images we can have a separate cache, so that we can manage it differently and even it may improve the performance while matching requests from the cache.

Also I think it would be safer to go with networkFirst for the website assets, with cache maxAge. 

And do we need maxEntries? Can't we let it be maximum as they will get expired by age and so the the caching of assets won't be limited.

service-worker.js
```diff
@@ -0,0 +1,21 @@
...
+
+var VERSION = 1
+  , contentCacheOptions = {
+  name: 'Babel-Cache-' + VERSION,
+  maxAgeSeconds: 60 * 60 * 6
+}
+  , cdnCacheOptions = {
+  name: 'cdn',
+  maxAgeSeconds: 60 * 60 * 6
+};
```
xtuc: Could you please make multiple var declarations instead here?

hereisnaman: okay sure.

PR #982, opened 2016-11-19
Add npm "--save-dev" option to install plugin. Change line endings from Win to Unix

seedofjoy:
It's no need to leave plugins on production. Just clarify this.
Changed line endings from win to unix in `/docs/plugins/transform-react-display-name.md`


hzoo:
Cool 😄

Issue #528, opened 2015-11-02
Pull README.md from babel/babel repo packages


tomByrer:
eg
http://babeljs.io/docs/plugins/transform-es2015-modules-systemjs/
vs
https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-es2015-modules-systemjs
The "via CLI" & "via Node API" are missing.  IMHO README should be the source for the webpage  automated builder.
Added preferred bonus: the web page links to README ("Edit doc contents in README").

hzoo:
Yeah it would be cool to pull in the info from the readme so we wouldn't have to maintain it on both.

thejameskyle:
@hzoo I worked on that a bit this weekend, didn't get very far if you want to give it a shot.

The one requirement is that we need to add `Edit this page` buttons which link to the repo directly.

hzoo:
This is done via https://github.com/babel/babel.github.io/issues/990 but we need to actually move the pages as another issue.

PR #990, opened 2016-11-19
Use Babel package doc

xtuc:
Refers to issue #528.

NPM package page shows the header but the website already have one. I had to write a custom plugin to drop these lines. It uses the `page.package` to determine its README location.

I assumed that the header (package title and description) takes three lines.

There is a git module to include babel package repository.

This PR shows only a demo. Many documentations from the babel repo and from the website are very different.


hzoo:
Sounds like a good start! - will check it out later

_plugins/markdown.rb
```diff
@@ -0,0 +1,23 @@
+=begin
```
thejameskyle: GitHub Pages will not build with any plugins, so you don't even have the option to use it

_includes/package_readme.html
```diff
@@ -0,0 +1,12 @@
+{% capture readme %}
+    {% include babel/packages/{{ page.package }}/README.md %}
+{% endcapture %}
+
+{{ readme
+    | newline_to_br
+    | split: "<br />"
+    | slice: 4, 1000
```
xtuc: I also can use `shift` there.

Like this : 
```
    | split: "<br />"
    | shift
    | shift
    | shift
    | shift
    | join: "<br />"
```

This way I don't need to limit the string to 1000 but I don't like it.
Do you mind @thejameskyle ?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "_includes/babel"]
```
hzoo: We'll need to add some instructions to the readme or add that automatically in make serve/npm start?

_includes/package_readme.html
hzoo: This is to remove the first 4 header lines at the top of every readme https://raw.githubusercontent.com/babel/babel/master/packages/babel-plugin-syntax-async-generators/README.md - they all ahve it so it should be ok as a workaround (using shift).

hzoo:
LGTM we can make an issue or just iterate on the whole remove first 4 lines thing

Issue #831, opened 2016-05-28
Curated Plugin List


hzoo:
Maybe we want to do this instead a super small line I added to search for npm

`Many other community maintained presets are available on npm!` http://babeljs.io/docs/plugins/

thejameskyle:
That would be good 👍

hzoo:
Mostly through searching [github](https://github.com/search?l=JavaScript&o=desc&q=babel-plugin-&s=stars&type=Repositories&utf8=%E2%9C%93) and [npm](https://www.npmjs.com/search?q=babel-plugin-)

Should probably be part of our babel/maintained
- https://github.com/59naga/babel-plugin-add-module-exports
- https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy

Specific to a library
- https://github.com/gaearon/babel-plugin-react-transform of course (348k/m)
- https://github.com/lodash/babel-plugin-lodash
- https://github.com/istarkov/babel-plugin-webpack-loaders
- https://github.com/nkt/babel-plugin-react-remove-prop-types
- https://github.com/jridgewell/babel-plugin-incremental-dom

Interesting stuff like
- https://github.com/yahoo/babel-plugin-react-intl
- https://github.com/dtinth/babel-plugin-__coverage__
- https://github.com/speedskater/babel-plugin-rewire
- https://github.com/codemix/babel-plugin-typecheck
- https://github.com/marten-de-vries/kneden (aka babel-plugin-async-to-promises

mathiasbynens:
https://github.com/mathiasbynens/babel-plugin-transform-unicode-property-regex compiles [Unicode property escapes](https://github.com/mathiasbynens/regexpu-core/blob/master/property-escapes.md) (`\p{}` and `\P{…}`) in Unicode regular expressions to ES5 or ES6 (!) that works in today’s environments.

ctrlplusb:
https://github.com/gcanti/babel-plugin-tcomb
Babel plugin for runtime type checking using tcomb

03eltond:
https://bitbucket.org/amctheatres/babel-transform-imports
For libraries which support pulling in modular components, such as react-bootstrap and lodash:

```javascript
import Grid from 'react-bootstrap/lib/Grid';
```

But you'd have to write a line for each item you want to pull in, where it is more convenient to write:

```javascript
import { Grid, Button, Row, Col } from 'react-bootstrap';
```

[...]

PR #995, opened 2016-11-22
Curated Plugin List

xtuc:
Refers to issue #831.

I'm not sure about the wording and the categories but it's a good start to discuss.

cool-stuff


hzoo:
I think it would be great to be able to search, and show more than just a description (would be better as a demo/repl)

xtuc:
Could we use Algolia or do we need a custom search for this purpose ?

xtuc:
List page:
awesome1

Project:
awesome2

I think it's a good idea to show the package readmes as Babel's documentation.
URL are like: `babeljs.io/docs/awesome-babel/babel-plugin-add-module-exports`

Adding a new project should be as easy as adding a new user to the Babel users page.

hzoo:
That would be cool 😄

xtuc:
@hzoo what about a label like this ?

label-comm

hzoo:
Yeah something like that? We can iterate - might look better under the name like near the github/npm/edit this page section (we should modify https://github.com/babel/babel.github.io/blob/168843dfec36b2f3fc82d3c01c777b3b995c3d66/_layouts/page.html#L12-L18) to make it show up for packages outside of babel/babel

docs/awesome-babel/babel-plugin-add-module-exports.md
```diff
@@ -0,0 +1,111 @@
+---
+layout: docs
```
hzoo: Also what would be interesting is instead of copying the readme we get the readme from npm?

https://unpkg.com/babel-plugin-add-module-exports@0.1.4/README.md - not sure why the readme wasn't published for https://unpkg.com/babel-plugin-add-module-exports/ for the latest version

xtuc: I'm not sure that Jekyll will allow to include remote content (specialy using GitHub pages).

xtuc:
Would be great with #968 :+1:

hzoo:
I'm not sure about actually keeping the readme of the pages in the website because it would be out of date. and also specifically for add-module-exports I don't want to showcase it too much since it's not helpful in the end for things that do understand import/export - it's just a commonjs helper

xtuc:
Sure, it will be unmaintainable.

Issue #987, opened 2016-11-19
Add to clipboard button for cli commands/code?


hzoo:
Maybe https://clipboardjs.com/? just did a simple search

![image](https://cloud.githubusercontent.com/assets/588473/20458732/8e19d80a-ae7a-11e6-8d84-7dc6e8e1be6d.png)

xtuc:
Can we close this ? Since the PR was merged.

hzoo:
Yeah!

PR #998, opened 2016-11-23
Add to clipboard button for cli commands/code

fredericmarx:
Solves #987 📋✨

screen shot 2016-11-23 at 06 22 28

I added a [clipboard.js](https://clipboardjs.com/) button to all cli command and code fields as suggested in #987.

This probably needs a feedback mechanism like a tooltip or [toast](https://material.google.com/components/snackbars-toasts.html#snackbars-toasts-usage). I'm happy to take suggestions as to what would be the best approach here.


xtuc:
Seems cool but could you please remove the `npm-debug.log` file https://github.com/babel/babel.github.io/pull/998/files#diff-19511cd8dc1fd5bc732a2a9f1f4ee2e1.

seedofjoy:
Seems like it also copies `$` symbol, or I'm wrong?

fredericmarx:
@xtuc Ha, you're right. I'm also adding `npm-debug.log` to `.gitignore`, if ok; this should probably never be committed.

xtuc:
@seedofjoy What do you mean ? Can you join a link to the code ?

hzoo:
Awesome stuff @fredericmarx!

> This probably needs a feedback mechanism like a tooltip or toast. I'm happy to take suggestions as to what would be the best approach here.

We have bootstrap enabled for scrollspy and other things so we should be able to use the tooltips from there? http://getbootstrap.com/javascript/#tooltips

As for the `$` if there isn't an easy solution we should just remove it from the code snippets instead.

Also if it looks clearer we can add a background color to the code snippets: like react has 

screen shot 2016-11-23 at 7 16 05 am

lodash

screen shot 2016-11-23 at 7 16 30 am

hzoo:
http://tutsplus.github.io/clipboard/ is pretty nice as well actually?

screen shot 2016-11-23 at 7 21 18 am

seedofjoy:
For example:

```sh
$ npm install --save-dev babel-plugin-check-es2015-constants
```
Would in this case `$` will be copied to clipboard?

fredericmarx:
Great feedback! I added a `Copy` label to the button to make it more obvious and changed the code background color to a light shade of Babel yellow.

This is what it looks like:

screen shot 2016-11-23 at 14 23 09

fredericmarx:
@seedofjoy At the moment the button also copies the `$`-symbol. I tend to leave it out altogether; having a character that’s opaquely omitted from copying is probably confusing.

hzoo:
Awesome - testing it out locally. looks like pages like http://127.0.0.1:4000/docs/plugins/preset-latest/ aren't rendering the clipboard but the main page is (difference between .highlight pre/code?

hzoo:
another nice thing about http://tutsplus.github.io/clipboard/ is that it changes the text to "copied" after clicking

screen shot 2016-11-23 at 8 33 45 am

```
copyCode.on('success', function(event) {
    event.clearSelection();
    event.trigger.textContent = 'Copied';
    window.setTimeout(function() {
        event.trigger.textContent = 'Copy';
    }, 2000);
});
```

hzoo:
Also weird (might just be me but chrome + osx) that the text is being selected 

copy

hzoo:
Ok we can modify the docs and remove `$` if necessary

hzoo:
copy

we should also try to restyle that particular buttons focus instead of the blue line?

hzoo:
Yeah the differences of the main page and the docs pages  http://127.0.0.1:4000/docs/plugins/transform-react-jsx-source/

screen shot 2016-11-23 at 9 14 35 am

The button doesn't seem to show up without removing xs-small? And the styling might need to be applied to .highlight rather than pre for the docs pages since specificity

screen shot 2016-11-23 at 9 15 25 am

fredericmarx:
I really like the solution with the changing button label!

I've also added clipboard buttons to the docs pages. The lower one in the screenshot is focused.

screen shot 2016-11-23 at 15 48 17

fredericmarx:
> The button doesn't seem to show up without removing xs-small?

I decided to hide the button on small screens because it was overlapping with one-liners of code: https://github.com/babel/babel.github.io/pull/998/commits/ffab056aee60731f24d455ffd2f08396e832e767

hzoo:
Awesome looks good! Might be cool if we could remove comments from the text when copying haha.

We can make another issue to remove `$` since there's a bunch of those

hzoo:
also add's it to learn es2015 which is interesting! anything else? LGTM just need to start updating the docs

hzoo:
I'l make another issue for the $ and formatting, thanks a lot @fredericmarx!! 🎉

hzoo:
@fredericmarx I can add you as a collab for this repo if you'd like as well!

fredericmarx:
That'd be awesome! Thanks, @hzoo 🎉

hzoo:
and feel free to join our slack.babeljs.io if you haven't already

Issue #825, opened 2016-05-24
Make REPL editors resizable by width


mull:
Right side of code editor & left side of output would be nice if they were draggable. I found myself editing the CSS for the code output today needing a bit more horizontal space in the editor.

hzoo:
If you'd like to make a PR to be able to to drag the panes, sounds good to me?

hzoo:
Done in https://github.com/babel/babel.github.io/issues/1003!

PR #1003, opened 2016-11-23
Make REPL editors resizable by width.

seedofjoy:
Solves #825 

Makes resizable editor columns (10px transparent zone).
Size set in percent, panel minimum size - 20%.
repl-resize

### DIFF
--- a/_sass/pages/_repl.scss
+++ b/_sass/pages/_repl.scss
@@ -58,6 +58,14 @@
   background: #f9f9f9;
 }
 
+.babel-repl-resize {
+  position: absolute;
+  width: 10px;
+  height: 100%;
+  cursor: col-resize;
+  z-index: 10;
+}
+
 .babel-repl-reporter {
   position: absolute;
   z-index: 4;
--- a/_sass/pages/_repl.scss
+++ b/_sass/pages/_repl.scss
@@ -50,7 +50,6 @@
 
 .babel-repl-input {
   left: 0;
-  border-right: 1px solid darken($gray-lighter, 5%);
 }
 
 .babel-repl-output {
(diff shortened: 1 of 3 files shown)


brigand:
Does this have any issues if you move the mouse too fast? I remember having issues with that on a past project.

seedofjoy:
Only that cursor may change from `col-resize` to `text`. I can fix this by adding class to parent component with `cursor: col-resize` during resize phase. But actually this almost imperceptible.

brigand:
Cool, this is much nicer than the implementation I went with (an invisible fixed position div). TIL

hzoo:
Awesome! - might be ok to make 90-10 if it still works. Thoughts on adding a different color or signal on hover? astexplorer.net has this

seedofjoy:
Forgot about `babel-repl-errors` and `babel-repl-console` panels. I think they should resize with inputs (`errors` = `input` and `console`=`output` width).
In next commit will add colors, hover, 90-10, (and resizing `errors` and `console`?)

hzoo:
yeah looks like there was an earlier similar issue about that as well https://github.com/babel/babel.github.io/issues/90

seedofjoy:
What do you think about this? It's work nice (`errors` and `console` panels resize well), but unfortunately add visual styles with resize bar to bottom panel without major changes in layout very hard.

hzoo:
Lgtm - there are some layout issues with those anyway - if necessary we should change then?

seedofjoy:
out
Made some minor layout changes to fit resize bar over the entire height

xtuc:
Awsome :+1:. LGTM also.

seedofjoy:
I have tested on Chrome and Safari. In IE I can check tomorrow

hzoo:
We can merge now too

Looks like the logs don't show up for me on the bottom right anymore?

test:

- [x] resize
- [x] syntax errors on the left console
- [ ] output on the right console

hzoo:
actually I tried it again and it works fine? dono what happened before

hzoo:
Awesome stuff! Thanks @seedofjoy (will add you to the repo if you'd like)

seedofjoy:
@hzoo Great, thank you! :octocat:

hzoo:
and feel free to join our slack.babeljs.io if you haven't already

Issue #999, opened 2016-11-23
Create docs for https://github.com/babel/babili


hzoo:
Given https://github.com/babel/babel.github.io/pull/995 (curated plugin list) and https://github.com/babel/babel.github.io/pull/990 (readme from repos), we should render the readme's for babel-minify on the website!

cc @kangax @boopathi

hzoo:
https://github.com/babel/babel.github.io/pull/1004 is merged

If you want to include the readme from the monorepo you can do `{% include package_readme.html from="babili" %}`

Now we'll want to create a separate main page for babili

hzoo:
Moving over the already created plugin pages to use the README from babili

by adding

```
{% include package_readme.html from="babili" %}
```

- [ ] babel-plugin-transform-inline-environment-variables
- [ ] babel-plugin-transform-member-expression-literals
- [ ] babel-plugin-transform-merge-sibling-variables
- [x] babel-plugin-transform-minify-booleans
- [ ] babel-plugin-transform-node-env-inline
- [ ] babel-plugin-transform-property-literals
- [ ] babel-plugin-transform-remove-console
- [ ] babel-plugin-transform-remove-debugger
- [ ] babel-plugin-transform-simplify-comparison-operators
- [ ] babel-plugin-transform-undefined-to-void

Will also want to add the new plugins not listed ^

hzoo:
This is all done in https://github.com/babel/babel.github.io/pull/1029? @geordidearns 

You can add a `Fixes #999` in your commit or description of the PR to make this close when your PR is merged

existentialism:
Yep, I believe this is finished!

Issue #997, opened 2016-11-22
Move main plugin/package docs from here to babel/babel


hzoo:
> If you have issues please comment here, join our slack (signup: slack.babeljs.io, url: babeljs.slack.com), ping @babeljs on twitter

## #528

> eg
http://babeljs.io/docs/plugins/transform-es2015-modules-systemjs/
vs
https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-es2015-modules-systemjs
The "via CLI" & "via Node API" are missing. IMHO README should be the source for the webpage automated builder.
Added preferred bonus: the web page links to README ("Edit doc contents in README").

https://github.com/babel/babel.github.io/issues/528 was implemented via https://github.com/babel/babel.github.io/pull/990 (git submodule and rendering the markdown from the README)

Now this website repo can pull in information from the actual readme.md file used for npm in the babel repo.

## Issue

What we need to do now is to move the docs in [`docs/`](/docs) to link to the README's in [babel/babel/packages](https://github.com/babel/babel/tree/master/packages)

Basically for those docs we want to take the contents there and replace them.

Example: https://github.com/babel/babel.github.io/pull/990/files#diff-b11e0ae0e12e1f69b9e7bd1a63ac06d2R20

<img width="994" alt="screen shot 2016-11-22 at 3 35 33 pm" src="https://cloud.githubusercontent.com/assets/588473/20540995/573a1830-b0c9-11e6-8d43-ccee9a71e17f.png">

## Instructions

For each docs page that needs to be changed:

- [ ] Find the correct page on the babel website docs under [`docs/`](/docs) to edit: Example is [`docs/plugins/preset-stage-0.md`](/docs/plugins/preset-stage-0.md)

<img width="1003" alt="screen shot 2016-11-22 at 3 36 18 pm" src="https://cloud.githubusercontent.com/assets/588473/20541022/71528a68-b0c9-11e6-80df-07495aa63782.png">

- [ ] Find the corresponding page on [`babel/babel`](https://github.com/babel/babel): Example is [`babel-preset-stage-0/README.md`]

<img width="992" alt="screen shot 2016-11-22 at 3 37 00 pm" src="https://cloud.githubusercontent.com/assets/588473/20541043/8b45647c-b0c9-11e6-867a-d8e7aed5e0eb.png">

(https://github.com/babel/babel/blob/master/packages/babel-preset-stage-0/README.md)
[...]

PR #1004, opened 2016-11-23
Include docs for babili

xtuc:
Refers to #999.

I wanted the API to be consistent from #990 and for #997. `from` parameter points to the source repository (`from="babel"` is by default).


hzoo:
Looks good for now, we'll want to create another page, etc

PR #1005, opened 2016-11-24
Assure --outfile doesn't overwrite other files in lib directory

willnwhite:
I had to check this out before compiling to `lib` because my `lib` already had other files I didn't want to overwrite.


hzoo:
Cool!

Issue #1000, opened 2016-11-23
Render babel-handbook on the website?


hzoo:
https://github.com/thejameskyle/babel-handbook

Was thinking we could render https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md and https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/user-handbook.md as a page and link to it for editing

nitin42:
One page for the plugins and the other for user ?

PR #1006, opened 2016-11-24
Include babel handbook

xtuc:
Refers to #1000.

I created a page `/docs/handbook/`.

babel-handbook

@thejameskyle what do you think ?


xtuc:
We could remove the table of contents. It is redundant with the right sidebar.

```
| slice: 73, 100000
```

Or remove the right sidebar.

hzoo:
We can't really slice it in that way because we will probably add/remove headings in the future

xtuc:
Ok, Well, we will just add more shift for those new header I guess.

hzoo:
Do you think we should do the shift thing and then update it with handbook updates?

xtuc:
I don't know because it will require unnecessary maintenance.

We could remove the right bar to avoid the duplicate table of content.

/cc @existentialism what do you think ?

hzoo:
Probably easier to remove the right bar instead

xtuc:
I removed the right bar.

handbook

xtuc:
We need to link it somewhere, maybe in the core-packages section (http://babeljs.io/docs/core-packages/)

hzoo:
it's also at the bottom of the homepage atm

screen shot 2016-12-14 at 12 33 36 pm

Should link it where we want to talk about making plugins

xtuc:
Isn't it confusing for new user to have the button on the homepage ?

There is a section for that purpose: http://babeljs.io/docs/plugins/#plugin-development

hzoo:
yeah we can remove it too

xtuc:
@hzoo @existentialism Can we merge this?

hzoo:
Yeah just have the merge conflicts

index.md
```diff
@@ -346,13 +346,6 @@ export default function ({types: t}) {
     </div>
   </div>-->
 
-  <div class="text-center">
-    <div class="btn-wrapper btn-wrapper-lg">
-      <a href="{{ site.baseurl }}/docs/learn-es2015/" class="btn btn-featured">Learn ES2015</a>
-      <a href="{{ site.baseurl }}/docs/setup/" class="btn btn-featured">Set up Babel</a>
-    </div>
-  </div>
-
```
xtuc: Since #1108

existentialism:
Really wish we could skip rendering the ToC and use the right bar, but we can improve it later.

Couple things:

- Edit this page link doesn't work
- Some internal anchor links don't work (for instance `#replacing-a-node-with-multiple-nodes` vs `#toc-replacing-a-node-with-multiple-nodes`)
- Should we augment code block titling to support `diff`? Should we disable copy/paste for these blocks? (probably diff PR)

docs/handbook.md
```diff
@@ -3,6 +3,8 @@ layout: page
 title: Babel Plugin Handbook
 description: This document covers how to create Babel plugins.
 permalink: /docs/handbook/
+package_source: babel-handbook
+package: babel-handbook
```
xtuc: At least now the Github button works, "Edit this page" redirect to the README of the handbook and the package in npm is not found

PR #1008, opened 2016-11-26
Fix Github link for babili plugins

boopathi:


xtuc:
Nice, thanks.

I was asking me if `page.from` should be an absolute link in some cases. I think this will be useful for #995.

In my PR I also added

```yaml
community_plugin: yes
package: 59naga/babel-plugin-add-module-exports
```

Which is then used to point to the right repository.

This is almost the same feature, what do you think ?

boopathi:
Ah ... Didn't see that PR. Cool... I thought we will be including only plugins under babel org. Anything works !

xtuc:
I meant #1006 sorry. Community plugins aren't likely to be included in the website I guess.

hzoo:
If we want to feature some plugins we could render them? I just dono how to get the readme? Maybe we have a list and then you fetch the readme before building so we don't need git submodules anymore

xtuc:
@hzoo This will require an offline build step. No more push to update workflow.

hzoo:
I guess we can close this now due to https://github.com/babel/babel.github.io/pull/1069

PR #1009, opened 2016-11-27
Create Babel-types page

xtuc:
PR related to #1006.

Babel-handbook section for `babel-types` points to the package README on Github ([see](https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md#babel-types)).

Since #990, I guess more `internal` package will be published on the website. I suggest the `/docs/internals/` path.

### DIFF
--- a/docs/core-packages/types.md
+++ b/docs/core-packages/types.md
@@ -1,17 +1,9 @@
 ---
-layout: page
+layout: docs
 title: Babel types
 description: Babel Types is a Lodash-esque utility library for AST nodes
 permalink: /docs/core-packages/types
 package: babel-types
 ---
 
-<div class="container docs-content">
-  <div class="step-wizard">
-    <div class="step">
-
-      {% include package_readme.html %}
-
-    </div>
-  </div>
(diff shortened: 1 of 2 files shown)


hzoo:
we can say `packages` too?

xtuc:
Babel has many packages. These seems to me more `internal`.

This is also related to the handbook. We could just name it `handbook/types` ?

hzoo:
I don't think we need to relate it to the handbook - it's part of babel/babel like in https://github.com/babel/babel#core-packages

xtuc:
Ok, so what about `/docs/core-packages/`?

xtuc:
The preview:

babel-type-website-doc

hzoo:
This isn't linked from anywhere yet so we'll want to create a page for core

xtuc:
@hzoo Could we merge this ? I created an issue #1021.

hzoo:
Ok but yeah no one will be able to find it atm 😄 ?

PR #1010, opened 2016-11-29
Pull React transform docs from babel/babel

existentialism:
Companion to https://github.com/babel/babel/pull/4903

Couple notes:

* The `strip_html` call when [parsing a README](https://github.com/babel/babel.github.io/pull/990/files#diff-5b919284089b446b968a163795522ffbR10) is problematic for examples/code blocks that actually contain html 😜  For local, I replaced it with `| replace: "<br />", ""` but didn't do extensive testing.

image

* The current strategy of stripping the first four lines means I made some errors when updating the package READMEs: we need to make ensure they're one line. Before I push up a PR to fix, do we want to re-evaluate at all?

* Will be pushing a PR to fix -self and -source needing new lines for Out:

image

/cc: @xtuc @hzoo


xtuc:
Yes, sorry my bad. I'll look at `strip_html`.

xtuc:
I left a comment in babel/babel#4903.

Could you add the description of the package in the header ?

existentialism:
@xtuc yep, will send a PR to fix them up

existentialism:
@xtuc 

Pushed PR on babel/babel to fix up descriptions + issues: https://github.com/babel/babel/pull/4917

Tested https://github.com/babel/babel.github.io/pull/1015 and found one issue with the "Out" block inside `transform-jsx-source`:

image

Compare with:
image

hzoo:
I guess we just have to update this?

existentialism:
Rebased, need to spend a little more time finding a cleaner solution for the previous comment. It looks like Jekyll doesn't like the double braces =/

hzoo:
looks like it was doing ` {% raw %}` before

existentialism:
Yeah, trying to find a non brittle way of injecting it when we parse the package's README. I'm assuming we don't want `{% raw %}...{% endraw %}` to show up in the README?

xtuc:
We can't because READMEs are also visible on npmjs and Github.

existentialism:
@xtuc yep, I had the same concern.

I haven't been able to come up with a reliable/non brittle way to deal with the double quotes (GH Pages seems limiting here).

Any ideas? Or maybe we just do a quick/unsightly fix and just separate the double quotes and move on for now?

```
<sometag __source={
  { fileName: 'this/file.js', lineNumber: 10 }
} />

// or

<sometag __source={ { fileName: 'this/file.js', lineNumber: 10 } } />
```

hzoo:
If putting a space in between fixes it we could do that

existentialism:
Yep:
image

Will push a fix and then this will be 👍

PR #1011, opened 2016-11-29
Link check-es2015-constants's doc to its package's README

lukyth:
According to #997 and a follow up to babel/babel#4908


xtuc:
LGTM :+1:

I prefer waiting for babel/babel#4908 before merging.

/cc @hzoo

Issue #1002, opened 2016-11-23
Cleanup code snippets after 


hzoo:
![copy](https://cloud.githubusercontent.com/assets/588473/20567660/9c19adb4-b167-11e6-8b6e-6e295b3c6f28.gif)

cc @fredericmarx 

We merged https://github.com/babel/babel.github.io/pull/998 which adds a "copy" button to copy to clipboard for code snippets. We should look through to make sure everything is rendered correctly and the copied text makes more sense!

- [x] remove `$` for bash/cli commands since it will get copied. If we need to show that it is a command we should do something else.

`$ npm install --save-dev babel-cli`

- [ ] split up larger code snippets so it's more easily copiable
- [ ] make sure snippets are valid json/js/commands

hzoo:
Also it think it for differentiating the snippets maybe we could add some css to the code frame to signify which language it is? (based on the css class `language-js`, etc)

<img width="708" alt="screen shot 2016-11-23 at 10 36 41 am" src="https://cloud.githubusercontent.com/assets/588473/20567974/cefeef40-b168-11e6-878d-dcad90d3d208.png">

@stefanpenner do you know how it's done for ember?

xtuc:
I close this issue because I created an issue #1049 regarding the last comment.

PR #1013, opened 2016-11-29
Cleanup main page bash code snippets

seedofjoy:
For #1002 

On main page:
- Changed highlight language from JS to shell in cli commands
- Remove `$` symbols
- Add `$` with :before content


xtuc:
This is great :+1: 

I added support for code blocks generated from markdown.

xtuc:
As said @hzoo, some codes must be splited. This is causing weird situations : 

basic-setups

seedofjoy:
I just started to work with code and realized that adding `$` with `before` maybe not a good idea. Some blocks no need to split, but `$` displayed only on first line.
For example:

What do you think?

xtuc:
@hzoo @hzoo I guess we should split blocks of code first ?

hzoo:
Yeah I did some of that for the presets at least - will have to look at the other places

seedofjoy:
Sorry, will open new PR

PR #1015, opened 2016-11-30
Fix package_readme.html include

xtuc:
As pointed out in #1010.

`strip_html` was not a good idea since some snippets (JSX for example) contains HTML.

For now, only `<br />` will be removed from snippets.

### DIFF
--- a/_includes/package_readme.html
+++ b/_includes/package_readme.html
@@ -14,6 +14,6 @@
     | split: "<br />"
     | shift | shift | shift | shift
     | join: "<br />"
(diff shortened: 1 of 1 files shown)


_includes/package_readme.html
```diff
@@ -14,6 +14,6 @@
     | split: "<br />"
```
xtuc: I wanted to convert newlines to tags to be able to split them and remove n lines from the header.

Split by `\n` does not work but would solve the problem.

PR #1031, opened 2016-12-04
Provide a keyboard shortcut to get to the search box.

### LINKED ISSUE #1027: Provide a keyboard shortcut to get to the search box?
Not sure what the common shortcut would be (or if others do this). I know tab doesn't seem to move there because it's hidden by default anyway.

screen shot 2016-12-03 at 8 30 54 am

fleischie:
This PR should implement a solution for issue #1027 

It adds an event-handler for the 'keypress' event, that checks the key codes
and opens the searchInput box. It is necessary to use the 'keypress'
event, as the question mark is a key combo on the us-en keyboard layout
(and de-de and possibly others, too).

EDIT @aaronang: Fixes #1027.

### DIFF
--- a/scripts/search.js
+++ b/scripts/search.js
@@ -42,8 +42,14 @@
                     return;
             }
 
-            event.preventDefault();
-            $toggleCheckbox.click();
+            // don't type '/' or '?', if the search box is closed or it's not
+            // focused (i.e. to regain focus, without typing the character)
+            if (!$toggleCheckbox.prop('checked') ||
+                !$searchInput.is(':focus')) {
+                    event.preventDefault();
+            }
+
+            $toggleCheckbox.prop('checked', true);
             $searchInput.focus();
         });
 
--- a/scripts/search.js
+++ b/scripts/search.js
@@ -8,6 +8,7 @@
     var KEYBOARD_ESCAPE = 27;
     var KEYBOARD_SLASH = 47;
     var KEYBOARD_QUESTION_MARK = 63;
+    var KEYBOARD_S = 115;
 
     // config is https://github.com/algolia/docsearch-configs/blob/master/configs/babeljs.json
     var $search = docsearch({
@@ -33,12 +34,13 @@
             $searchInput.focus();
         });
 
-        // Open the search when pressing / or ?
+        // Open the search when pressing / or ? or s
         // Use 'keypress' events to handle key combos
         // (e.g. en: '?' = Shift + '/')
         $(document).on('keypress', function(event) {
             if (event.keyCode !== KEYBOARD_SLASH &&
-                event.keyCode !== KEYBOARD_QUESTION_MARK) {
+                event.keyCode !== KEYBOARD_QUESTION_MARK &&
+                event.keyCode !== KEYBOARD_S) {
                     return;
             }
 
--- a/scripts/search.js
+++ b/scripts/search.js
@@ -5,6 +5,10 @@
     var $selector = '#babel-search-input';
     var $searchInput = $($selector);
 
+    var KEYBOARD_ESCAPE = 27;
+    var KEYBOARD_SLASH = 47;
+    var KEYBOARD_QUESTION_MARK = 63;
+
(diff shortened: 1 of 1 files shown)


scripts/search.js
```diff
@@ -29,6 +29,18 @@
             $searchInput.focus();
         });
 
+        // Open the search when pressing / or ?
+        // Use 'keypress' events to handle key combos
+        $(document).on('keypress', function(event) {
+            if (event.keyCode !== 47 && event.keyCode !== 63) {
```
xtuc: I would use constants for that.

```js
var KEYCODE_SLASH = 47;
var KEYCODE_QUESTION_MARK = 63;
```

scripts/search.js
```diff
@@ -29,6 +29,18 @@
...
+                return;
+            }
+
+            event.preventDefault();
+            $toggleCheckbox.click();
+            $searchInput.focus();
+        });
+
         // Hide the search when pressing Escape
         $(document).on('keydown', function(event) {
```
xtuc: By the way, could we clear the input when pressing escape ?

scripts/search.js
```diff
@@ -29,6 +29,18 @@
...
+        // Open the search when pressing / or ?
+        // Use 'keypress' events to handle key combos
+        $(document).on('keypress', function(event) {
+            if (event.keyCode !== 47 && event.keyCode !== 63) {
+                return;
+            }
+
+            event.preventDefault();
+            $toggleCheckbox.click();
+            $searchInput.focus();
```
xtuc: I noticed that we can't use `/` or `?` in search input anymore since it will click / focus it.

If the user is currently typing, we need to ignore these events (exept for espace).

What do you think about this ?

fleischie:
Thanks for the review. :bowing_man: 

I will add the variables, the input-clearing and fix the input-issues. I hope I manage to finish until tomorrow evening. :smile:

fleischie:
I have found several issues regarding my PR as of now (f7a8c15):

1. Typing a search-term in the search input, closing the search box with ESC, opening the search box again and pressing ESC without typing in text, blurs the focus on the input element and adds the text previously in the search input before closing it the first time with the ESC key. I don't know where this text is coming from, actually.
2. My installation of firefox opens the build-in search (similar to Ctrl-F in chrome/chromium/opera next) and does not act at all on '?'. This is probably due to the 'Slash' key not being keyhandled appropriately, but I didn't have time to look into this en detail. Thus I might add more keys, but I don't want to do this without prior consultating the community. :smiley:

fleischie:
I updated, rebased and force pushed my PR to be on top of master again, I hope this is ok.

The firefox-issue is resolved now, in that it opens the search-box by using a key (`s`, `/` and `?` is supported now) and closing it (`esc`). The search bar is cleared on escapes.

Chromium and Opera (the newer version) still handles escapes weirdly:
- Chromium loses focus on the search field, but cleares and closes on the second `esc`. Entering `s`, `/` or `?` opens the cleared search field but an immediate `esc` will return the previous search entry, which I cannot explain to be perfectly honoest.
- Opera opens and clear/closes correctly (each on the first press of the keys) but has the same behavior regarding `esc`-presses after opening the search fields.

Tips are very welcome. :wink:

raspo:
We should make sure to not trigger the search box when focussed on another input field.

i.e.
On the homepage, focus on the "mini-repl" textarea (where it says "Put in next-gen JavaScript"), pressing `s` shouldn't open up the search box.

aaronang:
Fixes #1027.

@fleischie Thank you for your contribution! Could you in the future refer to the issue you are fixing using this syntax: https://github.com/blog/1506-closing-issues-via-pull-requests? This makes it a bit easier for us 😬

aaronang:
Fixes #1027.

aaronang:
Fixes #1027.

fleischie:
@aaronang Yes, I'm sorry. The next time I will be more considerate. :smile: 

@raspo I extended the PR to check the activeElement and only open the search field, if it's not an input-element.

PR #1039, opened 2016-12-06
Remove caveat from transform-object-rest-spread.md

kaicataldo:
https://github.com/babel/babel/pull/4755#issuecomment-265071657


hzoo:
nice haha

olalonde:
Shouldn't we wait for https://github.com/babel/babel/issues/4851 to be fixed before changing those docs?

PR #1047, opened 2016-12-06
Resources page v2

xtuc:
Update from #968.

I think this should be part of #995 (Curtaed Plugin List).


xtuc:
@hzoo @existentialism We could merge this with the #995 Curtaed Plugin List ?

We can keep project pages:

7305437e-b17c-11e6-8afb-c038aff3103e

PR #1064, opened 2016-12-08
Update VS Code babel extension

joshpeng:
Babel ES6/ES7 extension originally referenced is outdated and does not correctly scope many JS contexts.

### DIFF
--- a/docs/editors.md
+++ b/docs/editors.md
@@ -40,7 +40,7 @@ Another option is to use [yajs.vim](https://github.com/othree/yajs.vim) with
 
 ### Visual Studio Code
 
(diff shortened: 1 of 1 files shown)


docs/editors.md
```diff
@@ -40,7 +40,7 @@ Another option is to use [yajs.vim](https://github.com/othree/yajs.vim) with
 
 ### Visual Studio Code
 
-Install the [Babel ES6/ES7](https://marketplace.visualstudio.com/items?itemName=dzannotti.vscode-babel-coloring)
```
hzoo: funny that it's called Sublime Babel - should we just be more specific and add `sublime-babel-vscode` as the title?

joshpeng: Good call, I changed it to your suggestion for this PR.

hzoo:
Thanks!!

PR #1065, opened 2016-12-08
Switch to using the GitHub Pages gem

existentialism:
Noticed some oddities while running the site locally vs what is published. The most obvious example is on the setup page:

Published on babeljs.io:
image

Local:
image

Notice the headings are different? The odd thing is, is that it's loading `docs.js` onto the page even though it's using the `default` layout.

I haven't completely traced it exactly, but I did notice that we're using `jekyll@3.1.3`, which suffers from the likely culprit: https://github.com/jekyll/jekyll/issues/4897. Upgrading to `jekyll@3.1.4` does solve the problem.

Either way, I think it makes more sense to switch over to using the `github-pages` gem which will allow us to mimic the same deps/env that is used during publishing.

Thoughts?


hzoo:
Oh Cool I didn't know about that

xtuc:
Nice job :+1:

Issue #1049, opened 2016-12-07
Indicate language in code snippets


xtuc:
Initially from #1002.

@hzoo :
> Also it think it for differentiating the snippets maybe we could add some css to the code frame to signify which language it is? (based on the css class language-js, etc)

<img width="708" alt="cefeef40-b168-11e6-878d-dcad90d3d208" src="https://cloud.githubusercontent.com/assets/1493671/20967298/00dad6ac-bc7f-11e6-8bef-51b6376648a4.png">

We could inject it using CSS based on the classname of the code block.

![code-blocks-deo](https://cloud.githubusercontent.com/assets/1493671/20967379/7ad88760-bc7f-11e6-80c2-f1a4e578f603.png)

raspo:
I can work on PR for this.
Here is a proof of concept hacked together with the inspector. 
The header is inspired by what can be found on the [react docs](https://facebook.github.io/react/docs/state-and-lifecycle.html).

<img width="895" alt="screen shot 2016-12-07 at 12 56 07 pm" src="https://cloud.githubusercontent.com/assets/927264/20986237/9c1487d8-bc7c-11e6-82e1-cf3d1cfb06db.png">

```css
div.highlighter-rouge:before {
    content: 'blah';
    display: block;
    height: 18px;
    line-height: 18px;
    background: #fbf4d0;
    font-size: 12px;
    font-weight: bold;
    padding: 0 10px;
    color: #b0a673;
}

.language-js.highlighter-rouge:before,
.language-javascript.highlighter-rouge:before {
    content: 'Javascript';
}

.language-sh.highlighter-rouge:before {
    content: 'Shell';
}
```

We could even consider making the "header" a big bigger and move the copy button to the top-right.
Thoughts?

existentialism:
Nice T, maybe just a bit of top/bottom padding (3px) on the header and smaller font?

fredericmarx:
I think this is a good opportunity to make the code snippets more beginner friendly.

One thing I would propose is getting rid of the `$`-symbol altogether. With the language indicator in place, it doesn't really add any value and is only confusing for people unfamiliar with that convention.

Another idea would be to add a link to a basic "What even is npm"-101 tutorial. This might be useful for people whose first point of contact with npm is Babel.

This is what this could look like:

![babel](https://cloud.githubusercontent.com/assets/954294/20990991/2b99e5ce-bcdc-11e6-98b7-3be4ab6437a8.png)

[...]

PR #1066, opened 2016-12-08
Indicate language in code snippets

raspo:
Solves #1049 
- Adds language indicator to code examples.
screen shot 2016-12-08 at 2 10 37 pm

- Remove `$` sign from shell commands.

- Fix inconsistencies in the way code blocks are structured.

As discussed in https://github.com/babel/babel.github.io/issues/1049#issuecomment-265823164, Markdown and Liquid templates produce different markups.
The vast majority of code snippets are generated via markdown and only two pages use the `{% highlight %} ` tag.

My solution was to replace this syntax:
```
{% highlight javascript %}
  // some code here
{% endhighlight %}
```

With this ugliness:

    {% capture c %}
    ```js
      // some code here
    ```
    {% endcapture %}{{ c | markdownify }}

I think it is very ugly and hackish-looking but it does solve all inconsistencies in markup letting us write actual markdown, and again, it should only affect two pages.


existentialism:
@raspo and I are also discussing an option which should eliminate the need for the capture/markdownify jazz.

raspo:
Pushed new changes, so now we can do 

    <div markdown="1">
    ```js
      // some code here
    ```
    </div>

instead of the `capture` stuff.

In order to enable markdown inside HTML, the file itself needs to have a `.md` extension, that's why I moved the content of `index.html` into `home.md`. And for the record, we can't rename `index.html` to `index.md` or the pagination plugin will complain.

/cc @existentialism

docs/plugins.md
```diff
@@ -248,10 +248,10 @@ vs
 
 This also works with scoped packages
 
-```js
+```json
 {
-  presets: ["@org/babel-preset-name"], // actual package
-  presets: ["@org/name"] // shorthand name
+  "presets": ["@org/babel-preset-name"], // actual package
```
hzoo: Is it ok to make it json if we add comments? I think it was highlighting wrong

hzoo: haha cool then

hzoo: Oh right so true. We might need to remove all comments from configs then.. I wish we could just strip them out before the clipboard haha

raspo: I split the code blocks so it's consistent with the rest of the paragraph.

screen shot 2016-12-11 at 9 20 32 am

index.html
```diff
@@ -6,347 +6,5 @@
...
-      <div class="btn-wrapper">
-        <a href="{{ site.baseurl }}/users/" class="btn btn-sm btn-featured">Meet more Users</a>
-      </div>
-    </div>
-  </div>
-
-  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/slick-carousel/1.4.1/slick.css">
-</div>
+{% capture page %}{% include home.md %}{% endcapture %}
+{{ page | markdownify }}
```
xtuc: Do we need to move this page?

xtuc: @raspo since #1092 could you move index.html back and rename it to index.md?

existentialism: Merged!

xtuc:
I moved back `home.md` in the `index.md`.

raspo:
I fixed what looked like a regression from #1087. 

LGTM

existentialism:
Nit: I think it looks better with the in/out boxes being equal heights:

image

(This PR left/current right)

raspo:
fixed alignment

screen shot 2016-12-17 at 1 16 01 pm

existentialism:
👍

docs/plugins.md
```diff
@@ -261,7 +260,7 @@ This means if two transforms both visit the "Program" node, the transforms will
 - Plugin ordering is first to last.
 - Preset ordering is reversed (last to first).
 
-```js
+```json
 "plugins": [
   "transform-decorators-legacy", // will run first
```
hzoo: We should remove comments for json snippets right? (and move them somewehre else)

hzoo:
If copy, highlighting is still good + view on tablet/mobile 👍

docs/plugins.md
existentialism: @raspo posted earlier that they still render fine: https://github.com/babel/babel.github.io/pull/1066#discussion_r91809624, but it might be better to re-word some of these areas. In this particular case maybe something like:

====

This means if two transforms both visit the "Program" node, the transforms will run in either plugin or preset order.

* Plugins run before Presets.
* Plugin ordering is first to last.
* Preset ordering is reversed (last to first).

For example:

```json
{
  "plugins": [
    "transform-decorators-legacy",
    "transform-class-properties"
  ]
}
```

Will run `transform-decorators-legacy` then `transform-class-properties`.

It is important to remember that with presets, the order is _reversed_. The following:

```json
{
  "presets": [
    "es2015",
    "react",
    "stage-2"
  ]
}
```

Will run in the following order: `stage-2`, `react`, `es2015`.

This is mostly the result of ensuring backwards compatibility, since most users list "es2015" before "stage-0". For more information, see [notes on potential traversal API changes](https://github.com/babel/notes/blob/master/2016-08/august-01.md#potential-api-changes-for-traversal).

====

Thoughts?

existentialism:
Things look ok on tablet, copying seems to work (at least on iOS devices). The copy buttons are actually hidden on (hidden-xs) smaller viewports.

docs/plugins.md
xtuc: I think we shouldn't use comment at all in code snippets. Most of the users will remove it anyway.

existentialism:
@xtuc @hzoo @raspo took a stab at removing the comments from snippets on plugins page, let me know what you think

raspo:
@existentialism Nice work on rewording that *"Yes this is confusing"* part 👍

existentialism:
@raspo nice work!

xtuc:
@raspo nice job :+1:

PR #1067, opened 2016-12-09
Fix Broken links

maxdow:
Shortcuts in the main page doesn't work. This is not perfect because links are hardcoded (it will break if the domain change ) but it should work

### DIFF
--- a/_posts/2016-12-07-the-state-of-babel.md
+++ b/_posts/2016-12-07-the-state-of-babel.md
@@ -11,10 +11,10 @@ custom_js_with_timestamps:
 - docs.js
 ---
 
-- [Some History](#some-history)
-- [Current Status](#current-status)
-- [The Future](#the-future)
-- [Community](#community)
+- [Some History](https://babeljs.io/blog/2016/12/07/the-state-of-babel#some-history)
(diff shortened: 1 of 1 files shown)


xtuc:
Links aren't broken, they are pointing to anchors in the page.

Did you encourter this issue on the blog post on the Babel's website (https://babeljs.io/blog/2016/12/07/the-state-of-babel) ?

maxdow:
Exact but it doesn't work when anchors are not in the page, like in the main blog page ( https://babeljs.io/blog/ ) . This is the issue , on babel's website yes

xtuc:
Ok yes, good point.

I guess we could also update the description of the post to avoid theses links on the page.

/cc @hzoo

maxdow:
Absolutely, I think it's a better fix ;)

hzoo:
Yeah its just auto adding the first thing there so I dono 😄 we could remove.

xtuc:
Thanks @maxdow :+1:

maxdow:
You're welcome, thanks for all the hard work and thanks @hzoo for this really good recap on the state of babel

Issue #1042, opened 2016-12-06
Docs Sidebar


hzoo:
Copied from https://github.com/babel/babel.github.io/issues/930#issuecomment-256480579

> I think the docs need a sidebar, like the React docs. @danharper 

![screen shot 2016-12-06 at 2 39 37 pm](https://cloud.githubusercontent.com/assets/588473/20940745/d79b0644-bbc1-11e6-835f-4e1d413a591c.png)

```
Primary Nav (left):

* Docs
* Try it out
* Community
* Blog

Secondary Nav (right):

* Twitter
* GitHub
```

Then on the Docs page, the following sidebar:

```
# Quick Start

* Browser
* Node
* Plugins
* Polyfills
* Caveats (or.. Browser Support)
* Editors
* Linting
* Babili (Minification)

# Advanced Usage

* Options
* CLI
* babel-node
* In-Browser Compilation
* Babel API
```

I'm unsure where a "Learn ESXXX" would go, but wherever it goes perhaps just give "highlights" and link off-site for the details. Plenty of community resources for that.

bernard-lin:
Willing to give this a shot! @hzoo

existentialism:
@bernard-lin thanks! join us on slack if you have any questions

hzoo:
This will be done in https://github.com/babel/website/issues/1485 but will iterate

PR #1073, opened 2016-12-10
Grouping elements in navbar

### LINKED ISSUE #1060: Link core-packages page
Following #1026. We are missing a link to it.

The page is already online http://babeljs.io/docs/core-packages/.

We could add it in the docs dropdown?

screenshot from 2016-12-08 09-08-27

Or we could replace this page http://babeljs.io/docs/usage/api/ which has the same in it.

@hzoo @existentialism Any ideas?

xtuc:
Fixes #1060

#### Desktop:
nav-desktop

 #### Collasped:
nav-mobile

Given #1042 this is a quick fix.

Navigation data has a bit changed:
* actual item
* subnav
    * name (optionnaly set a group name)
    * items
        * actual item

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -3,24 +3,31 @@
 - title: Docs
   url: /docs
   subnav:
-  - title: Setup
-    url: /docs/setup/
-  - title: Plugins
-    url: /docs/plugins/
-  - title: Polyfill
-    url: /docs/usage/polyfill/
-  - title: Caveats
-    url: /docs/usage/caveats/
-  - title: Editors
-    url: /docs/editors
-  - title: API
-    url: /docs/usage/api/
-  - title: CLI
-    url: /docs/usage/cli/
-  - title: babel-register
-    url: /docs/usage/babel-register/
-  - title: .babelrc
-    url: /docs/usage/babelrc/
+    - items:
+        - title: Setup
+          url: /docs/setup/
+        - title: Editors
+          url: /docs/editors
+        - title: Caveats
+          url: /docs/usage/caveats/
+    - name: Packages
+      items:
+        - title: API
+          url: /docs/core-packages/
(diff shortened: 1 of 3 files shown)


_data/navigation.yml
```diff
@@ -3,24 +3,31 @@
...
+        - title: Setup
+          url: /docs/setup/
+        - title: Editors
+          url: /docs/editors
+        - title: Caveats
+          url: /docs/usage/caveats/
+    - name: Packages
+      items:
+        - title: API
+          url: /docs/core-packages/
```
xtuc: `API` is pointing babel's core packages. We can rename it if you want.

existentialism:
As an alternative to the indention, what are your thoughts about:

image

Also, I think renaming "API" -> "Core" under Packages is a little clearer? @hzoo?

xtuc:
Yes, Looks better.
Have you tried with the collasped version ?

existentialism:
Yeah, we can do something similar for now... when we attack the sidebar we may have to rethink a bunch of this anyway:

image

xtuc:
It seems great. Would you push it there so you can merge this.

existentialism:
@xtuc pushed changes, check em out

xtuc:
Nice job, thanks!

LGTM, can we merge this ?

hzoo:
Awesome looks good.

Although now I think maybe would could rename a few things?

Should core be babel-core and polyfill babel-polyfill or is the shorthand better?

Maybe should remove the browser in usage since it redirects to setup anyway?

existentialism:
Good point re: browsers, agree we should drop.

And yeah, while I think Core and Polyfill _look_ better, it might be better to go with babel-core and babel-polyfill to eliminate any chance someone might be confused (esp since babel-register is below, and Register would be pretty strange by itself).

I'll update!

PR #1078, opened 2016-12-12
Tweak docs menu items

existentialism:
Slight re-ordering to make things look a little better:

image


hzoo:
should we do `babel-cli` then? 😄 

Maybe move .babelrc above cli?

existentialism:
Yeah, you're right :)

Updated!

PR #1086, opened 2016-12-15
babelrc.md: export env variables in example

kenzierocks:
POSIX shells won't pass variables as environment variables unless they are explicitly exported, with the exception of the `$PATH` variable.


xtuc:
Yes, thanks. LGTM :+1:

PR #1084, opened 2016-12-13
Include babel-preset-env documentation

xtuc:
Fixes #1081.

PR #1087, opened 2016-12-15
Fix preset-env doc link

xtuc:
Since #1084 is merged the page is available.


existentialism:
We should update the env link on the Plugins page too

hzoo:
We can remove that too, either is good

PR #1092, opened 2016-12-15
Change pagination settings to point to blog

existentialism:
This PR should make things easier for https://github.com/babel/babel.github.io/pull/1066 and eliminates the need to have a markdown include.

This keeps pagination settings the same for the blog (currently set to 1, but will likely want to increase as we get more blog posts), and allows us to rename index.html to index.md without the jekyll error:

image

/cc: @raspo


raspo:
Nice and clean! :+1:

xtuc:
There is a big warning on the Jeykyll documentation about this (https://jekyllrb.com/docs/pagination).

jekyll-pagination.
We heavily rely on permalinks (https://github.com/babel/babel.github.io/search?utf8=%E2%9C%93&q=permalink).

I'm not sure why, do you know?

existentialism:
The warning is specific to page where you're using pagination (i.e. our blog) since it will generate multiple pages (blog/index.html, blog/page2/index.html, etc) depending on how many posts you have.

We don't use pagination (as of now) for docs/*, so we should be fine there.

We can drop the permalink from the blog/index front matter anyway.

raspo:
:shipit:

xtuc:
:shipit: :+1:

PR #1097, opened 2016-12-19
Fix malformed <p> tag in index

Pablo-Merino:
I found a malformed <p> tag in the index page, on the "ES2015 and beyond" section. This is just a simple pull request to fix it :)


hzoo:
👍 good catch!

Pablo-Merino:
@hzoo yep, I thought so 😄

PR #1104, opened 2016-12-24
Try in REPL

### LINKED ISSUE #1017: Open in REPL button?
screen shot 2016-12-03 at 11 35 08 pm

---

We have the copy button via https://github.com/babel/babel.github.io/pull/998 now

I was wondering if we could label certain code snippets somehow and then also add a button that opens up a link to the repl with the code that was in the snippet?

Could make http://babeljs.io/docs/learn-es2015/ more interesting

xtuc:
Fixes #1017.

try-magic-icon

Opens a new window with the REPL and the code in it.


_sass/components/_buttons.scss
```diff
@@ -80,6 +80,18 @@
   }
 }
 
+.btn-open-in-repl {
+  position: absolute;
+  bottom: 0;
+  padding: 1rem;
+  right: 55px; // size of the button.btn-clipboard
```
xtuc: I'm not a CSS expert. This is causing some issues:
bug.

Would it possible to use a flexbox wrapping both buttons?

scripts/open-in-repl.js
```diff
@@ -0,0 +1,33 @@
+/* global jQuery */
+
+(function ($) {
+    'use strict';
+
+    var presets = [ "es2015", "stage-0", "react" ];
```
xtuc: This is meant to be available on the entire website. I tried to be as compatible as possible while not including all presets...

I guess you could use the `package` value in the header to determine which preset is using in the examples.

scripts/open-in-repl.js
```diff
@@ -0,0 +1,33 @@
...
+      + '<strong class="btn-icon__label small">Try</strong>'
+      + '</button>'
+    );
+
+    button.on('click', function() {
+      var $code = $(this).parent().find('code');
+
+      var url = '/repl/#'
+              + '?babili=false'
+              + '&evaluate=false'
```
xtuc: Auto-evaluation is disabled since some examples don't produce an output or can't run.

scripts/open-in-repl.js
```diff
@@ -0,0 +1,33 @@
...
+      + clipboardIcon
+      + '<strong class="btn-icon__label small">Try</strong>'
+      + '</button>'
+    );
+
+    button.on('click', function() {
+      var $code = $(this).parent().find('code');
+
+      var url = '/repl/#'
+              + '?babili=false'
```
xtuc: We know the source of the documentation (`package_source` in the header). We could enable babili for example from the babili documentation.

scripts/open-in-repl.js
```diff
@@ -0,0 +1,33 @@
+/* global jQuery */
+
+(function ($) {
+    'use strict';
+
+    var presets = [ "es2015", "stage-0", "react" ];
+
+    var selector = '.language-js,.language-javascript';
```
xtuc: This functionality is only for JavaScript code for obvious reasons.

satya164:
May be the triangular run icon is a better icon instead of a magic wand?

hzoo:
Good point about the icon @satya164 - not sure the best icon but yeah maybe a play/run icon is more clear. Or we should just increase the text to say "try it out"

existentialism:
What about moving the actions into the header?

image

It solves the issue of dealing with wrapping around copy/repl links on the last few lines of code.

scripts/open-in-repl.js
existentialism: Yeah, decorating a page with `package_source` as a class (i.e. `doc--babel`, `doc--babili`) is probably good first step. We could even enable more granular specific repl settings via convention or a front-matter property.

_config.yml
```diff
@@ -34,6 +34,7 @@ default_js:
   - "https://cdn.jsdelivr.net/docsearch.js/2/docsearch.min.js"
   - "/scripts/search.js"
   - "/scripts/clipboard.js"
+  - "/scripts/open-in-repl.js"
```
existentialism: To avoid scanning and modifying the same code blocks twice, we may want to consider combining clipboard/open-in-repl?

scripts/open-in-repl.js
xtuc: Yes that would be great. I added the comment as a reminder because it will be a separate PR.

_config.yml
xtuc: I don't know, I like the separation of responsibility here.

We do update the same JavaScript code blocks twice but I think it's not a big deal.

xtuc:
@existentialism I personally prefer the button in the right corner.

@satya164 @hzoo Maybe we could just write "Try in REPL" without an icon?

satya164:
@xtuc an icon still looks nice IMO

xtuc:
@satya164 Sure. Play button might be common for that.

existentialism:
I'm fine with keeping it in the lower right-hand corner, though I'd definitely increase the space between them.

Also, we'll need to fix cases like:

image

xtuc:
@existentialism Yes, it might be better in the header. I was worried about the size of the buttons in it.

xtuc:
try-in-repl-buttons

@satya164 Yes, the play icon looks better

hzoo:
Nice that looks good!

satya164:
Looks nice 👍

hzoo:
Oh I do like it in the header as well!

existentialism:
@xtuc @hzoo did a quick example at https://github.com/existentialism/babel.github.io/tree/1104-experiment, thoughts?

hzoo:
😍

xtuc:
I also :heart_eyes:. @existentialism could I steal you that?

existentialism:
@xtuc of course, I did a little bit of testing in some browsers, but could def use a second set of eyes.

xtuc:
LGTM. @hzoo ok for merge?

hzoo:
👍 just need to fix copy button

hzoo:
This is awesome! Thanks everyone for the feedback

PR #1105, opened 2016-12-24
Corrects "plugins" to "presets" in .babelrc

justafish:
Using "plugins" threw an error for me!

### DIFF
--- a/docs/usage/babelrc.md
+++ b/docs/usage/babelrc.md
@@ -11,7 +11,7 @@ All Babel API [options](/docs/usage/options) except the callbacks are allowed (b
 
 ```json
 {
-  "plugins": ["transform-react-jsx"],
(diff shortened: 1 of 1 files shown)


brigand:
@justafish what error?

xtuc:
The documentation on the website is correct.

Could you please fill an issue in Babel: https://github.com/babel/babel/issues/new.

hzoo:
hey @justafish, yeah the documentation itself is correct because there is no `babel-preset-transform-react-jsx` preset.

There is a `babel-plugin-transform-react-jsx` though which is why it's in plugins.

Maybe you are just using a preset like `es2015`, etc. It was just an example so maybe we want to use a preset or both in the docs if it's not clear? Can make another PR for that if you want

Issue #1106, opened 2016-12-25
Missing Babylon documentation


xtuc:
We could add [Babylon's README](https://github.com/babel/babylon/blob/master/README.md) in the core-package > other-packages.

![screenshot from 2016-12-25 15-14-04](https://cloud.githubusercontent.com/assets/1493671/21471472/e52ddbcc-cab4-11e6-8f2f-d235ec149686.png)

PR #1107, opened 2016-12-25
Add Babylon documentation

xtuc:
Fixes issue #1106 

babylon

### DIFF
--- a/docs/core-packages/babylon.md
+++ b/docs/core-packages/babylon.md
@@ -1,0 +1,13 @@
+---
+layout: docs
+title: A JavaScript parser
+description: Babylon is a JavaScript parser used in Babel
+permalink: /docs/core-packages/babylon/
+package: babylon
+package_source: babylon
+---
+{% capture readme %}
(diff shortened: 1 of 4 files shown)


docs/core-packages/babylon.md
```diff
@@ -0,0 +1,13 @@
+---
+layout: docs
+title: A JavaScript parser
+description: Babylon is a JavaScript parser used in Babel
+permalink: /docs/core-packages/babylon/
```
xtuc: Babylon is not exactly part of Babel's core.

hzoo: I would consider it as one (especially since it used to be in the monorepo). Since Babel = parser, transformer(s), generator

existentialism: The title is strange to me... maybe:

```
title: Babylon
description: Babel's JavaScript parser
```

xtuc: Oh yeah, thanks!

PR #1110, opened 2016-12-28
Move documentation from preset-react

xtuc:
Documentation should be moved in the package README, see babel/babel#5051


existentialism:
👍

PR #1116, opened 2016-12-31
Fix 404 for babel-register menu item

fkling:
It appears that URLs are treated differently if they contain a trailing slash. `/docs/core-packages/babel-register` works, `/docs/core-packages/babel-register/` results in a 404.

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -25,7 +25,7 @@
         - title: babel-cli
           url: /docs/usage/cli/
         - title: babel-register
-          url: /docs/core-packages/babel-register/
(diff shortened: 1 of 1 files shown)


hzoo:
Hmm that's super weird and we should investigate @xtuc but will merge to fix the issue - thanks so much for looking into it @fkling!

xtuc:
Thanks for pointing out the issue.

To be consistent in the website I prefer adding a trailing slash for each URL listed in the core-package page (https://github.com/babel/babel.github.io/blob/7b6348f2ab91ec7954e0ae1e26de50fc8402b4c3/docs/core-packages.md)

PR #1123, opened 2017-01-06
Reword "serializable JSON"

kix:
"Serializable JSON" doesn't make much sense to me: JSON is a serialized format, while this sentence sounded like you could serialize JSON further. Perhaps, this wording would be better?


hzoo:
👍 Thanks!!

PR #1126, opened 2017-01-07
Added "Contribute" page with active issues.

### LINKED ISSUE #1099: Add contributing section to the website
From discussion in slack:

- List `beginner-friendly` and `help wanted` tagged issues (ex: https://hapijs.com/contribute), we might need to do a round of issue tagging/cleanup first.

- At the minimum link to [CONTRIBUTING.md](https://github.com/babel/babel/blob/master/CONTRIBUTING.md), but we probably want to integrate it? 

- Link to resources like [babel-handbook](https://github.com/thejameskyle/babel-handbook/), [AST explorer](https://astexplorer.net/), etc.

hpierce1102:
Fixes #1099

The links to AST-Explorer and the handbook are included in the CONTRIBUTING.md content, Though maybe they deserve their own callouts?

Maybe this shouldn't be part of the top level nav, but inside the docs menu?

babel-contributing


hzoo:
so awesome! thanks for "contributing" this @hpierce1102 (heh)!

I wanted to revamp the contributing anyway so it's ok

scripts/contributing.js
```diff
@@ -0,0 +1,36 @@
+/* global jQuery*/
+
+(function($){
+    var githubIssuesEndpoint = 'https://api.github.com/repos/babel/babel/issues';
```
hzoo: May not have to be in this pr but we could actually link up issues in other repos too

babel/x such as babili, babel-preset-env, babylon, etc

hzoo:
Not sure yet but maybe we should move contributing.md to the website and just link to it. Maybe the open issues can be put front and center (or we make another separate page as well)?

scripts/contributing.js
hpierce1102: I initially didn't include them because I wanted to know how well this would be received and I was a bit worried about sending too large of a PR.

I'll add issues for more repos over the next day over the next day or so.

hpierce1102:
I like the idea of linking from babel/babel to the website for this kind information.  If I understand correctly (and I may not, this is my first experience using Jekyll), that would give us more control over how this information is presented.

For instance, I see a picture a decent layout keeping the "Setup | Running linting/tests | Writing tests | Internals" and "Before contributing, please read our code of conduct." sections above the fold with active issues following.

I'll add another commit to this addressing those points.

scripts/contributing.js
xtuc: Note that:
> This endpoint may also return pull requests in the response. If an issue is a pull request, the object will include a pull_request key.

I don't know if we want PR there.

scripts/contributing.js
```diff
@@ -0,0 +1,36 @@
...
+            success : function(data){
+                for(var i = 0; i < data.length; i++){
+                    $('.helpWantedIssues').append('<li><a href="' + data[i].html_url + '"><span class="issueNumber">#' + data[i].number +'</span></a><span>' + data[i].title + '</span></li>');
+                }
+            },
+            error : function(xhr, status, error){
+                throw error;
+            }
+        }
+    );
```
xtuc: We could use one single call for both labels.

```
https://api.github.com/repos/babel/babel/issues?labels=beginner-friendly,help%20wanted
```

> labels (string):	A list of comma separated label names. Example: bug,ui

scripts/contributing.js
xtuc: Do you know if there quotas on API calls?

scripts/contributing.js
```diff
@@ -0,0 +1,36 @@
...
+(function($){
+    var githubIssuesEndpoint = 'https://api.github.com/repos/babel/babel/issues';
+
+    $.ajax(githubIssuesEndpoint, {
+            data : {
+                labels : 'beginner-friendly'
+            },
+            success : function(data){
+                for(var i = 0; i < data.length; i++){
+                    $('.beginnerFriendlyIssues').append('<li><a href="' + data[i].html_url + '"><span class="issueNumber">#' + data[i].number +'</span></a><span>' + data[i].title + '</span></li>');
```
xtuc: Do you mind if the whole text is clickable?

_data/navigation.yml
```diff
@@ -32,3 +32,5 @@
   url: /blog/
 - title: FAQ
   url: /docs/faq
+- title: Contribute
+  url: /docs/contributing
```
xtuc: There is an issue for detecting the currently active item in the menu.

I will take a look, I think @existentialism already mentionned it.

i

scripts/contributing.js
hpierce1102: Using more than one label seems to restrict the query more by only including issues with both "beginner-friendly" and "help wanted". So that url only provides 2 issues while this url:

    https://api.github.com/repos/babel/babel/issues?labels=help%20wanted

provides 9 issues.

I don't think we'll be able to combine these.

scripts/contributing.js
hpierce1102: >  Note that: ... I don't know if we want PR there.

Yeeeeeeep. That was an oversight. I even made a note to check for that. I'll address that in a future commit.

>  Do you know if there quotas on API calls?

Great question. After looking into it there's a 60 request/hour/IP limit for unauthenticated requests according to the [Github docs](https://developer.github.com/v3/#rate-limiting). There's a caching header on the response of 60 seconds, so theoretically we could do a single call as much as we want. Since I've used 2 here, there's a _possible_ issue if someone were to refresh every minute for long enough, but that doesn't seem practical. During my testing, I wasn't able to use more than 20/hour. 

I guess I don't see this as a practical enough issue _as is_ - though this might be a reason to split up repos across multiple pages rather than aggregating all issues for all repos on the one page.

hpierce1102:
Few changes:

- Embedded CONTRIBUTING.md in order to move open issues up to a more prominent position.

- Removed the "NOTE: DO NOT OPEN..." - I don't think it makes sense in the context of the website personally (open to feedback).

- Made entire issue text clickable.

babel-contributing2

scripts/contributing.js
xtuc: Ok, seems ok.

scripts/contributing.js
xtuc: Yes, that's right.

docs/contributing.md
```diff
@@ -7,7 +7,23 @@ custom_js_with_timestamps:
 - contributing.js
 ---
 
-{% include babel/CONTRIBUTING.md %}
```
xtuc: I would prefer keeping the documentation in Babel/babel repo. We can still update the file there.

scripts/contributing.js
xtuc: It seems possible to query for issues in an entier org:
> GET /orgs/:org/issues

hpierce1102:
babel-contributing3

I couldn't figure out this bit - multiple tabs are highlighted on the babel-preset-env page. I suspect it's the same issue as the docs being highlighted incorrectly as was mentioned earlier.

scripts/contributing.js
hpierce1102: The docs mention authentication for that endpoint - and I'm getting a 404 when trying to access it (despite *expecting* a 401):

https://api.github.com/orgs/babel/issues

DrewML: I get a 404 for that endpoint even when using my API key :thinking:. Same with my other repositories. May just be broken.

existentialism:
@hpierce1102 yeah, the website is currently using:

```
{% if page.url contains item.url%}
```

to highlight the current section and anything underneath it.

Since we're treating `contributing` as a top level nav item in the bar, let's place it at the root (./contributing/ vs ./docs/contributing)?

existentialism:
Aligning both the issue number boxes and the edge of the text helps scan-/read-ability:

image

xtuc:
This looks awesome!

@existentialism I think It would looks even prettier with the text centered in the yellow boxes.

@hpierce1102 I think we could add a fallback link to Babel's issue tracker in case the Github API limit is exceeded.

The limit isn't that high, it's IP based, in France some users from the same ISP can share the same IP address. It would be bad If people coudn't contribute to Babel because of this :)

xtuc:
highlighting Issue in at #1141

hpierce1102:
The failing tests are pointing out that relative links in the documentation are 404'ing because they're relative to the git repo, and not to the website. I guess we have a few possible solutions here:

1. Edit the links to use absolute links in each repo's documentation files.

Not great. That means we'd have to enforce rules on the documentation in those repos for the website's benefit. Even then, clicking a link from the website and being directed to a markdown file on Github doesn't seem like a great experience - I think it would make more sense to keep them on the website.

2. Add pages on the website for each of the documentation pages in repos.

Not great either. It would be super important to keep documentation in the repos in line with the website, otherwise links would start to 404 (or otherwise fail to link, whatever the behavior of *that* is) due to changes in those repos.

3. Embed the documentation on the website, changing links as we please.

This seems like the best option IMO. We could maintain links here without worrying about changes to other repos. I know @xtuc had some opposition to this - so maybe the winning solution is to maintain embedded copies of the documentation on the website **and** in their respective repo? We would need to keep the documentation on the website updated, but that seems like a better solution than occasionally letting links break entirely, right?

4. Something else...?

xtuc:
I'm not sure [this](https://travis-ci.org/babel/babel.github.io/builds/195040874#L557) should be failing on the website. Files are actually in babel/babel repo but included via git submodules, it should be fine.

/cc @existentialism

aaronang:
I have played around with this branch and here I will summarize my observations.

1. It would be nicer if the issue titles wouldn't continue on the next line right below the issue number.

alignment_issue_titles

2. Shown in image above, the `beginner-friendly` and `help-wanted` sections have some margin/padding to their left, which is also kept on the mobile version, which IMO should then be removed. See image below. 

alignment_issue_titles

3. When on one the contribute pages, `Contribute` in the navigation bar should be highlighted.

contribute_not_highlighted

4. In the image above, we can see that the title is "Contributing", but in the navigation bar it is "Contribute".

5. Fallback works, but does not shows a correct message when there are just no `beginner-friendly` or `help-wanted` issues.

no_labeled_issues

6. Nitpicking: In the fallback message, I would change "Beginner-Friendly" to `beginner-friendly` and "Help-Wanted" to `help-wanted`.

7. The code block shows a `COPY` button that looks different from the one on the website. I am assuming that the branch is just a bit out-of-date, and that it will be fine once it is merged? 

copy_button

## Conclusion

I think we could merge this, and create a new `beginner-friendly` issue that lists the issues that I just pointed out. 

@hpierce1102 Thank you for your effort!

hpierce1102:
@aaronang 

> I think we could merge this, and create a new beginner-friendly issue that lists the issues that I just pointed out.

I'd agree with this - I've been tied up lately and haven't been able to pay as close attention to this as I'd like. I'll plan on contributing again in the future - but I think this is a good approach to unblock this.

PR #1137, opened 2017-01-15
Remove "class-constructor-call" plugin

xtuc:
Will be removed in Babel 7 (babel/babel#5119)


hzoo:
I know we normally delete it but since it's the docs it might be better to say unmaintained/removed - we don't want the website url to get remove (unless we create another page for removed proposals and redirect there)

xtuc:
Yes of course. I think we should have a deprecated / do-not-use label for the website.

hzoo:
But we should update it if you think it could be worded up now that we are removing it from the repo/unmaintained

xtuc:
There are already depreciation notes in both plugin list and on the plugins page. In the README on babel/babel there is also a note.

I think it's pretty obvious.

We could remove it from the plugin list but keep the page.

PR #1142, opened 2017-01-17
Make the REPL work when cookies are blocked.

goto-bus-stop:
If cookies are disabled, merely accessing `window.localStorage`
throws an error. So the REPL would fail to load when creating its
StorageService instance. This change moves the `localStorage`
property access inside the try-catch blocks, so the REPL can
still load even if `localStorage` is unavailable.

| Before | After |
|---|---|
| Before | After |

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -77,21 +77,24 @@
   /*
    * Long term storage for persistence of state/etc
    */
-  function StorageService () {}
-
-  StorageService.prototype.getStore = function () {
-    return window.localStorage;
+  function StorageService () {
+    try {
+      this.store = window.localStorage;
+    } catch (e) {
+      console.warn('Could not access localStorage. Disabling state persistence.');
+      this.store = null;
+    }
   }
 
   StorageService.prototype.get = function (key) {
     try {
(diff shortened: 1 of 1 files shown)


hzoo:
Oh nice, lgtm

scripts/repl.js
```diff
@@ -77,19 +77,21 @@
...
+
+  StorageService.prototype.getStore = function () {
+    return window.localStorage;
   }
 
   StorageService.prototype.get = function (key) {
     try {
-      return JSON.parse(this.store.getItem(key));
+      return JSON.parse(this.getStore().getItem(key));
     } catch(e) {}
```
xtuc: If think we should emit a warning here. It might be useful for debugging purposes.

goto-bus-stop: Sounds good! I changed it to log a warning in the constructor, so that it's only logged once.

xtuc:
Nice work @goto-bus-stop :+1:. OK for merging this? @hzoo @existentialism

existentialism:
:+1: thanks @goto-bus-stop!

PR #1148, opened 2017-01-24
REPL: Add env to available presets.

yavorsky:
This PR adds [env](https://github.com/babel/babel-preset-env) to the list.

It also adds new dropdown menu that emulates preset-env configuration: Ability to define browsers in [borwserslist](https://github.com/ai/browserslist)'s format and electron + node.js as additional options. 
After you change one of the options REPL recompiling all stuff into the **code that works on specified targets with the minimum possible transformation**.

I've created [babel-preset-env-standalone](https://github.com/yavorsky/babel-preset-env-standalone) - wrapper for original preset that doesn't change behavior but just making a bundle. So we can just add it with `Babel.registerPreset`.

Tested on last chrome, safari, FF, ie11, edge.


existentialism:
Awesome @yavorsky!

Should we hide the targets option by default, to avoid it from appearing then disappearing on initial load:

mm

existentialism:
Should we catch this error:

image

yavorsky:
@existentialism 
Yes, I'll update targets option to be hidden by default. Thanks 👍

I don't think we need to hide browserslist error. It's very useful to understand that query has the wrong format. I can write something like `current chrome and safari` and without errors think that it's correct behavior.

hzoo:
Awesome - and maybe we could turn on the debug option on for logging? can be another pr

yavorsky:
Thanks, guys!
@hzoo Cool! Thinking where is better to output it. Just in the console or make additional section the prettiest way... 🤔

hzoo:
I was thinking just the console for now is good but would be nice to have a UI as well

Daniel15:
@yavorsky 

> I've created [babel-preset-env-standalone](https://github.com/yavorsky/babel-preset-env-standalone) - wrapper for original preset that doesn't change behavior but just making a bundle.

I'm happy to merge this directly into babel-standalone if you send a pull request 😃

yavorsky:
@Daniel15 Sure! But it weights about 3mb, currently working on reducing the size.

Daniel15:
We could bundle it with babel-standalone but make it a separate file, like how babili-standalone is separate. People that want to use it would load both `babel.js` and `env.js` (or whatever we call it), similar to how things are currently.

PR #1153, opened 2017-02-03
Update plugins.md

chicoxyzzy:


docs/plugins.md
```diff
@@ -50,7 +50,7 @@ Any transforms in stage-x presets are changes to the language that haven't been
   </p>
 </blockquote>
 
-The [TC39](https://github.com/tc39) categorises proposals into 4 stages:
+The [TC39](https://github.com/tc39) categorizes proposals into following stages:
```
hzoo: the following

docs/plugins.md
```diff
@@ -50,7 +50,7 @@ Any transforms in stage-x presets are changes to the language that haven't been
   </p>
 </blockquote>
 
-The [TC39](https://github.com/tc39) categorises proposals into 4 stages:
```
existentialism: Maybe a nit, but are we replacing the British spelling for words? I personally don't have a preference, only that we be consistent

hzoo: We can change where possible but there's probably a good amount?

PR #1159, opened 2017-02-09
Add mini-repl

### LINKED ISSUE #1144: Homepage: add the REPL to the top
Anyone can help with this but @existentialism made a cool mock of it in photoshop already as well

pasted image at 2017_01_13 02_11 pm

(can fix the margins between the repl to be smaller)

existentialism:
Fixes #1144

mm

(note, gif above has some visual artifact issues)

Desktop:
image

Mobile:
image

Just some polish items to tweak.

Would love feedback, and/or additional examples to use in the demo?

TODO:

- [ ] Scale font a bit based on the number of lines returned from compiled output?

### DIFF
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -94,7 +94,8 @@ var BABEL_MINI_REPL = (function() {
     simulateKey();
   }
 
-  function showError(babelError) {
+  function showError(editor, babelError) {
+    editor.setValue('');
     $('.hero-repl__error').text(babelError).addClass('hero-repl__error--visible');
   }
 
@@ -112,7 +113,7 @@ var BABEL_MINI_REPL = (function() {
         babelrc: false,
       });
     } catch (e) {
-      showError(e.message);
+      showError(targetEditor, e.message);
     }
 
     if (transformed) {
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -1,10 +1,12 @@
 var BABEL_MINI_REPL = (function() {
   var miniReplExamples = [
     '[1, 2, 3].map(n => n + 1);',
-    'var [a, ,b] = [1,2,3];',
+    'var [a,,b] = [1,2,3];',
     'function f(x, y = 12) {\n  return x + y;\n}',
-    'const x = {\n  ["i" + bar]: 5,\n};',
-    'async function foo() {\n  await bar();\n}',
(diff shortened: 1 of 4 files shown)


hzoo:
Probably best to `hub checkout https://github.com/babel/babel.github.io/pull/1159` this to test locally but gifs are good!

existentialism:
Added showing error:

mm

xtuc:
This looks awesome @existentialism !

screenshot from 2017-02-10 18-24-16

Sometimes the output is too big and difficult to read. What do you think if we limit the output in the mini-repl and add a message with a link to the big-repl?

> kid, you're too old to play in this sandbox.

hzoo:
screen shot 2017-02-10 at 12 40 38 pm

I think we need to clear the output when there's an error since it overlaps

yavorsky:
Awesome!!! 🎉
Maybe let's change titles at the top to more explicit? It's home page and many people could not understand this tautology.
Something like:
`Put the latest JS` -> `And get supported by all browsers version.`

scripts/index.js
```diff
@@ -1,4 +1,173 @@
+var BABEL_MINI_REPL = (function() {
+  var miniReplExamples = [
+    '[1, 2, 3].map(n => n + 1);',
+    'var [a, ,b] = [1,2,3];',
+    'function f(x, y = 12) {\n  return x + y;\n}',
+    'const x = {\n  ["i" + bar]: 5,\n};',
+    'async function foo() {\n  await bar();\n}',
```
hzoo: something like

```var obj = {
  shorthand,
  method() {
    return "a";
  }
};```
object shorthand

template strings

```
var name = "Bob", time = "today";
`Hello ${name}, how are you ${time}?`
```

exponentiation

`2 ** 3`

File: scripts/index.js [hunk 2]
```diff
@@ -1,4 +1,173 @@
...
+      });
+    } catch (e) {
+      showError(targetEditor, e.message);
+    }
+
+    if (transformed) {
+      hideError();
+
+      targetEditor.setValue(
+        // Remove 'use strict' just for demonstration purposes
```
Daniel15: I think there's an option somewhere to avoid `use strict` in the output.

Daniel15:
This is pretty nice! Thanks for building it 😃

PR #1163, opened 2017-02-14
algolia search api hidden

midhruvjaink:
Algolia search api and other tokens hidden with js Obfuscator for safe use of the api key

This way the api key cant be easily copied up by just viewing the source codes.


scripts/search.js
```diff
@@ -6,13 +6,13 @@
     var $searchInput = $($selector);
 
     // config is https://github.com/algolia/docsearch-configs/blob/master/configs/babeljs.json
+    var _0xac38 = ["\x64\x34\x32\x39\x30\x36\x62\x30\x34\x33\x63\x35\x34\x32\x32\x65\x61\x30\x37\x62\x34\x34\x66\x64\x34\x39\x63\x34\x30\x61\x30\x64", "\x62\x61\x62\x65\x6C\x6A\x73"];
     var $search = docsearch({
-        apiKey: 'd42906b043c5422ea07b44fd49c40a0d',
```
xtuc: Theses keys are meant to be public.

I'm not a fan of this because it will be more complicated to change the keys.

midhruvjaink: This way any one can misuse the key!

existentialism:
Thanks for the PR @midhruvjaink, but as @xtuc mentioned, the keys only have read-only access to the babel website search index provided by Algolia.

For mroe info check out:
https://community.algolia.com/docsearch/
https://www.algolia.com/doc/api-client/javascript/api-keys/#generate-key

/cc: @maxiloc

PR #1165, opened 2017-02-15
Trailing comma after object rest assignment in blog (React on ES6+)

adammockor:
This creates syntax error and it's against current spec.


hzoo:
Thanks! At the time we didn't know about this (and this is an older blog post)

PR #1171, opened 2017-02-17
Lint code examples

xtuc:


docs/faq.md
```diff
@@ -21,7 +21,7 @@ Arrow functions **are not** synonymous with normal functions. `arguments` and `t
 reference their *outer function* for example:
 
 ```javascript
-var user = {
+const user = {
```
xtuc: The eslint configuration prefers const. Given that most modern browser support `const`, I think that's fine.

hzoo: It's going to be run through Babel anyway though so definetely ok

hzoo:
I guess we would need to run it on the markdown in babel/babel - everything getting more complicated haha

xtuc:
Yes, this is the goal. I'm just POCing right now.

PR #1193, opened 2017-03-08
Get started seciton: Prevent h3 breaking.

yavorsky:
Turns
screen shot 2017-03-08 at 12 25 28
Into
screen shot 2017-03-08 at 12 27 44


hzoo:
^ should we figure out how to center the npm install or make the font bigger?

PR #1202, opened 2017-03-20
New User: Nau Studio

trongthanh:
https://naustud.io

We are a small production studio based in Sai Gon (Ho Chi Minh city), Vietnam. We focus on Front End and NodeJS full stack development. Our whole stack based on JavaScript (no PHP, .NET or Java whatsoever), and use babel to write ES6 and future ES in React, KeystoneJS, Meteor... projects.


trongthanh:
Thanks. 👍

PR #1206, opened 2017-03-21
fix link to const plugin

andimarek:
I guess the link for the const plugin should point to the transform plugin, as with the others links around it ... this fixes this.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -82,12 +82,12 @@ These plugins apply transformations to your code.
 
 ### ES2015
 
+- [check-es2015-constants](/docs/plugins/check-es2015-constants/)
 - [es2015-arrow-functions](/docs/plugins/transform-es2015-arrow-functions/)
 - [es2015-block-scoped-functions](/docs/plugins/transform-es2015-block-scoped-functions/)
 - [es2015-block-scoping](/docs/plugins/transform-es2015-block-scoping/)
 - [es2015-classes](/docs/plugins/transform-es2015-classes/)
 - [es2015-computed-properties](/docs/plugins/transform-es2015-computed-properties/)
-- [check-es2015-constants](/docs/plugins/check-es2015-constants/)
 - [es2015-destructuring](/docs/plugins/transform-es2015-destructuring/)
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -87,7 +87,7 @@ These plugins apply transformations to your code.
 - [es2015-block-scoping](/docs/plugins/transform-es2015-block-scoping/)
 - [es2015-classes](/docs/plugins/transform-es2015-classes/)
 - [es2015-computed-properties](/docs/plugins/transform-es2015-computed-properties/)
-- [es2015-constants](/docs/plugins/check-es2015-constants/)
```
existentialism: Actually it should probably just be renamed `es2015-check-constants`?

andimarek: No sure what you mean: all other links point the transform plugin. Only for const/let the link goes to the syntax checker, which itself links to the transformer. This changes the link directly to the const transformer.

hzoo: we already link to block scoping.

> [es2015-block-scoping](/docs/plugins/transform-es2015-block-scoping/)

the const checking was moved to the plugin that is already there.

andimarek: ah ...I see. Then my change doesn't make sense of course. I was confused to find a none transform plugin under the `transfomer` section and automatically assumed this is wrong.

I will close the PR.
Thanks!

andimarek:
closing because there is already a link to the block transformer plugin.

docs/plugins.md
existentialism: No worries @andimarek!

I agree it is semi-out of place in a section titled "Transform Plugins", but I think it's a fine exception for now (vs. creating a whole new "check" section).

Do you mind reopening and renaming it to [check-es2015-constants] to help others who might be confused what this links to?

andimarek:
reopening with the suggested rename

andimarek:
please squash when commit :-) .. otherwise it should be as suggested ..thanks

existentialism:
@andimarek thanks again!

PR #1215, opened 2017-04-01
Take me to Flavortown!

STRML:


existentialism:
👍  LOL exactly what I wanted to do before I had to run out for the day!

Thoughts on having a more limited set of Fieri/Wikipedia-plea/Babel-esque quotes (vs just pure fieri)?

STRML:
Sure I can trim them down to some of the funnier ones. Got some good alt quotes?

existentialism:
Here's a few random ones riffing on Wikipedia's, but you def have as much creative freedom here as you want! :)

- A personal appeal from major of Flavortown Guy.

- If everyone reading this donated $5 our fundraiser would be over today. Donate to keep all your modules transpiled for free.

- Imagine you had a nickel for every time Babel transpiled your code. We need those nickels.

- We're almost to Flavortown. Thank you."

- "Some people are just born to cook and talk. And transpile code."

STRML:
Updated

sorianog:
This is wonderful :fire: :fire: :fire:

PR #1225, opened 2017-04-06
Add debug option if env preset selected.

yavorsky:
With evaluate:

screen shot 2017-04-07 at 02 53 58

With huge output (scrollable):
screen shot 2017-04-07 at 02 53 29

`debug` option is appearing only when env preset was selected (like built-ins).

### DIFF
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 (function(babel, $, _, ace, window) {
   'use strict';
   var UPDATE_DELAY = 500;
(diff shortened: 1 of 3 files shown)


scripts/repl.js
```diff
@@ -204,6 +204,29 @@
     evt.stopPropagation();
   }
 
+  function envItemToString(item) {
+    return item.name + ' ' + JSON.stringify(item.targets);
+  }
+
+  function onEnvBuild(node, opts, envResult) {
```
existentialism: One thing to consider is, that once https://github.com/babel/babel-preset-env/pull/241 lands, we're likely going to follow that up with plugin-dependent debug output.

We might want to expose the debug output from directly from the plugin on `onPresetBuild` instead of needing to reconstruct it from structured data?

(And, maybe even stepping further back, @hzoo and I were discussing that maybe having debug output for _every_ transform would be really useful... so this might be the beginnings of a consistent way of accessing it?)

yavorsky: @existentialism Yeah. I think there would be a couple of changes after 2.0 merge. So we'll keep our fingers on the pulse. 

>We might want to expose the debug output from directly from the plugin on onPresetBuild instead of needing to reconstruct it from structured data?

Initially, I've chosen this way, but decided that we could come up with awesome UI idea to display plugins/built-ins neat (maybe in the table or with the better colorful console). 🌅

existentialism: Yeah, definitely! I think there's value in both raw debug + structured data.

yavorsky: Also, seems like we depend on https://github.com/yavorsky/babel-preset-env-standalone, so maybe move it to babel/babel-preset-env-standalone? I can ask him 🙂.

PR #1240, opened 2017-04-24
Place env debug to a separate container.

yavorsky:
Fixes https://github.com/babel/babel/issues/5657


existentialism:
Verified the issue is fixed, though I wonder if we want to tab out debug vs console output? Not sure that discussion needs to block this though

PR #1241, opened 2017-05-02
Added stricter clarification note about transform-react-inline-elements

Andarist:


docs/plugins/transform-react-inline-elements.md
```diff
@@ -10,14 +10,16 @@ Replaces the `React.createElement` function with one that is more optimized for
 
 <blockquote class="babel-callout babel-callout-info">
   <p>
-    Note that this requires the global polyfill due to React's use of ES6 <code>Symbol</code> when validating React Elements.
-    If <code>Symbol</code> is not present in the browser, your application will fail to render, as Babel will polyfill <code>Symbol</code>
-    but React will not have access to that polyfill.
+    Note that this might require the Symbol's global polyfill due to React's use of ES6 <code>Symbol</code> when validating React Elements.
```
existentialism: I think we can should add some additional context here. Maybe something like:

> Note that this requires `Symbol` to be available (or polyfilled) _globally_ since React uses them when validating elements.
>
> When using <code>transform-runtime</code>, polyfills (by default, including `Symbol`) are specifically scoped to not pollute globally, causing your application to fail when rendering as React will not have access to it.
>
> We recommend importing/requiring `babel-polyfill` in the entry point of your application and using `babel-preset-env` with the `useBuiltIns` option to only include the polyfills your targets need.

(feel free to edit/change/augment as much as you want)

Andarist: i think this
```
Note that this requires Symbol to be available (or polyfilled) globally since React uses them when validating elements.
```
is really misleading, cause this transform doesnt need it at all, it ONLY needs it if:
- we polyfill locally (`'transform-runtime'`)
- OR include external helpers via `transform-runtime` (`['transform-runtime', { helpers: true, polyfill: false }]`), because those helpers come precompiled so they import `Symbol` polyfill even with `polyfill: false`
- and possibly with `external-helpers` too, i guess they might be precompiled too, but didnt use this and hadnt time to check

If we use `transform-react-inline-elements` alone, everything works fine, cause it inlines `_jsx` helper at the top of the module which is not precompiled and has this:
```javascript
var REACT_ELEMENT_TYPE = (typeof Symbol === 'function' && Symbol.for && Symbol.for('react.element')) || 0xeac7;
```
which is not in a conflict with a version within React so they work just ok together

existentialism: Good point, we should probably just rewrite the whole section and focus on these edge cases since they'd be most helpful for users.

Andarist: maybe just remove this whole Note and add explanation of those edge cases from the top of the page and something like Caveats to the bottom of the main `Readme.md` for this plugin?

existentialism: Yeah, and to your point in the original issue, they probably belong in the package's README so that users viewing docs in npm/GitHub and the website alike can get the info. At least it will make it less confusing where the source is :)

/cc: @xtuc @hzoo

Andarist: ok, gonna ammend this PR later so it removes this note entirely and ill move it contents to the package's `README` and we'll continue from there :)

PR #1242, opened 2017-05-02
update links for regenerator-runtime/runtime

nebulousdog:
old link is being deprecated: https://github.com/facebook/regenerator/blob/master/runtime.js

closes https://github.com/babel/babel/issues/5692.


existentialism:
@nebulousdog thanks!

PR #1256, opened 2017-05-30
Minor grammar fix

rivertam:
I think?

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -9,10 +9,10 @@ redirect_from:
 
 ## Why is the output of `for...of` so verbose and ugly?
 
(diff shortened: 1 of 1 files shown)


rivertam:
Your suggestion has a comma splice... maybe "In order to comply with the specification, the iterator's `return` method must be called on errors.`"?

xtuc:
Yes, because this sentence is quite hard to understand but I'm not a native speaker.

Maybe @existentialism could help here.

existentialism:
Let's go with @rivertam's latest suggestion! 👍

PR #1262, opened 2017-06-08
Switch links to preset-env from deprecated preset-latest

ZhangYiJiang:
The `preset-es*` pages still point to `preset-latest` in a banner at the top of the page, even though it has been deprecated. This PR changes that to `preset-env`.


hzoo:
Oh cool thought we changed all of those, thanks @ZhangYiJiang!!

PR #1272, opened 2017-07-11
inline plugin to turn echo into console.log

hzoo:


scripts/php.js
```diff
@@ -263,6 +262,28 @@ window.global = window;
     try {
       transformed = babel.transform(code, {
         presets: ['php'],
+        plugins: [function() {
+          return {
+            visitor: {
+              CallExpression: function(path) {
+                if (path.node.callee.type === 'Identifier' &&
+                    path.node.callee.name === 'echo') {
```
xtuc: We can add `print` here.

`print "foo" ` generates a print function call.

PR #1273, opened 2017-07-12
7.0.0-alpha.15

hzoo:
horrible but it works


xtuc:
I would like to have https://github.com/babel/babel.github.io/pull/1187 done. We could then publish 7 doc.

PR #1274, opened 2017-07-12
Sidebar to overflow when fixed

ndelangen:
## Bug:
The sidebar contains more items then can be displayed on some pages, the sidebar cannot be scrolled.

## What I did:
I've added a bottom-position and overflow:auto to the sidebar when it's fixed.

## How does this fix the problem:
The sidebar is now scrollable.

## Demo:
[demo of the feature](http://www.youtube.com/watch?v=GvyDnd50I1U "Demo of feature")

### DIFF
--- a/_sass/components/_sidebar.scss
+++ b/_sass/components/_sidebar.scss
@@ -81,6 +81,8 @@
     &.affix {
       position: fixed; /* Undo the static from mobile first approach */
       top: 70px;
+      bottom: 0;
+      overflow: auto;
     }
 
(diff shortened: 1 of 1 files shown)


xtuc:
I just noticed some this happens:

screenshot from 2017-07-12 14-30-04

ndelangen:
Did I introduce that with this? I'll take a look..

xtuc:
Yes, I think so.

This is what is see on my screen:

babel

Issue #1275, opened 2017-07-12
Guest Blog Post: Babel-plugin-preval/babel-macros


hzoo:
I think it would be interesting to hear from the community on ideas/plugins that add new behavior/etc

Recently both of these have had some attention so it would be great to hear from @kentcdodds about what they are and why via Babel, maybe some more info on implementation?

kentcdodds:
Sure, so to do this I just create a new file [here](https://github.com/babel/babel.github.io/tree/master/_posts) and follow the format of [this one](https://github.com/babel/babel.github.io/blob/master/_posts/2017-03-01-upgrade-to-babel-7.md)? What do I name the file? Is the date the date that it's created or the date it's supposed to be published?

hzoo:
Good question haha.. I'm always confused too, just do created and we'll change it after. Yeah same format for the top part but otherwise do whatever!

hzoo:
https://github.com/babel/website/pull/1277

PR #1277, opened 2017-07-12
Create 2017-07-12-zero-config-with-babel-macros.md

kentcdodds:
Start work on #1275

Just to get things going


hzoo:
love it already

I was going to write a blog post about Babel the other day and it would just say `We need help`

existentialism:
@kentcdodds yay! I was scared you forgot about us :)

kentcdodds:
I just published 1.0.0 of `babel-macros` :)

kentcdodds:
Thanks @ericclemmons! Maybe you missed the paragraph where I list pretty much every existing macro, and then the one later where I express a hope that more libraries will implement macros... Can I make it more clear?

kentcdodds:
I'm guessing the build failures are unrelated to my changes. Please let me know what I need to update about this PR if anything :)

hzoo:
I think we should probably mention at the top or somewhere we don't intend for macros to be used for syntax transforms (the main purpose of Babel, to transform latest js into older js) but for other kinds of code transformation. Might be related to @ericclemmons' question. It's not really that much of a thing for most users - just a better ux for lib authors that are doing "some" kinds of code transformation.

kentcdodds:
I end the blog post with: 

> `babel-macros` takes a slightly different approach however. While macros have often been associated with defining new syntax for a language, that's not the goal of `babel-macros` at all. In the case of `babel-macros` it's more about code transformations.

I'm not sure of a good place to put that other than where it is... Ideas?

hzoo:
Yeah I saw, I'd like it at the top in some form if not just to say Babel is used for more things than just esnext, and Babel macros helps with that or something?

kentcdodds:
Alrighty, I [updated some of the text](https://github.com/babel/website/pull/1277/commits/9d859a5217a8702f51e8dde032d2621e44ee9b4e). I think that should work.

hzoo:
I think this is good, we can land 🙂

maybe we could have a section that explains what you wouldn't use babel-macros for though, so other community plugins/users have a better idea why/why not use babel-macro

existentialism:
@kentcdodds thanks so much!

kentcdodds:
Thank you @existentialism and @hzoo! 👏

PR #1280, opened 2017-07-22
Switch Babel 7 REPL to babel-standalone

Daniel15:
I released a build of babel-standalone that uses Babel 7, so I'm switching the REPL to use it. It's much faster to load than the `packd.now.sh` one. I also re-added the preset selector.

I'd like to add the `env` preset too... Looks like @yavorsky created a "babel-preset-env-standalone". I'd like to merge that directly into babel-standalone if possible.

Live demo: http://babeldev.dan.cx/7/
babel-standalone branch: https://github.com/babel/babel-standalone/tree/v7


yavorsky:
Wow, cool!
I think we can get rid of env-standalone and modify original repo for this purposes. Merging https://github.com/babel/babel-preset-env/pull/379 might be enought to use it with packd.now.sh. Or add build script (like in env-standalone), because packd.now.sh is really slow.
The huge benefit is that we don't need to update 2 repos every time we made changes, just running build script will be enough.

Also, I've created [2.0 branch](https://github.com/babel/babel.github.io/tree/preset-env-2.0) for https://babeljs.io/7 + 2.0 preset-env and could merge it with your work.

hzoo:
@Daniel15 is the v7 releases automated? I did a alpha.17 release the other day

Daniel15:
Once the PR is merged to add the "env" preset to Babel-standalone, we can add it here. https://github.com/babel/babel-standalone/pull/90

@hzoo It's not automated yet, but it could be! I was just worried about breaking changes messing things up :P I'll update the auto-update script to accommodate it. 

_Sent from my Samsung SAMSUNG-SM-G925A using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

hzoo:
yeah would like if the repl was always the latest and if theres a bug then we can fix, or my other idea about making a new repl link when a babel pr is merged

PR #1290, opened 2017-08-05
Limit issues displayed on contributor page

kedromelon:
Added feature mentioned in https://github.com/babel/babel.github.io/pull/1192#issuecomment-284830515

(will gladly take on the other bullet point later too!)

### DIFF
--- a/scripts/contributing.js
+++ b/scripts/contributing.js
@@ -38,10 +38,10 @@
                 labels: 'beginner-friendly'
             },
             success: function(data){
-                var limitedData = _.take(data, ISSUES_LIMIT);
+                var limitedData = data.slice(0, ISSUES_LIMIT);
 
-                _.forEach(limitedData, function(issue) {
-                    $('.open-issues-section--beginner .issues').append('<li class="issue"><a href="' + issue.html_url + '"><span class="issue__number">#' + issue.number +'</span><span class="issue__title">' + issue.title + '</span></a></li>');
+                limitedData.forEach(function(issue) {
+                    $('.open-issues-section--beginner .issues').append(buildIssueRow(issue));
                 });
             },
             error : function(xhr, status, error){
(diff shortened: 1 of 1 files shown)


kedromelon:
also -- what should the limit be? 5 seemed good to me, but perhaps that's a little low.

tkgogogo:
https://github.com/babel/babel.github.io/pull/1192#issuecomment-284830515

tkgogogo:
https://localhost:50443/#Scratch

kedromelon:
added filtering of beginner issues out from the "help wanted" issues list

also changed callbacks to filter/limit data via lodash -- i think this might be more readable? let me know if tweaks need to be made to better fit code style here though :)

scripts/contributing.js
```diff
@@ -36,9 +38,11 @@
                 labels: 'beginner-friendly'
             },
             success: function(data){
-                for(var i = 0; i < data.length; i++){
-                    $('.open-issues-section--beginner .issues').append('<li class="issue"><a href="' + data[i].html_url + '"><span class="issue__number">#' + data[i].number +'</span><span class="issue__title">' + data[i].title + '</span></a></li>');
-                }
+                var limitedData = _.take(data, ISSUES_LIMIT);
+
+                _.forEach(limitedData, function(issue) {
```
Daniel15: What's the advantage of `_.forEach` over `limitedData.forEach`?

scripts/contributing.js
```diff
@@ -36,9 +38,11 @@
                 labels: 'beginner-friendly'
             },
             success: function(data){
-                for(var i = 0; i < data.length; i++){
-                    $('.open-issues-section--beginner .issues').append('<li class="issue"><a href="' + data[i].html_url + '"><span class="issue__number">#' + data[i].number +'</span><span class="issue__title">' + data[i].title + '</span></a></li>');
-                }
+                var limitedData = _.take(data, ISSUES_LIMIT);
```
Daniel15: `data.slice(0, ISSUES_LIMIT)`

scripts/contributing.js
```diff
@@ -36,9 +38,11 @@
...
+
+                _.forEach(limitedData, function(issue) {
+                    $('.open-issues-section--beginner .issues').append('<li class="issue"><a href="' + issue.html_url + '"><span class="issue__number">#' + issue.number +'</span><span class="issue__title">' + issue.title + '</span></a></li>');
```
Daniel15: I know this is similar to how it was before, but ideally we should remove the XSS vulnerabilities from this 😛 

We have jQuery available here, so this should **not** be building HTML elements like this:
```
<span class="issue__title">' + issue.title + '</span>
```

scripts/contributing.js
kedromelon: hm, i suppose none. at least, nothing intentional. i see no reason to use lodash over it -- i'll change it

scripts/contributing.js
kedromelon: great point! i'll change that here

tkgogogo:
1dccffa

kedromelon:
gonna close this -- it's sat for a while and is trivial enough to be implemented elsewhere easily. feel free to reopen if needed though!

hzoo:
Hey Noah, would totally like to add something like this again for the new site

and if we could find a way to link a label to a page that would be cool

PR #1296, opened 2017-08-07
Revert "Remove submodules and instead download READMEs from GitHub"

hzoo:
Reverts babel/babel.github.io#1293

my bad


Daniel15:
Copying my comment from that PR: I think it's totally fine to have this merged, Github will just fail to build the site and keep the old version. Don't worry. We should be able to get James to switch the DNS today?

hzoo:
Oh ok 👍 , we aren't really updating the site anyway

Daniel15:
That's what I *suspect* will happen. Not verified. 😛

PR #1298, opened 2017-08-08
Remove redirect_from in REPL

Daniel15:
I think this is breaking the REPL on Netlify. If we do want to redirect people hitting `/repl.html`, we can configure a Netlify redirect in `_redirects` for that (which is a true server-side redirect).


Daniel15:
Confirmed this fixes it: https://deploy-preview-1298--babel.netlify.com/repl/

PR #1299, opened 2017-08-08
change readme, add bootstrap

hzoo:


Makefile
```diff
@@ -1,5 +1,11 @@
 .PHONY: build serve
 
+bootstrap:
+	bundle install
```
DanBuild: `make serve` does a more thorough check - Should we move that here?

```
	@if ! which bundle >/dev/null; then \
	echo "bundler is not installed, please install it with 'gem install bundler'."; \
	exit 1; \
	fi
 
	bundle check || bundle install;
```

Daniel15: `make serve` does a more thorough check; should we move that here?

```
	@if ! which bundle >/dev/null; then \
	echo "bundler is not installed, please install it with 'gem install bundler'."; \
	exit 1; \
	fi
 
	bundle check || bundle install;
```

hzoo: yeah feel free to change!

PR #1301, opened 2017-08-08
ignore me plz - Another Netlify PR test

Daniel15:
Just checking if the pull request status works


Daniel15:
So the comment works, but the PR status doesn't. It's supposed to show something here:


Back to Netlify support I go. 😛

Daniel15:
Never mind, I fixed it I think.

Issue #1291, opened 2017-08-06
Source map viewer


hzoo:
https://twitter.com/jasonlaster11/status/893914807090520064 via @jasonlaster

Similar to https://twitter.com/left_pad/status/885547523233984512 - we should integrate this too @boopathi

bvaughn:
So far as I can tell, this is redundant with #65. Safe to close this one?

hzoo:
Only other source map tool I remember is http://sokra.github.io/source-map-visualization/

bvaughn:
Ah~ I didn't realize this ticket was about _that_. I thought this was just saying we should add source maps to the output (like I've done [here](https://bvaughn.github.io/babel-repl/#?babili=false&browsers=&builtIns=true&code_lz=MYGwhgzhAEBiD29oG8BQ1oCMwCcAUAlCutAL6qlA&debug=true&evaluate=false&lineWrap=true&presets=react&prettier=false&showSidebar=true&targets=), although you may need to use an incognito link since it's got some caching).

PR #1315, opened 2017-08-15
REPL attaches source maps to eval()

### LINKED ISSUE #65: REPL: Include Source Maps?
Thought:  Including source maps in the transpiled REPL output would allow browser debugging through the ES6 (REPL input) vs ES5 (REPL output) code.

bvaughn:
This PR adds a couple of improvements to the new REPL:
* `eval` is now done within an `iframe` rather than a context-less `new Function` so that compiled code can access `window` without potentially clobbering the REPL itself.
* Inline source maps are passed to `eval` so errors will map to pre-compiled code. This works fine on Safari and Firefox but I can't get it working on Chrome. I'm aware of some issues with this (see [459499](https://bugs.chromium.org/p/chromium/issues/detail?id=459499#c22) and some potential workarounds (see [`EvalSourceMapDevToolModuleTemplatePlugin.js`](https://github.com/webpack/webpack/blob/3f79150d6ab2ad7a614b94e55c9314adf581b07a/lib/EvalSourceMapDevToolModuleTemplatePlugin.js#L65) but I'm not having much luck.

Potential downsides:
* The inline source maps do not work for Chrome. (This is also the case for eg the Preact REPL so maybe we're stuck with it for now. I've reached out to Jason about it on Twitter though.)
* Firefox 55 seems to cache the source maps so that they don't stay in-sync with the evaled code (see devtools-html/devtools-core/issues/606) but Firefox Nightly updates correctly (as does Safari).

Resolves #65

### DIFF
--- a/js/repl/scopedEval.js
+++ b/js/repl/scopedEval.js
@@ -1,20 +1,20 @@
 // @flow
 
-// Some libraries are loaded directly from CDN to improve page speed.
-// It's best to mask them so that evaled code doesn't accidentally interact with the page.
-// eg ReactDOM.render() shouldn't be able to unmount the REPL app.
-// Other globals are just potentially problematic to eval,
-// eg document.body.innerHTML = '' shouldn't be able to replace the Babel website.
-const globals = [
-  "React",
-  "ReactDOM",
-  "LZString",
-  "CodeMirror",
-  "document",
-  "window",
-];
+let iframe = null;
 
-export default function scopedEval(code: string) {
-  // $FlowFixMe
-  new Function(...globals, code)();
+export default function scopedEval(code: string, sourceMap: string) {
+  if (iframe === null) {
+    iframe = document.createElement("iframe");
+    iframe.style.display = "none";
+
+    document.body.append(iframe);
+  }
+
+  // Append source map footer so errors map to pre-compiled code.
+  if (sourceMap) {
+    code = `${code}\n//# sourceMappingURL=data:application/json;charset=utf-8;base64,${btoa(sourceMap)}`;
+  }
+
+  // Eval code within an iframe so that it can't eg unmount the REPL.
+  iframe.contentWindow.eval(code);
 }
--- a/js/repl/scopedEval.js
+++ b/js/repl/scopedEval.js
@@ -14,7 +14,7 @@ export default function scopedEval(code: string, sourceMap: ?string) {
   // Append source map footer so errors map to pre-compiled code.
   if (sourceMap) {
     code = `${code}\n//# sourceMappingURL=data:application/json;charset=utf-8;base64,${btoa(
-      sourceMap
+      unescape(encodeURIComponent(sourceMap))
     )}`;
   }
 
--- a/js/repl/scopedEval.js
+++ b/js/repl/scopedEval.js
@@ -2,17 +2,20 @@
 
 let iframe = null;
 
-export default function scopedEval(code: string, sourceMap: string) {
+export default function scopedEval(code: string, sourceMap: ?string) {
   if (iframe === null) {
     iframe = document.createElement("iframe");
     iframe.style.display = "none";
 
+    // $FlowFixMe
     document.body.append(iframe);
   }
 
   // Append source map footer so errors map to pre-compiled code.
   if (sourceMap) {
-    code = `${code}\n//# sourceMappingURL=data:application/json;charset=utf-8;base64,${btoa(sourceMap)}`;
+    code = `${code}\n//# sourceMappingURL=data:application/json;charset=utf-8;base64,${btoa(
+      sourceMap
+    )}`;
   }
 
   // Eval code within an iframe so that it can't eg unmount the REPL.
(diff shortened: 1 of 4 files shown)


Daniel15:
Could you provide an example of how someone would use this? I see the value when the "evaluate" option is used, but what's the use case for showing the raw source map in the UI?

bvaughn:
> Could you provide an example of how someone would use this?

I was just trying to pick up issue #65. I wasn't the advocate for this feature. 😄

sparty02:
Oh snap, I saw ``` to5.transform ``` in my original comment; blast from the past!  The original reason I opened it was in line with [Daniel's comment](https://github.com/babel/website/pull/1315#issuecomment-322969866), to allow for sourcemap support when using the "evaluate" option.

I like the idea of @bvaughn's suggestion of using a source map link, a la ``` //# sourceMappingURL=repl.js.map ``` in order to make the output less cluttered, but I would omit the sourcemap display in the UI.

Daniel15:
Based on @sparty02's comment, it sounds like we should not show the raw source map in the output pane, but instead **only** use it when evaluating code.

> Oh snap, I saw to5.transform in my original comment; blast from the past! 

lol, my fork of the Babel site was still called `6to5.github.com` until recently. 😛

bvaughn:
> Based on @sparty02's comment, it sounds like we should not show the raw source map in the output pane, but instead **only** use it when evaluating code.

That sounds reasonable. I was unclear of the use-case for this feature. Just wanted to get the ball rolling on it. 😄

loganfsmyth:
It'd be awesome to have a toggle to enable a view like https://sokra.github.io/source-map-visualization/ in the long run.

hzoo:
@loganfsmyth I think that is covered by https://github.com/babel/website/issues/1291, good ideas 👍

loganfsmyth:
Yay!

bvaughn:
Okay. Couple of improvements have been made:
* `eval` is now done within an `iframe` rather than a context-less `new Function` so that compiled code can access `window` without potentially clobbering the REPL itself.
* Inline source maps are passed to `eval` so errors will map to pre-compiled code. This works fine on Safari and Firefox but I can't get it working on Chrome. I'm aware of some issues with this (see [459499](https://bugs.chromium.org/p/chromium/issues/detail?id=459499#c22) and some potential workarounds (see [`EvalSourceMapDevToolModuleTemplatePlugin.js`](https://github.com/webpack/webpack/blob/3f79150d6ab2ad7a614b94e55c9314adf581b07a/lib/EvalSourceMapDevToolModuleTemplatePlugin.js#L65) but I'm not having much luck.

bvaughn:
Interesting! The approach I'm using in this PR works fine in a stand-alone create-react-app based project, but fails when run locally via `make serve` - so maybe one of our Webpack transforms is interfering and it will actually work fine once deployed to production?

js/repl/scopedEval.js
```diff
@@ -1,20 +1,23 @@
...
+    iframe = document.createElement("iframe");
+    iframe.style.display = "none";
+
+    // $FlowFixMe
+    document.body.append(iframe);
+  }
+
+  // Append source map footer so errors map to pre-compiled code.
+  if (sourceMap) {
+    code = `${code}\n//# sourceMappingURL=data:application/json;charset=utf-8;base64,${btoa(
```
loganfsmyth: FYI `btoa` will break with any code containing non-ASCII characters, so we may want to pull in a library to cover base64 encoding.

bvaughn: Great tip! Thanks @loganfsmyth.

I'm now calling `unescape(encodeURIComponent(sourceMap))` which should resolve this.

Can be verified here as soon as Netlify updates the preview:
https://deploy-preview-1315--babel.netlify.com/repl2#?babili=false&browsers=&builtIns=false&code_lz=MYGwhgzhAEBiD29oG8BQ1rHgOwgFwCcBXYPeAgCgEoV0NoB6B6QXg3BAPboF9VvsBTAO5xE1ANxA&debug=false&evaluate=true&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&showSidebar=true&targets=

bvaughn:
Okay. I'm happy with this PR. There's a follow-up item for Firefox (although this works in Firefox Nightly so maybe it will just work itself out) and Chrome (no idea still) but the `eval` change is nice and source maps in Safari/Firefox are nice.

js/repl/scopedEval.js
loganfsmyth: Fancy, that does appear to work.

js/repl/scopedEval.js
```diff
@@ -1,20 +1,23 @@
...
-export default function scopedEval(code: string) {
-  // $FlowFixMe
-  new Function(...globals, code)();
+export default function scopedEval(code: string, sourceMap: ?string) {
+  if (iframe === null) {
+    iframe = document.createElement("iframe");
+    iframe.style.display = "none";
+
+    // $FlowFixMe
+    document.body.append(iframe);
```
Daniel15: Does it actually need to be inserted into the body? I think just declaring the iframe and not inserting it should work fine too.

bvaughn: Yeah. contentWindow is null until it's been appended.

bvaughn:
Thanks Daniel!

Issue #1313, opened 2017-08-14
REPL should lazy-load Babel standalone from CDN


bvaughn:
Lazy-load Babel standalone from CDN as well to better support loading different versions of Babel.

We'll have to decide how to handle compatibility with presets as well as other tools (eg Prettier).

This is a follow-up item for PR #1297

bvaughn:
cc @hzoo to add this to the ["New REPL" milestone](https://github.com/babel/website/milestone/1). (I don't have permissions to.)

Daniel15:
Basically, reimplement this PR in the new REPL: https://github.com/babel/website/pull/1284

PR #1318, opened 2017-08-15
Lazy load babel on REPL

existentialism:
Fixes https://github.com/babel/website/issues/1313.

**EDIT**:

Reduced it to a simple version string output:

image


js/repl/src/ReplOptions.js
```diff
@@ -362,56 +411,65 @@ const styles = {
...
     }
   }),
   expandedContainer: css({
-    minWidth: '150px',
+    minWidth: '180px',
     display: 'flex',
-    overflow: 'auto',
-    boxSshadow:
+    flexDirection: 'column',
+    boxShadow:
```
existentialism: @bvaughn found a typo we all missed :)

js/repl/src/Repl.js
```diff
@@ -160,6 +162,27 @@ export default class Repl extends React.Component {
     );
   }
 
+  _setupBabel() {
+    loadBabel(this.state.babel, success => {
```
existentialism: We _could_ refactor this alongside loadPlugins to be a bit more generic, but figured we could do that in a follow up, if anything.

js/repl/src/ReplOptions.js
bvaughn: 🤡  It's not widely know, but the `box-sshadow` is awesome

hzoo:
I think it's fine for the invalid build number since you would be going to the link via babel-bot not manually (you shouldn't be at least), make the 99% case fast (also circle ci isn't doing so hot atm https://status.circleci.com/)

hzoo:
Would be nice to link to the corresponding PR, or github release

bvaughn:
Maybe we should hide this label _unless_ the URL flag is specified? This way we don't eat up so much space.

existentialism:
@bvaughn yeah, I'm still experimenting with other options. I think it's valuable to show the current Babel version at _somewhere_, though I agree it's currently taking up a bit more space than we'd like.

bvaughn:
I was kind of picturing an info/question icon that's either FAB-style or maybe shown only on hover.

js/repl/src/loadBabel.js
```diff
@@ -0,0 +1,50 @@
...
+    /\/version\/(.+)\/?$/
+  );
+
+  if (versionFromPath) {
+    version = versionFromPath[1];
+  }
+
+  // No specific version passed, so just download the latest release.
+  if (!version) {
+    version = 'next'; // This should be changed to "latest" when Babel 7 is stable
```
Daniel15: Maybe this should actually be `6` since this is the stable REPL.

js/repl/src/loadBuildArtifacts.js
```diff
@@ -0,0 +1,39 @@
...
+
+  xhr.onload = function() {
+    const response = JSON.parse(xhr.responseText);
+    if (response.message) {
+      alert(`Could not load Babel build #${build}: ${response.message}`);
+      return;
+    }
+
+    const artifacts = response.filter(x =>
+      /babel-standalone\/babel.js$/.test(x.path)
```
Daniel15: You have way more newlines than I did in my version... were my lines too long? 😛

existentialism: I ran prettier :)

bvaughn: Prettier ftw 😁

js/repl/src/loadBuildArtifacts.js
```diff
@@ -0,0 +1,39 @@
...
+  // first call CircleCI's API to get the URL to the artifact.
+  const buildURL = `https://circleci.com/api/v1.1/project/github/${repo}/${build}/artifacts`;
+
+  const xhr = new XMLHttpRequest();
+  xhr.open('get', buildURL, true);
+
+  xhr.onload = function() {
+    const response = JSON.parse(xhr.responseText);
+    if (response.message) {
+      alert(`Could not load Babel build #${build}: ${response.message}`);
```
Daniel15: I wanted to display these in a nicer way, but there was no good way to get the error message into the UI. Maybe we could improve that now with the new REPL?

Daniel15:
The "repo" (Daniel15/babel in your example) is going to be very very rarely used, so I wouldn't bother showing it in the UI. Essentially, it will only differ if the PR is going to a different repo (just for testing purposes). For any PRs to the "real" Babel, it'll always be `babel/babel` as that's the CircleCI project that's being built.

For CircleCI builds, we append the PR number to the version number: https://github.com/babel/babel/blob/7.0/packages/babel-standalone/src/gulpTasks.js#L19-L20. It'll look like `7.0.0-alpha.15+pr.12345` where `12345` is the PR number. Given that, we probably don't need to show the build number in the UI. Although we could have multiple builds per PR (every time the user pushes to the branch, we'll build it again) so maybe it's beneficial to show the build number. Hmm.

js/repl/src/loadBuildArtifacts.js
Daniel15: Prettier is SO NICE. I love it. We ran it across millions of lines of code at work.

existentialism:
@bvaughn @Daniel15 rebased/updated... basically reduced this to _only_ output the version number.

bvaughn:
I like the loading UI! 😄

I'm still not a fan of showing the version by default though. I think we should only show the version label if the user has specified a target version in the URL.

Daniel15:
I like a showing the version. It makes it obvious which version is being
used (for example, if there's an issue with babel-standalone and it doesn't
publish correctly, we'll see that there's an older version in use).

bvaughn:
> for example, if there's an issue with babel-standalone and it doesn't publish correctly, we'll see that there's an older version in use

Do we really need the REPL UI to detect this case? 🤔

Sidebar space is precious, especially as we expand to add more options there (eg 3rd party plugins from NPM). I can see us needing to switch to an accordion UI in the near future to make more room for things.

hzoo:
Knowing the version of babel-core is important if we want to use the repl as a way to report issues (ideally we'd have every version in some menu then I guess)

existentialism:
One thing we could do is have the debug option always available (and not just for env's debug output), and output the version info there?

bvaughn:
Sounds like I'm in the minority here then 😄 in which case, that's cool. I'll defer.

However, I still think we should at least change the visual treatment of it to be a part of the rest of the sidebar rather than being a sticky footer part.

bvaughn:
I was going to take a quick stab at tweaking the version label but on this branch, `make serve` just hangs for me. One time it failed with:
> Liquid Exception: Liquid syntax error (line 50): Variable '{{a}' was not properly terminated with regexp: /\}\}/ in js/node_modules/balanced-match/README.md

The other 3x I've tried, it just hangs. Switching back to `master` works fine though.

Wondering if anyone else has seen this?

existentialism:
@bvaughn jekyll hates you :) I have a fix for making the label part of the sidebar, will push in a sec

bvaughn:
> jekyll hates you

Yeah maybe so...but I think it's worth confirming with at least one other person before merging.

> I have a fix for making the label part of the sidebar, will push in a sec

Great 👍

Daniel15:
> I think it's worth confirming with at least one other person before merging.

Travis and Netlify are both OK with this diff... But they're robots so maybe they don't count as "other people" 😛 

Also they're building the site (`make build`) not watching. Could you try `make build` and see if that works for you?

bvaughn:
Pfft 😆 Robots!

Yes, `make build` works for me. `make serve` hangs (although it's pretty speedy on `master`).

```sh
$ make serve
bundle check || bundle install; \
	bundle exec jekyll serve
The Gemfile's dependencies are satisfied
Configuration file: /Users/bvaughn/Documents/git/babel.github.io/_config.yml
Configuration file: /Users/bvaughn/Documents/git/babel.github.io/_config.yml
            Source: /Users/bvaughn/Documents/git/babel.github.io
       Destination: /Users/bvaughn/Documents/git/babel.github.io/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
```

Daniel15:
Hmm that's very weird. `make build` and `make serve` should be near-identical except serve also watches for changes. 

I could try it on my computer but it won't be until tonight - All my Babel stuff is on my home PC rather than my work PC.

bvaughn:
Fetched and pulled and `make serve` is behaving now.

js/repl/loadBuildArtifacts.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    const artifacts = response.filter(x =>
+      /babel-standalone\/babel.js$/.test(x.path)
+    );
+
+    if (!artifacts || artifacts.length === 0) {
+      alert(
+        `Could not find valid babel-standalone artifact in build #${build}`
+      );
```
existentialism: Was originally trying to get the first version of this as a straight port of @Daniel15's original PR, but I'll go ahead bubble up the error through the callbacks and show it in the "Loading Babel" message.

existentialism:
@bvaughn thanks for the review, made the changes and now error messages bubble up to the loading screen; can follow up this PR with improved versions (tweak copy, can detect 404s on builds, probably add an error icon instead of sad face emoji, etc).

bvaughn:
> add an error icon instead of sad face emoji

Love it 😁  Happy to help with this

js/repl/PresetLoadingAnimation.js
```diff
@@ -1,8 +1,14 @@
...
-const PresetLoadingAnimation = () =>
-  <div className={styles.loadingAnimation}>
+type PresetLoadingAnimationProps = {
+  className: ?string,
+};
+
+const PresetLoadingAnimation = ({
+  className = "",
+}: PresetLoadingAnimationProps) =>
+  <div className={`${className || ""} ${styles.loadingAnimation}`}>
```
bvaughn: nit: `className || ""` is redundant with the initializer above

js/repl/Repl.js
```diff
@@ -110,18 +116,32 @@ export default class Repl extends React.Component {
...
+    this._setupBabel();
   }
 
   render() {
     const state = this.state;
 
+    if (!state.babel.isLoaded) {
+      let message = "Loading Babel...";
+
+      if (!state.babel.isLoading && state.babel.didError) {
```
bvaughn: nit: `state.babel.didError` should be sufficient?

js/repl/Repl.js
```diff
@@ -167,6 +187,19 @@ export default class Repl extends React.Component {
...
 
+  _setupBabel() {
+    loadBabel(this.state.babel, babelState => {
+      this.setState(state => ({
+        ...babelState,
+        ...(babelState.isLoaded ? this._compile(state.code, state) : null),
+      }));
+
+      if (babelState.isLoaded) {
+        this._checkForUnloadedPlugins();
```
bvaughn: Should this be done in the `setState` callback?

js/repl/loadBabel.js
```diff
@@ -0,0 +1,43 @@
...
+import { BabelState, LoadScriptCallback } from "./types";
+
+export default function loadBabel(config: BabelState, cb: LoadScriptCallback) {
+  function doLoad(url) {
+    loadScript(url, cb);
+  }
+
+  // See if a CircleCI build number was passed in the path
+  // Prod (with URL rewriting): /repl/build/12345/
+  // Dev: /repl/#?build=12345
```
Daniel15: Heh, this is my comment (this is copied across from `7.html`). I try to keep them informative. 😛

js/repl/loadBuildArtifacts.js
```diff
@@ -0,0 +1,43 @@
...
+
+      if (!artifacts || artifacts.length === 0) {
+        error = `Could not find valid babel-standalone artifact in build #${build}`;
+      } else {
+        url = artifacts[0].url;
+      }
+    }
+
+    cb(url, error);
```
Daniel15: Would be good to use promises here, but then I guess we'd need to load a polyfill (maybe?)

PR #1319, opened 2017-08-15
Fix affix sidebar flickering

imagentleman:
When fixing an overflow issue on the sidebar with this PR https://github.com/babel/website/pull/1274

A flickering issue was introduced that was pointed out in this comment by @xtuc (his last one on the thread):
https://github.com/babel/website/pull/1274#issuecomment-315534841

The issue can be fix by using a max-height instead of a bottom with value zero.

How it looks on production (https://babeljs.io/docs/core-packages/):
https://screencast-o-matic.com/watch/cbjor0ldMM

How it would look like with the fix:
https://screencast-o-matic.com/watch/cbjorTldM8

### DIFF
--- a/_sass/components/_sidebar.scss
+++ b/_sass/components/_sidebar.scss
@@ -81,7 +81,6 @@
     &.affix {
       position: fixed; /* Undo the static from mobile first approach */
       top: 70px;
-      bottom: 0;
       overflow: auto;
     }
 
--- a/scripts/docs.js
+++ b/scripts/docs.js
@@ -75,9 +75,19 @@
 
         return offsetTop - navbarOuterHeight - sidebarMargin;
       }),
-      bottom: _.memoize(function() {
-        return $('.footer').outerHeight(true);
-      })
+      bottom: _.throttle(function() {
+        return $('.footer').outerHeight(true) + $('#discourse-comments').outerHeight(true);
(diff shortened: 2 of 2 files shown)


xtuc:
Thanks for your first contribution :+1:

screenshot from 2017-08-16 09-33-55
----
screenshot from 2017-08-16 09-34-09

I believe the sidebar will be fixed with https://github.com/babel/website/issues/1042

imagentleman:
Oh great, if https://github.com/babel/website/issues/1042 is going to fix the sidebar i see no point in this PR hehe (so, you can just close it).

But, for what is worth i updated it with a little fix for the other thing that i think you are pointing out (the sidebar overlapping the discussion). We just have to include it in the bottom calculation of the affix (not just include the footer), and since the discussion loads async, we can't just memoize it, but we can throttle it but an arbitrary big number.

There are all kinds of strange things happening with that bootstrap affix sidebar, like when you get to the community discussion, it disappears sometimes if the previous sidebar item had a lot of sub-items (because they get collapsed and the sidebar height changes). I guess, all the stuff in this pr could be only a band-aid fix in the meantime.

existentialism:
@imagentleman thanks again for the PR, we landed https://github.com/babel/website/pull/1343 as a short term fix!

PR #1320, opened 2017-08-16
Integrate new REPL into website

Daniel15:
Integrates @bvaughn's new REPL into the website. The styling still needs a little tweaking, but it seems to work fine.

Changes made:
- `repl2.html` file was added! w00t
- `create-react-app` was replaced with a regular Webpack config, as this made it easier to integrate into the site
- React is now loaded from a CDN (previously, all third-party libraries **except** React were being loaded from a CDN)
- CodeMirror was changed to use `import` + a Webpack `external` rather than directly referencing `window.CodeMirror`
- REPL's `package.json` has been merged into the main `package.json`, and all unused packages (like Jest, Promise, sw-precache-webpack-plugin, url-loader, whatwg-fetch, webpack-manifest-plugin, and postcss-loader) were removed
- Files were moved from `js/repl/src` to `js/repl` to flatten the directory structure a bit (particularly now that nothing else is in the `js/repl` directory)

In dev mode, the Webpack dev server is used to serve the JS. This avoids the slowness of Jekyll's build, and also allows hot reloading. Consequently, `make serve` (or `npm run start` or `yarn start`) will now execute both the Webpack server **and** the Jekyll server. Everything is still loaded from the Jekyll server *except* for the JS, which is loaded via Webpack. If the Webpack server is not running for whatever reason, the JS file will be loaded via Jekyll instead.

In prod mode, the JS is built into the `build` directory which Jekyll then copies across to `_site`


Daniel15:
I have no clue what this build error is about:
> 10:42:30 PM: ERROR: YOUR SITE COULD NOT BE BUILT:
> 10:42:30 PM: No repo name found. Specify using PAGES_REPO_NWO environment variables, 'repository' in your configuration, or set up an 'origin' git remote pointing to your github.com repository.

I'm going to sleep, I'll take a look later.

In the meantime, you can get a screenshot as a teaser 😛

Daniel15:
Turns out the build error is because `JEKYLL_ENV=production` breaks the build 😛  I'll fix that first (#1321)

js/repl/CodeMirror.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 
 import { css } from 'glamor';
+import CodeMirror from 'codemirror'
```
bvaughn: I chose to go with CDN-based `CodeMirror` so we could share with Prettier and others.

Daniel15:
CodeMirror is still loading from the CDN. I'm using Webpack externals.

bvaughn:
Oh, nice! I hadn't noticed that.

Daniel15:
Thanks to @existentialism's change in #1322, this is successfully building now. 🎉

I'm going to merge this so we can keep iterating quickly. @bvaughn, please feel free to let me know if you don't like any of my changes or if I've done anything incorrectly 😃

bvaughn:
I haven't had the chance to do more than glance through this PR so far, but everything I saw looked great! Thanks Daniel!

repl2.html
```diff
@@ -0,0 +1,23 @@
...
+- https://unpkg.com/react-dom@16.0.0-beta.3/umd/react-dom.production.min.js
+- https://unpkg.com/codemirror@5.28.0/lib/codemirror.js
+- https://unpkg.com/codemirror@5.28.0/mode/javascript/javascript.js
+- https://unpkg.com/codemirror@5.28.0/mode/xml/xml.js
+- https://unpkg.com/codemirror@5.28.0/mode/jsx/jsx.js
+- https://unpkg.com/codemirror@5.28.0/keymap/sublime.js
+- https://unpkg.com/codemirror@5.28.0/addon/display/placeholder.js
+- https://unpkg.com/codemirror@5.28.0/addon/selection/active-line.js
+- https://unpkg.com/lz-string@1.4.4/libs/base64-string.js
+- https://unpkg.com/lz-string@1.4.4/libs/lz-string.min.js
```
Daniel15: I forgot to mention - I changed all of these to exact versions in order to get more efficient caching. With a version range, unpkg redirects to the latest version matching the range, and the redirect is only cached for 5 minutes to avoid caching an old version for a long period of time. That means that basically every time you load the REPL, it'd need to send a network request and do the redirect. With an exact version, it can be cached "forever" and loaded directly from cache, totally avoiding the network request if the files have already been cached.

Exact versions also ensures that the live site won't suddenly break (in the same way that a Yarn lockfile ensures a local build won't break)

Only exception is for babel-standalone because we do always want the very latest version.

Issue #1321, opened 2017-08-16
Remove GitHub Pages specific Rubygems


Daniel15:
We're not using "production" Jekyll builds (`JEKYLL_ENV=production`) at the moment, but I want to do so in #1320. However, they're currently broken with:
> No repo name found. Specify using PAGES_REPO_NWO environment variables, 'repository' in your configuration, or set up an 'origin' git remote pointing to your github.com repository.

If we set this correctly (eg. `PAGES_REPO_NWO=babel/website`), a bunch of URLs are broken. 😛 

This is coming from the [github-metadata](https://github.com/jekyll/github-metadata) Rubygem. This looks like it's only useful for sites hosted on GitHub Pages. We should get rid of it. It's being pulled in via the `github-pages` Gem. We should remove that Gem and instead only reference the Gems we actually care about.

PR #1322, opened 2017-08-16
Switch back to jekyll gem

existentialism:
Fixes https://github.com/babel/website/issues/1321


Gemfile
```diff
@@ -1,5 +1,10 @@
 source 'https://rubygems.org'
-gem 'github-pages', group: :jekyll_plugins
+gem 'jekyll'
```
Daniel15: Ideally these versions should be locked, at least to a minor version:
```
gem 'jekyll''~>3.5.2'
```

so that the site doesn't completely fall apart if someone decides to update the Rubygems after a new major release of anything comes out.

but this is OK too. Thanks!

Gemfile
```diff
@@ -1,5 +1,10 @@
...
+gem 'jekyll-paginate'
+gem 'jekyll-redirect-from'
```
Daniel15: Now that we're using Netlify, I'd like to kill this and replace them with proper server-side redirects (in `_redirects` file)

existentialism: Definitely, I'll follow up with some cleanup, and pin the gem versions.

PR #1323, opened 2017-08-16
Pin versions in gemfile and drop redirect_from

existentialism:


_redirects
```diff
@@ -6,3 +6,25 @@ layout: null
 /repl/build/:build /7/ 200
 # Previous versions of the REPL
 /repl/version/:version /7/ 200
+
+# Ported from redirect_from
+
+/docs/using-babel/                  /docs/setup/
```
Daniel15: I was going to say "make these 301 (permanent) redirects", but Netlify uses 301 by default: https://www.netlify.com/docs/redirects/#http-status-codes. Nice.

_redirects
```diff
@@ -6,3 +6,25 @@ layout: null
...
+/browser.html                       /docs/setup/
+/docs/core-packages/babel-register/ /docs/usage/babel-register/
+/docs/usage/require/                /docs/usage/babel-register/
+/docs/usage/options/                /docs/usage/api/
+/docs/advanced/plugins/             /docs/plugins/
+/docs/usage/modules/                /docs/plugins/
+/caveats.html                       /docs/usage/caveats/
+/usage.html                         /docs/usage/cli/
+/docs/faq/                          /faq/
+/php.html                           /php/
```
Daniel15: Maybe add `/repl.html` too, We removed that as it was causing issues with the site build, but the redirect might still be useful.

PR #4, opened 2015-01-06
New website

thejameskyle:
screen shot 2015-01-05 at 11 34 51 pm
screen shot 2015-01-05 at 11 35 50 pm
screen shot 2015-01-05 at 11 36 07 pm
screen shot 2015-01-05 at 11 37 16 pm

PR #1326, opened 2017-08-17
Tweaked REPL styles to accommodate for surrounding Babel site CSS

bvaughn:
Admittedly some of this is a bit hacky, but I didn't want to edit the surrounding styles much (at least not yet 😁 )

### DIFF
--- a/_sass/pages/_repl.scss
+++ b/_sass/pages/_repl.scss
@@ -194,6 +194,8 @@
 
 .babel-repl-v2 {
   height: 100%;
+  margin-bottom: $padding-large-vertical * -4; // TODO (bvaughn) Offset margin-top in _footer.scss
+  -webkit-font-smoothing: antialiased;
 
   * {
     box-sizing: content-box;
--- a/js/repl/ReplOptions.js
+++ b/js/repl/ReplOptions.js
@@ -125,6 +125,7 @@ class ExpandedContainer extends Component {
 
             <input
               checked={envConfig.isEnvPresetEnabled}
+              className={styles.envPresetCheckbox}
               type="checkbox"
               onChange={this._onEnvPresetEnabledChange}
             />
@@ -403,24 +404,24 @@ const styles = {
     color: colors.inverseForegroundLight,
 
     [media.large]: {
-      height: '4rem',
-      width: '2rem',
-      top: 'calc(50% - 2rem)',
-      borderTopRightRadius: '4rem',
-      borderBottomRightRadius: '4rem'
+      height: '5rem',
+      width: '3rem',
+      top: 'calc(50% - 3rem)',
+      borderTopRightRadius: '5rem',
+      borderBottomRightRadius: '5rem'
     },
 
     [media.mediumAndDown]: {
-      height: '2rem',
-      width: '4rem',
-      left: 'calc(50% - 2rem)',
-      borderBottomLeftRadius: '4rem',
-      borderBottomRightRadius: '4rem'
+      height: '3rem',
+      width: '5rem',
+      left: 'calc(50% - 3rem)',
+      borderBottomLeftRadius: '5rem',
+      borderBottomRightRadius: '5rem'
     }
   }),
   closeButtonIcon: css({
-    width: '1.5rem',
-    height: '1.5rem',
+    width: '2rem',
+    height: '2rem',
 
     [media.mediumAndDown]: {
       transform: 'rotate(90deg)'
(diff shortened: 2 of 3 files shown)


js/repl/ReplOptions.js
```diff
@@ -450,7 +450,7 @@ const styles = {
     }
   }),
   inputCheckboxLeft: css({
-    marginRight: '0.5rem'
+    margin: '0 0.5rem 0 0 !important' // Clashes with default input[type="checkbox"] styles in main.css
```
Daniel15: Is it possible to make the selector more specific instead of using `!important`?

_sass/pages/_repl.scss
```diff
@@ -194,6 +194,8 @@
 
 .babel-repl-v2 {
   height: 100%;
+  margin-bottom: $padding-large-vertical * -4; // Compensate for _footer.scss margin-top
```
Daniel15: Why `-4`? Is there a constant for that somewhere that you can reuse?

The current REPL has the footer hanging off the bottom of the page: https://babeljs.io/repl/. I think we should just keep that same style.

js/repl/ReplOptions.js
bvaughn: I don't think more-specific would be any less hacky in this case, since nested selectors are slow and kind of gross. The problem is that we set (gross) global styles for checkboxes and radio. I think _that's_ where we should fix this TBH, but I'm new here and didn't want to make too many splashes.

_sass/bootstrap/_scaffolding.scss
```diff
@@ -20,7 +20,7 @@
 // Body reset
 
 html {
-  font-size: 10px;
+  font-size: 13px;
```
Daniel15: Is this going to affect the font size on every other page?

_sass/pages/_repl.scss
bvaughn: I thought the comment explained this, but `_footer.scss` defines:
```css
margin-top: $padding-large-vertical * 4;
```

This offsets that. The top margin doesn't look right in between the footer and the REPL.

_sass/bootstrap/_scaffolding.scss
bvaughn: No. We set an overriding `<body>` style in `main.css`. This only affects rem unites (which I used in REPL).

js/repl/ReplOptions.js
Daniel15: > The problem is that we set (gross) global styles for checkboxes and radio.

Isn't that relatively common though, particularly for sites with custom-styled checkboxes (I'm not sure if Babel site has that)?

> I don't think more-specific would be any less hacky in this case,

I think it's less hacky than `!important`... Your intent is to use more specific styling for this particular checkbox, and that's what nested selectors are for in CSS 😛 

I guess either way is fine.

_sass/pages/_repl.scss
Daniel15: @bvaughn Could we set a variable `$footer-margin-top` and reuse it across both?

Daniel15:
lgtm, feel free to merge.

Daniel15:
Actually hold up, the font size in the mini REPL on the home page is too big:


vs


See the "Put in next-gen JavaScript" and "Get browser-compatible JavaScript out" titles.

bvaughn:
I'll be transparent about this diff: I don't really like _any_ of these changes. Unfortunately the Babel site sets a lot of global styles and I don't have a lot of good choices without making larger impact than I feel comfortable making _at the moment_. (This is something I wouldn't mind working on in the future though...)

At this point, I'm hesitant to change styles that impact the rest of the site because I don't know the history of the site, why things are the way they are, if we have any long-term plans for the UI, etc.

bvaughn:
Interesting. Good catch. I guess our Ace styles use REMs. I'll take another pass at this..

Daniel15:
> At this point, I'm hesitant to change styles that impact the rest of the site because I don't know the history of the site, why things are the way they are, if we have any long-term plans for the UI, etc.

You and me both! Maybe someone that's worked on Babel for a long time has more context here. It looks like the original PR is https://github.com/babel/website/pull/4 but there's no discussion in there at all. Haha.

Daniel15:
> Interesting. Good catch. I guess our Ace styles use REMs. I'll take another pass at this..

I looked through many pages and that seemed to be the only thing that was different. Feel free to hack that one particular use case.

bvaughn:
I felt kind of dirty about this. I took a second pass and updated my REMs instead, so I can just leave the `<html>` font size like it was before. We can clean the global styles up at some point in the future. 😄

hzoo:
> because I don't know the history of the site, why things are the way they are, if we have any long-term plans for the UI, etc.

If there were long term plans they are definetely not in progress atm 😄. Feel free to change anything, if we can remove the global styles even better.

bvaughn:
Cool. I'd be happy to take a pass at it and maybe clean it up at some point. 😄

PR #1332, opened 2017-08-17
Prepared a precommit hook, flow + prettier

### LINKED ISSUE #1310: Run Flow and Prettier via precommit hook
Run `yarn flow` and `yarn prettier` as part of precommit hook (or a CI check).

Currently, Prettier is setup to run against `js/repl` only but it would be nice if it ran against all of the Babel JS.

This is a follow-up item for PR #1297

Andarist:
fixes #1310

There is one problem though
```
Error: js/repl/index.js:10
 10:   module.hot.accept();
       ^^^^^^^^^^^^^^^^^^^ call of method `accept`. Method cannot be called on
 10:   module.hot.accept();
       ^^^^^^^^^^ property `hot` of unknown type
```

and honestly I aint how to handle this. The hacky way that I have figured out is this
```js
class module {
  static hot: {
    accept: Function,
  }
}
```
but its obviously hacky and ugly. Any ideas?

cc @bvaughn

### DIFF
--- a/js/repl/CodeMirror.js
+++ b/js/repl/CodeMirror.js
@@ -1,19 +1,19 @@
 // @flow
 
-import { css } from 'glamor';
-import CodeMirror from 'codemirror'
-import React from 'react';
-import { colors } from './styles';
+import { css } from "glamor";
+import CodeMirror from "codemirror";
+import React from "react";
+import { colors } from "./styles";
 
 const DEFAULT_CODE_MIRROR_OPTIONS = {
   autoCloseBrackets: true,
-  keyMap: 'sublime',
+  keyMap: "sublime",
   lineNumbers: true,
   matchBrackets: true,
-  mode: 'text/jsx',
+  mode: "text/jsx",
   showCursorWhenSelecting: true,
   styleActiveLine: true,
-  tabWidth: 2
+  tabWidth: 2,
 };
 
 type Props = {
@@ -22,19 +22,20 @@ type Props = {
   options: Object,
   placeholder?: string,
   value: ?string,
-  preserveScrollPosition: boolean
+  preserveScrollPosition: boolean,
 };
 
 export default class ReactCodeMirror extends React.Component {
   static defaultProps = {
     autoFocus: false,
     preserveScrollPosition: false,
-    onChange: (value: string) => {}
+    // eslint-disable-next-line no-unused-vars
+    onChange: (value: string) => {},
   };
 
   props: Props;
   state = {
-    isFocused: false
+    isFocused: false,
   };
 
   _codeMirror: any;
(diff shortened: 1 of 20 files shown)


Daniel15:
`module.hot.accept` is part of Webpack's dev server, which is why Flow doesn't know about it. You could add a `// $FlowFixMe` comment, or a [library definition](https://flow.org/en/docs/libdefs/) (which I assume is what you were trying).

Daniel15:
The main Babel repo uses "Husky" to run Prettier as a Git commit hook. Could we do the same thing in this repo? Here's the commit that added it to Babel: https://github.com/babel/babel/commit/93cc22dae1744fd919c8b49909457e49c9b34e82

The Babel repo doesn't have a `.lintstagedrc` file. Instead, it has the config in `package.json`. It's probably a good idea to keep the two repos consistent with each other.

Andarist:
> module.hot.accept is part of Webpack's dev server, which is why Flow doesn't know about it. You could add a // $FlowFixMe comment, or a library definition (which I assume is what you were trying).

I guess `// $FlowFixMe` is better, thanks for the tip.

> The main Babel repo uses "Husky" to run ESLint + Prettier as a Git commit hook. Could we do the same thing in this repo? Here's the PR that added it to Babel: babel/babel#5412

Added eslint to the precommit hook + configured eslint with prettier plugin.

> The Babel repo doesn't have a .lintstagedrc file. Instead, it has the config in package.json. It's probably a good idea to keep the two repos consistent with each other.

Personally I prefer separate files, putting configurations of various tools into `package.json` bloats it quite a lot and also gets published to npm for no reason - but ofc I'll adjust my PR for the parity with babel repo.

Daniel15:
> Personally I prefer separate files, putting configurations of various tools into package.json bloats it quite a lot and also gets published to npm for no reason - but ofc I'll adjust my PR for the parity with babel repo.

👍 

If we want to change it, we should change it for both repo at the same time.

Andarist:
Ive amended my commit with configuration tweaks + run prettier and fixed some linting errors/warnings with the next commit so its should be easier to code review them.

js/repl/CodeMirror.js
```diff
@@ -1,19 +1,19 @@
...
-import { colors } from './styles';
+import { css } from "glamor";
+import CodeMirror from "codemirror";
+import React from "react";
+import { colors } from "./styles";
 
 const DEFAULT_CODE_MIRROR_OPTIONS = {
   autoCloseBrackets: true,
-  keyMap: 'sublime',
+  keyMap: "sublime",
```
Daniel15: Why do we use `"` instead of `'`? :'(

bvaughn: We should use whatever Prettier decides 😄

Daniel15: At FB we use prettier with single quotes.

Andarist: I was suggested with the fact that the main repo uses double quotes

PR #1335, opened 2017-08-18
Mask global variables before eval-ing REPL code

### LINKED ISSUE #1329: Sanitize evaluated input
Copying one of the examples from the [React "getting started" guide](https://facebook.github.io/react/docs/introducing-jsx.html) and enabling "evaluate" actually un-mounts the REPL and replaces it with the example code. ([Here is an example](http://babeljs.io/repl2/#?babili=false&browsers=&builtIns=false&code_lz=GYVwdgxgLglg9mABMOAnAtgQygOU-gUwAoQBnA1ASkQG8AoRRVAqEVJMigOmBlVNz4CiANSIA5BNGJOqLgBtMAvIQDcdAL506EBAJnlUiALy0GyPsqEAuCQAlMqAA4VxAGnOKrhW-IAKFAQAXuKa6jp6UIgE8gSEYFGmROYAPAAWAIwAfOaMdjHycG60KBjYKsSylBoAhKkA9Jk5lOEASgSY0AAiAPIAslzMYAAmFMmMMXEECR6Mw3AQIPFQXADmLACiscsAQgCeAJLDROKocHBQ4pR0LUA&debug=false&evaluate=true&lineWrap=false&presets=es2016%2Creact&prettier=false&showSidebar=true&targets=).)

bvaughn:
Maybe we should consider evaling code within a web worker to be safer, but this PR is at least a step in a better direction. We now mask globals to prevent evaled code from accidentally interacting with the Babel website or the REPL.

After this change the following code snippets will no longer mess up the REPL:
```js
document.body.innerHTML = ''
```

```jsx
ReactDOM.render(
  <div/>,
  document.getElementById('root')
);
```

Potentially resolves #1329

### DIFF
--- a/js/repl/scopedEval.js
+++ b/js/repl/scopedEval.js
@@ -1,0 +1,19 @@
+// @flow
+
+// Some libraries are loaded directly from CDN to improve page speed.
+// It's best to mask them so that evaled code doesn't accidentally interact with the page.
+// eg ReactDOM.render() shouldn't be able to unmount the REPL app.
+// Other globals are just potentially problematic to eval,
+// eg document.body.innerHTML = '' shouldn't be able to replace the Babel website.
+const globals = [
+  'React',
+  'ReactDOM',
+  'LZString',
+  'CodeMirror',
+  'document',
+  'window',
+];
+
+export default function scopedEval(code: string) {
+  new Function(...globals, code)();
+}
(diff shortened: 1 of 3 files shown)


js/repl/scopedEval.js
```diff
@@ -0,0 +1,19 @@
...
+  'React',
+  'ReactDOM',
+  'LZString',
+  'CodeMirror',
+  'document',
+  'window',
+];
+
+export default function scopedEval(code: string) {
+  new Function(...globals, code)();
```
Daniel15: I wonder if this will break people's expectations... It's likely fine though.

bvaughn:
That's an interesting suggestion.

I'll follow up with the iframe approach. I think I'll merge this for now and iterate though!

Issue #1324, opened 2017-08-16
"Try it out" menubar item should be selected when on new REPL


Daniel15:
https://babeljs.io/repl/ -> "Try it out" is selected.
https://babeljs.io/repl-old/ -> "Try it out" is not selected.
https://babeljs.io/php -> "Try it out" is not selected.

xtuc:
The selection of the correct item in the menu is [here](https://github.com/babel/website/blob/master/_includes/header.html#L43). The `page.url contains item.url` adds so much restrictions, it's not the first time we have this kind of issues.

hzoo:
Closed by https://github.com/babel/website/pull/1603!

PR #1339, opened 2017-08-19
"Try it out" menubar item should be selected when on REPL 2

chasestarr:
Fixes issue here: https://github.com/babel/website/issues/1324

An interesting thing I learned about Liquid's `contains` operator is that it can match based on an array. The PR here adds a `match` field to each navigation item. This does seem to duplicate information (which is unexciting), but since the "try it out" menu item is intended to be overloaded I think it's a reasonable solution.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -20,7 +20,7 @@
       <ul class="nav navbar-nav">
         {% for item in site.data.navigation %}
           {% if item.subnav %}
-            <li class="dropdown {% if page.url contains item.url %}active{% endif %}">
+            <li class="dropdown {% if page.url contains item.match %}active{% endif %}">
               <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-expanded="false">{{ item.title }} <span class="caret"></span></a>
               <ul class="dropdown-menu" role="menu">
                 {% for group in item.subnav %}
@@ -31,7 +31,7 @@
 
                   {% for item in group.items %}
 
-                    <li {% if page.url contains item.url %}class="active"{% endif %}>
+                    <li {% if item.match contains page.url %}class="active"{% endif %}>
                       <a href="{{ item.url | prepend: site.baseurl }}">{{ item.title }}</a>
                     </li>
                   {% endfor %}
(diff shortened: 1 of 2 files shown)


_includes/header.html
```diff
@@ -20,7 +20,7 @@
       <ul class="nav navbar-nav">
         {% for item in site.data.navigation %}
           {% if item.subnav %}
-            <li class="dropdown {% if page.url contains item.url %}active{% endif %}">
+            <li class="dropdown {% if page.url contains item.match %}active{% endif %}">
```
Daniel15: What if we make `match` optional and fall back to `url` when it's not provided. Something like this (untested, might need some tweaking):

```
{% assign match = item.match %}
{% if match == blank %}
  {% assign match = item.url %}
{% endif %}
...
<li class="dropdown {% if page.url contains match %}active{% endif %}">
```

chasestarr: Hey - good idea. Just pushed a change 👍

Daniel15:
Checked a few pages and it *almost* seems to work fine.

These all work:
https://deploy-preview-1339--babel.netlify.com/learn-es2015/
https://deploy-preview-1339--babel.netlify.com/docs/editors
https://deploy-preview-1339--babel.netlify.com/docs/usage/babel-register/
https://deploy-preview-1339--babel.netlify.com/repl/
https://deploy-preview-1339--babel.netlify.com/blog/
https://deploy-preview-1339--babel.netlify.com/faq/
https://deploy-preview-1339--babel.netlify.com/repl2/ <-- Awesome, it is fixed
https://deploy-preview-1339--babel.netlify.com/php/

Blog posts don't, though: 
https://deploy-preview-1339--babel.netlify.com/blog/2017/08/16/contributing-to-babel-three-lessons-to-remember

"Blog" is not highlighed. Compare to the live site, where it is: 
https://babeljs.io/blog/2017/08/16/contributing-to-babel-three-lessons-to-remember

I think this is because it's no longer matching substrings. The old code `page.url contains item.url` would match if the page URL is `/blog/foo/bar` (as `/blog/` is contained within it), but the new code `match contains page.url` does not, as the order is reversed (and `/blog/foo/bar` is of course not contained within `/blog/`)

hzoo:
This works now on the new site with a recent docusarus change http://new.babeljs.io/repl, thanks!

PR #1340, opened 2017-08-20
Custom loading animation/page for REPL

### LINKED ISSUE #1325: Nicer loading state for new REPL
While the JS is loading, the new REPL just shows some plain "Loading" text at the top left. Let's add a slightly nicer loading state - At least bigger text and centered, or something like that. I'm not a designer 😛

bvaughn:
Resolves #1325

screen shot 2017-08-19 at 10 49 50 pm

screen shot 2017-08-19 at 10 50 05 pm


repl2.html
```diff
@@ -19,5 +19,77 @@
...
+      <div class="repl-loading-animation-box repl-loading-animation-box-2"></div>
+      <div class="repl-loading-animation-box repl-loading-animation-box-3"></div>
+      <div class="repl-loading-animation-box repl-loading-animation-box-4"></div>
+      <div class="repl-loading-animation-box repl-loading-animation-box-5"></div>
+    </div>
+  </div>
+</div>
+
+<style type="text/css">
+  @keyframes repl-loading-animation {
```
Daniel15: Put this in `_sass/pages/_repl.scss`? We've already loaded the CSS by this point, as it's in the `<head>`. May as well put this in there 😃

Daniel15:
I think this also means that you can remove all the styling from `js/repl/PresetLoadingAnimation.js` and just use the CSS classes, right?

bvaughn:
Oh yeah, _sass/pages/_repl.scss ... Duh lol

I'll be less stupid in the morning 😁 will fix it up then

bvaughn:
> I think this also means that you can remove all the styling from `js/repl/PresetLoadingAnimation.js` and just use the CSS classes, right?

We _could_ but I'd prefer to keep the 2 separate. Sharing _within_ Glamor's styles would be okay since it would require explicit imports but I don't dig the idea of a hidden dep between the Jekyll/SASS styles and the Glamor ones.

Daniel15:
Tested on https://deploy-preview-1340--babel.netlify.com/repl2/, looks great!

PR #1341, opened 2017-08-20
Debounce compilation and plug-in changes

### LINKED ISSUE #1316: Should the REPL debounce changes?
The current REPL debounces changes and only runs the transformation after 500ms. The new one transforms the file immediately. I imagine the debouncing is useful when you're working with very large code blocks.

The new REPL should debounce all text inputs, including:
* The main code input
* Env preset textarea
* Env preset Node and Electronic number inputs

Note that this will be a little more involved than is ideal due to the way the `compile` function is currently (synchronously) hooked into `Repl.setState`. Debouncing input has a negative impact on controlled input components, so we'll need to refactor `Repl` to separate the compiled output (code, source maps, errors, etc) from the controlled inputs.

bvaughn:
Since we used controlled inputs, `state` changes for their values must remain synchronous. The actual compilation can be debounced though. I think this PR seems to do the tricky? 😅  I initially expected it to be more complicated than this.

Resolves #1316

Note that I'm moving discussion of moving compilation to a worker to #1342. I think it's worth separating from debounce.

### DIFF
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -1,6 +1,7 @@
 // @flow
 
 import { css } from "glamor";
+import debounce from "lodash.debounce";
 import React from "react";
 import CodeMirrorPanel from "./CodeMirrorPanel";
 import ReplOptions from "./ReplOptions";
@@ -51,6 +52,8 @@ type State = {
   runtimePolyfillState: PluginState,
   sourceMap: ?string,
 };
+
+const DEBOUNCE_DELAY = 500;
 
 export default class Repl extends React.Component {
   static defaultProps = {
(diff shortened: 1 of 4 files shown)


js/repl/Repl.js
```diff
@@ -292,6 +304,10 @@ export default class Repl extends React.Component {
     };
   };
 
+  _compileToState = (code: string) => {
```
Daniel15: Since we're already using this fancy property initializer syntax, may as well just call debounce here to avoid the extra code in the constructor:
```
_compileToState = debounce((code: string) => {
  this.setState(state => this._compile(code, state), this._persistState);
}, DEBOUNCE_DELAY);
```

PR #1345, opened 2017-08-24
Added Spec option to sidebar options list

devdevil666:
Added Spec option to left sidebar of the editor.


hzoo:
Cool, if we do something like this we should add loose and then remove es2015-loose

devdevil666:
What s your opinion about the realisation `Spec` / `Loose` options?

js/repl/ReplOptions.js
```diff
@@ -79,6 +83,18 @@ class ExpandedContainer extends Component {
     const disableEnvSettings =
       !envPresetState.isLoaded || !envConfig.isEnvPresetEnabled;
 
+    const es2015Enabled =
+      presetState["babel-preset-es2015"].isEnabled &&
+      !envConfig.isEnvPresetEnabled;
```
yavorsky: es2015 can be enabled, but with env option it will ignore options also. Let's rename constant to smth like `showOption`.

Also, we have `presetsSupportOptions ` with presets list that has options, but we are ignoring it there and in case of adding new items there, it's hard to determine what changes we need to provide also to add an other option for the other preset. We are too sticky to the case when only es2015 can have options and the case where spec and loose are the only 2 possible options. We need to be more scalable, so maybe it's better to keep config like:

```js
const supportedOptions = {
  es2015: ['spec', 'loose'],
}
```
so we could easily add new presets/options in the future.

PR #1346, opened 2017-08-25
Move Babel to web-worker

### LINKED ISSUE #1342: Investigate moving compilation to web-worker
Blocked by #1318; related to PR #880.

~~Possibly look into using a similar pattern as [`js-worker-search`](https://github.com/bvaughn/js-worker-search) for [falling back to the UI thread for browsers that don't support web workers](https://github.com/bvaughn/js-worker-search/blob/master/src/SearchApi.js#L26-L41).~~ Informal decision about discussing in Slack is to drop support for IE < 11 provided we have clear user-facing messaging about it.

#### Pros
* Compilation wouldn't cause the UI to feel slow/janky.

#### Considerations
* Evaluated code would not have access to `window` or `document` if it was evaluated in the worker. We would need to pass it back to the main thread.
* This will complicate the logic for JIT loading plugins and presets, as each will need to be (re)loaded in the worker context as well.

bvaughn:
Resolves #1342

## Perf Impact
Here's a super-unscientific example of processing a [large UMD file](https://unpkg.com/react-virtualized@9.9.0/dist/umd/react-virtualized.js) while scrolling around the page.
#### Before
screen shot 2017-08-26 at 9 51 48 am
#### After
screen shot 2017-08-26 at 9 51 55 am

### DIFF
--- a/js/repl/Worker.js
+++ b/js/repl/Worker.js
@@ -1,9 +1,8 @@
-// @flow
+// TODO Add comments about how we can't import code here.
 
-declare function importScripts(url: string): void;
-
-import compile from "./compile";
-import registerPromiseWorker from "promise-worker/register";
+importScripts(
+  "https://npmcdn.com/promise-worker/dist/promise-worker.register.js"
+);
 
 // This script should be executed within a web-worker.
 // Values returned below will be automatically wrapped in Promises.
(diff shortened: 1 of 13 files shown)


bvaughn:
Seeing a bit of weirdness with the env plug-in in a web-worker. When I pass it options, it errors with:

> Options {...} passed to a preset which does not accept options.

But if I try to add the preset without options, it errors with:

> Unknown option: foreign.availablePlugins. Check out http://babeljs.io/docs/usage/options/ for more information about options.
> 
> A common cause of this error is the presence of a configuration options object without the corresponding preset name. Example:
> 
> Invalid:
>   `{ presets: [{option: value}] }`
> Valid:
>   `{ presets: [['presetName', {option: value}]] }`

vjeux:
@azz moved prettier to a worker on the playground and it worked really well btw :)

bvaughn:
Nice! 😄

I'd love to do it for Flow too, after this lands. That's the most painful REPL in terms of locking the UI.

js/repl/Worker.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+
+import compile from "./compile";
+import registerPromiseWorker from "promise-worker/register";
+
+declare var Babel: any;
```
Daniel15: Do webpack `externals` not work in a web worker? `require('babel-standalone')` should be the same as referencing the global `Babel`.

js/repl/loadBabel.js
```diff
@@ -39,13 +48,12 @@ export default function loadBabel(
     return;
   }
 
+  let version = config.version;
```
Daniel15: I think the earlier position of this code was better, since the comment explains what this variable is for (`config.version` is passed in dev, which is mentioned in the comment)

Daniel15:
Nice work!

js/repl/Worker.js
bvaughn: Hm. Babel (or `babel-standalone`) isn't one of the things we alias though, is it?

js/repl/loadBabel.js
bvaughn: Didn't actually intend to move it. Have just been iterating on this code.

js/repl/Worker.js
```diff
@@ -0,0 +1,44 @@
...
+// This script should be executed within a web-worker.
+// Values returned below will be automatically wrapped in Promises.
+registerPromiseWorker(message => {
+  const { method } = message;
+
+  switch (method) {
+    case "compile":
+      return compile(message.code, message.config);
+
+    case "getBabelVersion":
```
azz: Ooh, I should totally steal this idea rather than building a [custom file](https://github.com/prettier/prettier/blob/master/website/static/lib/prettier-version.js)

js/repl/Worker.js
Daniel15: Oh good catch, we don't alias it because it's lazy loaded. Sorry for the confusion @bvaughn  :D

bvaughn:
Thanks @xtuc!

Wrapped up the last thing this morning. This PR should be good to go!

I tested Chrome, Firefox, Safari, Edge, Brave on Android, and iOS Safari.

bvaughn:
Thanks all for the feedback and reviews!

I made a couple of small tweaks and tidied up. As soon as CI finishes, I'm going to merge this. 😄

PR #1348, opened 2017-08-27
Use new React REPL

bvaughn:
For your consideration, this PR:
* Moves `/repl` to `/repl-old` and replaces it with what was previously `repl2`. (In other words, this makes the new REPL the one that's opened by default.)
* Adds a fallback link in the sidebar to the old repl (see screenshots below).
* Fixes pulldown menu layout to scroll properly on iOS+Safari.
* Removes the page footer because it took up space, made the REPL layout feel a bit awkward, and didn't add much value. (I realize this is a potentially controversial change I am proposing and I'm happy to discuss and back down if others have concerns.)

Note that I didn't (yet) rename any of the `babel-repl` CSS class names since those seem to be used in a lot of places outside of the two REPL files.

## Classic link

screen shot 2017-08-27 at 1 37 15 pm

screen shot 2017-08-27 at 1 37 27 pm

## Fixed iOS layout

screen shot 2017-08-27 at 1 54 13 pm
screen shot 2017-08-27 at 1 54 18 pm


existentialism:
LGTM, just two questions:

First, the REPL previews for PRs get the version the PR number attached:

image

I'm in favor of just dropping the "classic" repo altogether and am confident we can fix whatever bugs/regressions exist. If do wind up keeping the classic version, however, we may want to ensure the version number overflows w/ ellipsis or something.

Second, thoughts on removing the footer for the REPL only?

bvaughn:
> we may want to ensure the version number overflows w/ ellipsis or something.

Currently, a longer string like this would just cause the side menu to grow wider. But now that you mention it, it probably would be reasonable to impose a max width and overflow to "..." just in case. 😄 Done!

> Second, thoughts on removing the footer for the REPL only?

Great question! I'd considered it, but kind of thought it nice to just remove altogether b'c it doesn't really work the best. For example, the "Blog" and "FAQ" pages have the following footer-related bug:
screen shot 2017-08-27 at 3 55 34 pm

That being said, I'll re-add it to all but the (new) REPL page for now by way of a new frontmatter field. That seems less controversial I guess.

Daniel15:
Preview links (like `/repl/build/1234/`) still go to the old Babel 7 REPL... Will you update that too?

bvaughn:
Interesting. I didn't expect that, given that the new REPL is at the exact URL the old one was at.

Is `_redirects` the correct place to fix this?

Daniel15:
Yeah. It goes to /7.html which is likely still the old one.

bvaughn:
Gotcha! Thought so but wanted to confirm b'c that redirect fails for me locally (both dev and prod builds). 😄

Daniel15:
It's a Netlify feature so won't work locally. Jekyll has no support for
rewrites (actually no static site generator does,  since it needs server
side support). :(

Need to test it using the Netlify preview.

PR #1353, opened 2017-08-28
Fixed "Try" links to REPL

### LINKED ISSUE #1349: Try links doesn't work with the new REPL
When clicking on the "Try" link in http://babeljs.io/learn-es2015/#ecmascript-2015-features-destructuring the REPL opens without the content of the code snippets.

Does the content in URL conflicts with my cached content somehow? I think it should override but still keep the cached code in the local storage.

@bvaughn any idea?

Edit:
it doesn't work in incognito mode, it's not cache related.

bvaughn:
The underlying issue was that "Try" links don't use lz compression. I wrote the REPL expecting that the `code` search param would always be compressed as `code_lz`. I've updated the REPL now to accept `code` too- and to defer to `code_lz` when it's present.

Resolves #1349

### DIFF
--- a/js/repl/UriUtils.js
+++ b/js/repl/UriUtils.js
@@ -61,6 +61,7 @@ const parseQuery = () => {
       "browsers",
       "build",
       "builtIns",
+      "code",
       "debug",
       "circleciRepo",
       "evaluate",


hzoo:
oh good point

bvaughn:
Thanks for the speedy review @hzoo. You rock!

PR #1354, opened 2017-08-28
Redirect preview links to the new REPL

bvaughn:
Follow-up from [this comment](https://github.com/babel/website/pull/1348#issuecomment-325497275).

~~Going to wait and try this out on Netlify to see if it works~~ Seems to work okay on Netlify 😁

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -3,9 +3,9 @@ layout: null
 ---
 
 # REPL builds, to preview pull requests
-/repl/build/:build /7/ 200
+/repl/build/:build /repl/ 200
 # Previous versions of the REPL
-/repl/version/:version /7/ 200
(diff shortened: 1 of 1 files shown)


Daniel15:
Tested with a recent PR and it seems to work well! https://babeljs.io/repl/build/4713/. Tested a version link, and that works well too: https://babeljs.io/repl/version/6.18.1/

Only thing I noticed is for the PR links is that probably the most important part of the version number (the pull request number) is truncated:


I guess that will be fixed when we remove the "classic REPL" link.

bvaughn:
Yup. The truncation should go away after the classic link is removed. In
the meantime, hover over it for a tooltip or double click to select the
full version text.

PR #1360, opened 2017-08-30
Add flow and typescript presets to repl

bmax:
Add https://github.com/babel/babel/pull/6168 to website

### DIFF
--- a/js/repl/PluginConfig.js
+++ b/js/repl/PluginConfig.js
@@ -71,6 +71,11 @@ const presetPluginConfigs: Array<PluginConfig> = [
     isPreLoaded: true,
   },
   {
+    label: "flow",
+    package: "babel-preset-flow",
+    isPreLoaded: true,
+  },
+  {
     label: "stage-0",
     package: "babel-preset-stage-0",
     isPreLoaded: true,
@@ -88,6 +93,11 @@ const presetPluginConfigs: Array<PluginConfig> = [
   {
     label: "stage-3",
     package: "babel-preset-stage-3",
+    isPreLoaded: true,
+  },
+  {
+    label: "typescript",
+    package: "babel-preset-typescript",
     isPreLoaded: true,
   },
 ];
--- a/scripts/repl.js
+++ b/scripts/repl.js
@@ -8,11 +8,13 @@
     'es2015-loose',
     'es2016',
     'es2017',
+    'flow',
     'react',
     'stage-0',
     'stage-1',
     'stage-2',
-    'stage-3'
+    'stage-3',
+    'typescript'
   ];
 
   var targets = [


existentialism:
@bmax Thanks!

Couple thoughts:

- Should we disable typescript if flow is enabled and vice versa since they can't both be?

- Right now the [REPL defaults to 6](https://github.com/babel/website/blob/master/js/repl/loadBabel.js#L5) and selecting either flow/typescript results in an "Invalid preset" error. Maybe we can put a min version in  `presetPluginConfigs` so that flow/typescript only appear with babel-standalone `>=7.0.0-alpha.20`?

hzoo:
@existentialism the parser should already error if you turn on both so it just depends on if it's worth putting that in the UI to disable it (no?)

existentialism:
Yeah, I know... can defer, it's just better UX to me if we prevent users from doing things that lead to errors.

Daniel15:
> Maybe we can put a min version in presetPluginConfigs so that flow/typescript only appear with babel-standalone >=7.0.0-alpha.20?

Instead of adding a version restraint, let's filter them out when not available (if `Babel.availablePresets.flow` or `Babel.availablePresets.typescript` are null). As we used to say back in the older days, "feature detection > user-agent detection" 😛

existentialism:
@Daniel15 👍 good point!

hzoo:
Where can we make that check?

Daniel15:
Likely whatever is using [presetPluginConfigs in PluginConfig.js](https://github.com/babel/website/blob/47b27dede1b5573b0aab5bbc8c95991293988ad8/js/repl/PluginConfig.js) should check if the plugin is available. @bvaughn should have some ideas where it should be added, or if the logic already exists.

bvaughn:
> @bvaughn should have some ideas where it should be added, or if the logic already exists.

We could add a new attribute to the config, I guess, that specifies if a preset is mutually exclusive with another preset. This could still be "broken" by setting both values on in the URL, although I guess that would be unlikely. I don't really believe this is necessary though, or worth the complexity. I don't think users are going to try to turn on both Flow _and_ TypeScript, and if they did- I think whatever error would be shown to them in the UI would be self-explanatory.

By the way, this looks broken? 😄 

[With "flow" preset enabled"](https://deploy-preview-1360--babel.netlify.com/repl/#?babili=false&browsers=&build=&builtIns=false&circleciRepo=&code_lz=C4TwDgpgBAZg9nAXFAzsATgSwHYHMoC8UARAIYBGAxsQNxA&debug=false&evaluate=false&lineWrap=true&presets=flow&prettier=false&showSidebar=true&targets=&version=6.26.0): 
> Invalid preset specified in Babel options: "flow"

[With "typescript" preset enabled](https://deploy-preview-1360--babel.netlify.com/repl/#?babili=false&browsers=&build=&builtIns=false&circleciRepo=&code_lz=C4TwDgpgBAZg9nAXFAzsATgSwHYHMoC8UARAIYBGAxsQNxA&debug=false&evaluate=false&lineWrap=true&presets=typescript&prettier=false&showSidebar=true&targets=&version=6.26.0):
> Invalid preset specified in Babel options: "typescript"

PR #1362, opened 2017-08-31
Redirect temp babel 7 repl url to new repl

existentialism:


hzoo:
This will be v6 but i'm ok with that I guess?

existentialism:
@hzoo good point, we can do `/repl/build/7.0/` for now to point to latest build (until `version/7` works)?

hzoo:
yeah I would do that

PR #1364, opened 2017-09-01
Load runtime polyfill into scoped-eval iframe

### LINKED ISSUE #1350: New REPL doesn't include regenerator runtime
While the new REPL does include [`transform-regenerator`](https://github.com/babel/website/blob/master/js/repl/compile.js#L76), it doesn't use the regenerator runtime to be able to run async functions and generators:

```js
(async () => {})()
```

outputs the following:

```js
"use strict";

function _asyncToGenerator(fn) { return function () { var gen = fn.apply(this, arguments); return new Promise(function (resolve, reject) { function step(key, arg) { try { var info = gen[key](arg); var value = info.value; } catch (error) { reject(error); return; } if (info.done) { resolve(value); } else { return Promise.resolve(value).then(function (value) { step("next", value); }, function (err) { step("throw", err); }); } } return step("next"); }); }; }

_asyncToGenerator( /*#__PURE__*/regeneratorRuntime.mark(function _callee() {
  return regeneratorRuntime.wrap(function _callee$(_context) {
    while (1) {
      switch (_context.prev = _context.next) {
        case 0:
        case "end":
          return _context.stop();
      }
    }
  }, _callee, undefined);
}))();
```

which throws:

```
regeneratorRuntime is not defined
```

This works fine in the old REPL.

bvaughn:
This is done so that helper methods like regeneratorRuntime are accessible as globals on the iframe's `contentWindow`.

While adding this feature, I discovered that Firefox behaves differently than Chrome and Safari when a new iframe is appended- it (asynchronously) loads "about:blank" and clobbers any (synchronously) appended script tags. So we also check the `iframe.contentDocument.readyState` before appending.

This has been tested on the following:

Browser | OS
:---|:---
Chrome | OSX
Firefox | OSX
Safari | OSX
Brave | Android
Safari | iOS

Resolves #1350

### DIFF
--- a/js/repl/loadPlugin.js
+++ b/js/repl/loadPlugin.js
@@ -8,7 +8,7 @@ import type { LoadScriptCallback, PluginState } from "./types";
 export default function loadPlugin(
   state: PluginState,
   callback: LoadScriptCallback,
-  targetDocument: Document = document
+  target: ?HTMLIFrameElement
 ) {
   if (state.isLoading) {
     return;
@@ -33,6 +33,6 @@ export default function loadPlugin(
 
       callback(success);
     },
-    targetDocument
+    target
   );
 }
--- a/js/repl/loadPlugin.js
+++ b/js/repl/loadPlugin.js
@@ -7,7 +7,8 @@ import type { LoadScriptCallback, PluginState } from "./types";
 
 export default function loadPlugin(
   state: PluginState,
-  callback: LoadScriptCallback
+  callback: LoadScriptCallback,
+  targetDocument: Document = document
 ) {
   if (state.isLoading) {
     return;
@@ -19,16 +20,19 @@ export default function loadPlugin(
   const base = config.baseUrl || "https://bundle.run";
   const url = `${base}/${config.package}@${config.version || ""}`;
 
-  loadScript(url, success => {
-    if (success) {
-      state.isLoaded = true;
-      state.isLoading = false;
-      state.plugin = window[camelCase(state.config.package)];
-    } else {
-      state.didError = true;
-      state.isLoading = false;
-    }
+  loadScript(
+    url,
+    success => {
+      if (success) {
+        state.isLoaded = true;
+        state.isLoading = false;
+      } else {
+        state.didError = true;
+        state.isLoading = false;
+      }
 
-    callback(success);
-  });
+      callback(success);
+    },
+    targetDocument
+  );
 }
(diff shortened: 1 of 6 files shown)


js/repl/loadPlugin.js
```diff
@@ -19,16 +20,19 @@ export default function loadPlugin(
   const base = config.baseUrl || "https://bundle.run";
   const url = `${base}/${config.package}@${config.version || ""}`;
 
-  loadScript(url, success => {
-    if (success) {
-      state.isLoaded = true;
-      state.isLoading = false;
-      state.plugin = window[camelCase(state.config.package)];
```
bvaughn: I thought we might use this but we never did so I removed it in this PR for simplicity.

bvaughn:
Try harder, Netlify! 😠

Daniel15:
Do we actually need the entire polyfill? It's very big. Could we get away with just loading `regeneratorRuntime` instead? I'd prefer only loading regeneratorRuntime + loading only the polyfills the current browser needs, via a service like https://polyfill.io/

js/repl/loadScript.js
```diff
@@ -2,8 +2,36 @@
...
+) {
+  if (target != null) {
+    const targetDocument = target.contentDocument;
+
+    // If we're loading this script into an iframe, wait for it to load,
+    // Otherwise the script tag we insert will be removed on-load.
+    if (targetDocument.readyState !== "complete") {
+      target.addEventListener("load", () => {
+        // Use the newly-loaded document.
+        // $FlowFixMe We know target isn't null at this point
```
Daniel15: Might be able to pull it into a `const` variable to avoid the FlowFixMe:

```
export default function loadScript(
  url: string,
  callback: LoadScriptCallback,
  maybeTarget: ?HTMLIFrameElement
) {
  const target = maybeTarget;
  if (target != null) {
    // ...
```

bvaughn:
> Do we actually need the entire polyfill? It's very big and modern browsers (like the very latest Chrome and Firefox) actually need very little of the stuff in the polyfill. Could we get away with just loading regeneratorRuntime instead? I'd prefer only loading regeneratorRuntime + loading only the polyfills the current browser needs, via a service like polyfill.io

That sounds like a nice follow-up! This PR can be focused on the iframe issue though, if that's ok?

js/repl/loadScript.js
bvaughn: Cool. I'll change to use a local const then. 😄

Daniel15:
> This PR can be focused on the iframe issue though, if that's ok?

Sure. Sorry, I have a bad habit of accidentally increasing the scope of issues.

bvaughn:
It's all good. You have good suggestions. 😄

PR #1367, opened 2017-09-04
Added top level error boundary to repl fixes #1351

### LINKED ISSUE #1351: Add error boundaries [in-progress by @kaushlakers]
Since the REPL is already using React 16, we should take advantage of [the new `componentDidCatch` error handling lifecycle](https://facebook.github.io/react/blog/2017/07/26/error-handling-in-react-16.html).

As we move forward on issues like #858 and #1292, it would be nice to add `componentDidCatch` to a couple of key components. The REPL currently uses `try`/`catch` statements where errors are likely to occur (eg [`compile`](https://github.com/babel/website/blob/master/js/repl/compile.js#L71-L108), [`eval`](https://github.com/babel/website/blob/master/js/repl/WorkerApi.js#L43-L47)) but error boundaries would be a nice backup for any cases that slip through the cracks.

kaushlakers:


kaushlakers:
Yeah the failing build is puzzling. From the log [here](https://travis-ci.org/babel/website/jobs/271565761) looks like it's not able to get the directory listing. Is there so sort of client_id/secret setup I'm missing here?

bvaughn:
Yeah. I'm not sure.

Let's give it a day or so and see if someone else who's been here longer than I have has ideas. Otherwise this PR looks good. Thank you!

bvaughn:
I'm confident the CI failure is unrelated to this PR so I'm going to merge it. 👍

PR #1368, opened 2017-09-04
REPL sidebar persisted state tweaks

### LINKED ISSUE #1358: Sidebar accordion follow-up
Follow-up work for PR #1355.

WRT [@existentialism's comment](https://github.com/babel/website/pull/1355#issuecomment-325676886): Should we get rid of the sidebar toggle? Is it adding value to be able to hide the whole sidebar? This is questionable in the large layout view but I see the value in the smaller layouts. A lot of the available height on a phone is taken up even when all accordions are collapsed.

If we decide to keep the sidebar toggle, then we should consider lifting the accordion state up a level so that it's persisted between sidebar toggles. This isn't a _huge_ deal (since I don't expect people to toggle it off and back on very often) but it would be a nice touch if it could be done in a way as to not add much complexity.

bvaughn:
* a7d97fd: Don't store certain types of user preferences in the URL. (Some state is better kept in `localStorage`, as it's somewhat arbitrary.) Only store the bits of information important to share via URL in the URL.
* ebae836: Which accordion tabs are opened/closed persist between sessions (using `localStorage`).

Resolves #1358


bvaughn:
Tagging a few people that _might_ have opinions about this change. I'll give you a bit to review the PR. If you're too busy- no worries. I'm also happy to just give it another round of testing and merge as is.

js/repl/AccordionTab.js
```diff
@@ -8,49 +8,31 @@ import { colors, media } from "./styles";
 type Props = {
   children: React$Element<any>,
   className: ?any,
-  defaultExpanded?: boolean,
+  isExpanded?: boolean,
   label: string,
+  toggleIsExpanded: Function,
```
Daniel15: Very minor nit: Call it `onExpandedToggle` or something like that? At least in my team at Facebook, we prefix all callbacks with `on`. Not sure if that's common or not. If it's not common in open source, feel free to ignore this comment.

js/repl/AccordionTab.js
```diff
@@ -8,49 +8,31 @@ import { colors, media } from "./styles";
...
         </div>
-        {expanded &&
-          <div className={styles.Content}>
-            {this.props.children}
-          </div>}
+        <Svg
+          className={`${styles.Arrow} ${props.isExpanded
+            ? styles.ArrowExpanded
+            : ""}`}
+          path="M15.41,16.58L10.83,12L15.41,7.41L14,6L8,12L14,18L15.41,16.58Z"
```
Daniel15: Nit: This is up to you but I usually like splitting SVG paths onto separate lines for readability:
```
path="
  M15.41,16.58
  L10.83,12
  L15.41,7.41
  L14,6
  L8,12
  L14,18
  L15.41,16.58
  Z"
```

js/repl/ReplOptions.js
```diff
@@ -299,6 +315,27 @@ class ExpandedContainer extends Component {
...
   };
+
+  _toggleIsEnvPresetTabExpanded = () => {
+    this.props.onTabExpandedChange(
+      "isEnvPresetTabExpanded",
+      !this.props.isEnvPresetTabExpanded
+    );
+  };
+
+  _toggleIsPresetsTabExpanded = () => {
```
Daniel15: Maybe just call it `_togglePresetsTab`?

js/repl/AccordionTab.js
bvaughn: I think this is a fairly common convention, and one I normally follow too- but only for callbacks that receive event-like arguments. In this case, since the method is actually just a _toggle_ (eg it doesn't accept any parameters) I thought the "on" prefix would be misleading.

js/repl/AccordionTab.js
bvaughn: I'm a little disappointed that I can't just point to Prettier and say, "We do what it does" but it seems to honor user line breaks for string attributes so...fah. I really don't care. I'll change it. 😄

js/repl/ReplOptions.js
bvaughn: Sure.

bvaughn:
Thanks for the review buddy!

PR #1373, opened 2017-09-11
remove social

hzoo:
I don't think we need this anymore?


existentialism:
Can also drop the css:

https://github.com/babel/website/blob/master/_sass/pages/_blog.scss#L58

hzoo:
Feel free to remove that and land it then?

PR #1390, opened 2017-09-26
Migrate REPL css-in-js library from glamor to emotion.

tkh44:
This is a fairly straightforward conversion so I thought I'd give it a shot. emotion uses Babel to optimize for performance and developer experience so I thought it would be fitting to use it on the Babel repl.

repl.js size

master - **105kB**

<details>
     
</details>

---

emotion - **89kB**

<details>
     
</details>

---

Bonus: I wanted to see how source maps behaved when using a glamor like api. Only works in development.
babel-source-maps
Pretty awesome!

Thanks for taking a look!

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,16 @@
 {
+  "env": {
+    "production": {
+      "plugins":[
+        "emotion"
+      ]
+    },
+    "development": {
+      "plugins": [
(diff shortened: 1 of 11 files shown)


existentialism:
@tkh44 awesome! 💯

PR #1394, opened 2017-10-02
Distinguish between no and default presets in REPL

### LINKED ISSUE #1370: Unable to share repl url with every preset disabled
http://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&circleciRepo=&code_lz=Q&debug=false&evaluate=false&lineWrap=false&presets=&prettier=false&showSidebar=true&targets=&version=6.26.0

`http://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&circleciRepo=&code_lz=Q&debug=false&evaluate=false&lineWrap=false&`**`presets=`**`&prettier=false&showSidebar=true&targets=&version=6.26.0`

babel-repl-empty-presets

As you can see there are no presets specified, so the repl picks the default (es2015, react and stage-2).
I think that the default presets should be only used when there isn't the `presets=` query parameter, not when it is empty.

existentialism:
Fixes #1370.

### DIFF
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -81,12 +81,15 @@ class Repl extends React.Component {
       prettier: persistedState.prettier,
     };
 
-    const defaultPresets = (persistedState.presets || "es2015,react,stage-2")
-      .split(",")
-      .reduce((reduced, key) => {
-        reduced[`babel-preset-${key}`] = true;
-        return reduced;
-      }, {});
+    const presets =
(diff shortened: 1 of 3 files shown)


hzoo:
Do we still want the default to be stage-2? it should probably just be env (can be another pr)

existentialism:
@hzoo yeah I'll do a followup that defaults to env

PR #1395, opened 2017-10-02
Remove prettier from REPL

### LINKED ISSUE #1388: Babel repl - prettier is not defined
Original issue submitted by @BorisTB in https://github.com/babel/babel/issues/6301

Bug

On [babel repl](https://babeljs.io/repl/) when I check Prettify, the whole repl is lagged with console error `prettier is not defined` and I can't refresh repl. (I have to remove `replState` from localStorage and then refresh)

existentialism:
Fixes #1388.

Maybe revisit with a more lightweight prettier-as-a-service call?

### DIFF
--- a/js/repl/PluginConfig.js
+++ b/js/repl/PluginConfig.js
@@ -37,11 +37,6 @@ const pluginConfigs: Array<PluginConfig> = [
     package: "babili-standalone", // TODO Switch to babel-minify-standalone
     version: "0",
   },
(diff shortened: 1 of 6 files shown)


existentialism:
@hzoo it gets loaded via https://github.com/babel/website/blob/5f6c38f76bde21a019b07336c9e666ea75caa4f5/js/repl/Repl.js#L211

PR #1398, opened 2017-10-03
Remove classic REPL

existentialism:


hzoo:
I was saying we should do this now since it's been a while since the release (although we still have some bugs), and people are confused by the link at the bottom since it's next to the version

js/repl/ReplOptions.js
```diff
@@ -46,12 +46,15 @@ type Props = {
...
-      ? <ExpandedContainer {...props} />
-      : <CollapsedContainer {...props} />}
-  </div>;
+    {props.isExpanded ? (
+      <ExpandedContainer {...props} />
+    ) : (
+      <CollapsedContainer {...props} />
+    )}
+  </div>
+);
```
bvaughn: Are these formatting changes all from Prettier? 😄

js/repl/ReplOptions.js
```diff
@@ -243,20 +248,11 @@ class ExpandedContainer extends Component {
...
-              path={`
-                M19,3H5C3.89,3 3,3.89 3,5V9H5V5H19V19H5V15H3V19A2,2 0 0,0 5,21H19A2,2 0 0,
-                0 21,19V5C21,3.89 20.1,3 19,3M10.08,15.58L11.5,17L16.5,12L11.5,7L10.08,
-                8.41L12.67,11H3V13H12.67L10.08,15.58Z`}
-            />
-            classic repl
-          </a>
-          {babelVersion &&
+          {babelVersion && (
             <div className={styles.babelVersion} title={`v${babelVersion}`}>
```
bvaughn: nit: `styles.babelVersion` and `styles.versionAndClassicRow` divs can be collapsed now since the flex-row is no longer needed.

js/repl/ReplOptions.js
existentialism: Yeah, looks like it bumped. I'll revert and push to a different PR.

js/repl/ReplOptions.js
existentialism: 👍

js/repl/ReplOptions.js
bvaughn: No worries. I don't mind being pragmatic about it, so long as we also bump the Prettier version in `yarn.lock` so we aren't undoing each others changes on PRs.

PR #1400, opened 2017-10-03
Bump prettier

existentialism:


package.json
```diff
@@ -68,7 +70,6 @@
   },
   "lint-staged": {
     "js/**/*.js": [
-      "prettier --trailing-comma es5 --write",
```
existentialism: Not needed since we use `eslint-plugin-prettier`

PR #1407, opened 2017-10-05
Make area non-user-selectable

bogas04:
Double clicking the chevron cause text to be selected. I figured there's no good reason for this to be selectable anyway.

screen shot 2017-10-05 at 6 43 44 pm


existentialism:
@bogas04 thanks!

PR #1411, opened 2017-10-07
generate package in presetPluginConfigs (fixes #1365)

### LINKED ISSUE #1365: Remove redundancy in presetPluginConfigs
In `js/repl/PluginConfig.js`

```
const presetPluginConfigs: Array<PluginConfig> = [
  {
    label: "es2015",
    package: "babel-preset-es2015",
    isPreLoaded: true,
  },
  {
    label: "es2015-loose",
    package: "babel-preset-es2015-loose",
    isPreLoaded: true,
  },
  {
    label: "es2016",
    package: "babel-preset-es2016",
    isPreLoaded: true,
  },
```

We don't need the `package` key; we can just assume the package name is `babel-preset-${label}`

VictorHom:
The issue presented in #1365 asks to delete the package key from `presetPluginConfigs`

@Daniel15 
I'm not sure if that was the exact task to be done or if there is something more here. Checking `js/repl/ReplOptions.js`, it looks like the presetPluginConfig needs the package key in order for http://babeljs.io/repl/# page to handle its state.

Currently I just deleted the hardcoded package key and generated after the object is created, so that the `package` key is still used in `ReplOptions.js`

Happy to help on the issue in any way I can :)

### DIFF
--- a/js/repl/PluginConfig.js
+++ b/js/repl/PluginConfig.js
@@ -78,10 +78,6 @@ const presetPluginConfigs: Array<PluginConfig> = [
   },
 ];
 
-presetPluginConfigs.forEach(config => {
-  config.package = `babel-preset-${config.label}`;
-});
-
 export {
   envPresetConfig,
   envPresetDefaults,
--- a/js/repl/PluginConfig.js
+++ b/js/repl/PluginConfig.js
@@ -42,50 +42,45 @@ const pluginConfigs: Array<PluginConfig> = [
 const presetPluginConfigs: Array<PluginConfig> = [
   {
     label: "es2015",
-    package: "babel-preset-es2015",
     isPreLoaded: true,
   },
   {
     label: "es2015-loose",
-    package: "babel-preset-es2015-loose",
     isPreLoaded: true,
   },
   {
     label: "es2016",
-    package: "babel-preset-es2016",
     isPreLoaded: true,
   },
   {
     label: "es2017",
-    package: "babel-preset-es2017",
     isPreLoaded: true,
   },
   {
     label: "react",
-    package: "babel-preset-react",
     isPreLoaded: true,
   },
   {
     label: "stage-0",
-    package: "babel-preset-stage-0",
     isPreLoaded: true,
   },
   {
     label: "stage-1",
-    package: "babel-preset-stage-1",
     isPreLoaded: true,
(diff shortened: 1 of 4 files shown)


Daniel15:
This is close. You don't need to set config.package at all. Instead, modify the code that uses it to instead use config.label.

VictorHom:
@Daniel15 another question - was the intention to remove all `package` key from other configs in `PluginConfig.js` or just from `presetPluginConfigs`?

right now, I have it replUtils to still check the `package` key because `pluginConfig` is still using `package` key. (https://github.com/babel/website/blob/master/js/repl/ReplOptions.js#L118)

VictorHom:
@Daniel15 - just wanted to bring this back up and see what your thoughts are on this. thanks :)

Daniel15:
Sorry for the delay, I've been super busy recently. This looks good to me. @bvaughn  what do you think?

bvaughn:
I dunno. This looks like a clean-up task you filed, @Daniel15 ? 😄

That being said, if we are just assuming the plugin is "babel-preset-" + label, then I would expect to see us concatenating those strings somewhere here rather than just doing `package || label`.

Also I think this change will break the serialized/local-storage settings for the REPL since it would cause a mismatch.

hzoo:
I think the hardcoding is fine, and actually I guess this would break when we use scoped packages like `@babel/core` or something?

bvaughn:
I guess I didn't see the original code as problematic so I'm not sure what the motivation is. 😄

VictorHom:
I don't have a strong take on this. If this isn't the right fix, I can close. 

Was just trying to get a feel for the repo through working on an issue 😄

Daniel15:
>  I'm not sure what the motivation is. 

To remove duplication :) 

I tested it and old URLs still seem to work fine, so let's do it.

bvaughn:
>> I'm not sure what the motivation is.

> To remove duplication :)

At the cost of added complexity.

That isn't an obvious win to me.

hzoo:
^ this won't be the same next release since we moved to scoped like I since earlier https://github.com/babel/babel/pull/6495 😄

Daniel15:
@hzoo - The plugins are all in babel-standalone, so changing the package name likely won't affect anything

PR #1415, opened 2017-10-10
Add babel-preset-env 2.0 support. Controll version/build from the circle.

yavorsky:
Since we've moved babel-preset-env to the original repo and [going to move babel-preset-env-standalone](https://github.com/babel/babel/pull/6438), we can take the same advantage like we're used with babel-standalone.

This PR:
1. Add `envVersion` query param for babel-preset-env version, so /#?envVersion=2.0.0-beta.2 will use it for the repl.
2. Allows testing preset-env changes after each PR by using build version to get artifacts from the Circle CI along with babel-standalone.
3. Displays current version near the preset:
    screen shot 2017-10-10 at 13 04 42
4. Fixes env preset checkbox state when initializing. Change text color inside textarea.
5. Add the most features from the latest preset-env:
    - useBuiltIns: `usage`/`entry`
    - forceAllTransforms
    screen shot 2017-10-10 at 12 24 45
6. All preset options are links to the documentation:
    ezgif com-video-to-gif 1

🚧  I've disabled `debug` option unless [onPresetBuild support](https://github.com/babel/babel-preset-env/pull/404) will be merged. I'll move PR to the monorepo today.

The next steps:
- Enable debug
- Add stage-3 option (babel-plugin-transform-async-generator-functions is large, so we need to lazy load it and register, but before we registered preset-env :/).
- Since preset-env is 160kb, maybe load it right off the bat ?


yavorsky:
Seems like netlify preview breaks down with new commits after the first successful build.

existentialism:
Let's set these dropdowns to normal font-weight:

image

js/repl/PluginConfig.js
```diff
@@ -2,10 +2,21 @@
...
+  version: "6",
+  baseUrl: "https://unpkg.com",
+  instanceName: "Babel",
+};
+
 const envPresetConfig: PluginConfig = {
   label: "Env Preset",
   package: "babel-preset-env-standalone",
-  version: "0",
+  version: "2.0.0-beta.2",
```
existentialism: Should we default this to 1.x since we default to Babel 6.x?

yavorsky:
@existentialism Updated:
Now, babel-preset-env 1.6.2 is by default.
Also, we have different options according to the version:

**For ^2.0:**
screen shot 2017-10-13 at 12 59 33

**For older versions:**
screen shot 2017-10-13 at 12 55 09

yavorsky:
Changed status to WIP, because we need to land https://github.com/babel/babel/pull/6438 firstly and then fit it there.

yavorsky:
Seems like it's ready.

It works with babel-preset-env 2.0 betas and with circleci builds. So after each PR into babel-preset-env we could test preset's changes on the REPL also!
Tested with https://github.com/babel/babel/pull/6438 (5831 build).

The only thing is that I need to test `@babel/preset-env-standalone` fetching when it will be `npm publish`ed.

cc @hzoo  @existentialism

js/repl/loadBundle.js
```diff
@@ -70,23 +79,26 @@ export default async function loadBabel(
...
+    version = config.version;
   }
 
-  const babelStandalone =
+  const base = config.baseUrl;
+
+  const packageName =
     (semver.valid(version) && semver.gte(version, "7.0.0-beta.4")) ||
     version >= 7
-      ? "@babel/standalone"
```
hzoo: my thoughts: I kinda feel like it's easier to read/maintain with hardcoding this stuff? If we rename or anything like that I'm not sure I'd know where to look in this codebase with so much indirection? And in this case we can't substitute babel-standalone env in the same way because it won't be published until beta.33 not beta.4

yavorsky:
Ok, I've updated the PR with `shippedProposals` option and now if it's checked:

1. We are checking through the worker whether or not a plugin from the shipped proposals was registered (some of them are part of babel-standalone).
2. For unregistered plugins, we are using awesome bundle.run to load them and register. It takes the same version babel has.
3. We have implemented registerPlugins worker method and now can easily add custom plugins the same way, so it's, at least, a good start for this huge task.
4. The only problem we have is to match preset-env version with appropriate babel version, because preset-env *beta.34* doesn't work with babel *beta.5*.
5. Just noticed that it's good to add loader for shipped proposals option...

[DEMO with async-generator-functions](https://deploy-preview-1415--babel.netlify.com/repl/#?babili=false&browsers=chrome%2055&build=&builtIns=usage&code_lz=IYZwngdgxgBAZgV2gFwJYHsICobAOZwAUAlDAN4BQMuA7sKsjAIwDcVMYqApgDYAmMAExsAvkA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cenv&targets=&version=7.0.0-beta.32&envVersion=7.0.0-beta.32)

xtuc:
I'm not sure why but each time I refresh the page the Babel version decrements.
-  screenshot from 2017-12-14 08-36-28
- screenshot from 2017-12-14 08-36-21
...

And then bundle.run fails to find the version.

yavorsky:
>I'm not sure why but each time I refresh the page the Babel version decrements.
>And then bundle.run fails to find the version.

Yes, it's related to https://github.com/babel/babel/issues/7016 . Seems like we are creating bundle before `package.json` was updated.

Daniel15:
If I tick "shipped proposals", some enormous files are loaded: `plugin-proposal-async-generator-functions` loads an extra 1.6 MB of JavaScript, for example. This should likely be fixed before merging this diff, as I doubt it really needs to load 1.6 MB of JS 😄 

This is the URL I'm using to test: https://deploy-preview-1415--babel.netlify.com/repl/version/7.0.0-beta.35#?envVersion=7.0.0-beta.32

yavorsky:
> plugin-proposal-async-generator-functions loads an extra 1.6 MB of JavaScript, for example. This should likely be fixed before merging this diff, as I doubt it really needs to load 1.6 MB of JS

Yeah, it's 1.4mb and 291kb gzipped. But we are loading shipped proposals only when `Shipped Proposals` is checked.
The most of bundle's size is `@babel/core` stuff, because of we are bundling it with `bundle.run`, but other standalone plugins are using the core's stuff from the already loaded `Babel`. So maybe it's worth to add plugin-proposal-async-generator-functions to babel-standalone?
Btw, the rest stage-3 plugins is a part of bundle-standalone, so we just need async-generator-functions.

Daniel15:
>  So maybe it's worth to add plugin-proposal-async-generator-functions to babel-standalone?

That's a good idea. All standard Babel plugins should be bundled with babel-standalone :)

yavorsky:
Good! I'll create a PR.

But the most of this PR changes (loading plugins with `bundle.run`) could be useful for feature to add external plugins.

js/repl/PluginConfig.js
```diff
@@ -1,11 +1,57 @@
 // @flow
+import camelCase from "lodash.camelcase";
+import type { PluginConfig, MultiPackagesConfig, EnvFeatures } from "./types";
 
-import type { PluginConfig } from "./types";
+const convertPluginNameToInsanceName = pluginName =>
+  `_babel_${camelCase(`plugin-${pluginName}`)}`;
```
xtuc: I assumed that you meant `instance` here but it doesn't look very clear to me.
What about `normalize`, `toBabelPluginName`, `getFullyQualifiedPluginName`?

I can see that we can save a few CPU cycles (and thus probably trees :earth_africa: ) here, `plugin-` is being camelCase'd every time even if it's a constant.

yavorsky: > I assumed that you meant instance here but it doesn't look very clear to me.
What about normalize, toBabelPluginName, getFullyQualifiedPluginName

Sure! 👍

yavorsky:
Since [async-generator-function was added to the standalone package](https://github.com/babel/babel/pull/7073), we actually don't need to load `shippedProposals` from bundle.run.
But these PR changes could be useful for adding external plugins, so I'll move some of latest commits to the new brunch with external plugins support and open a new PR subsequently and revert them for current PR.

xtuc:
@yavorsky yes, bundle.run seems like a good way to support external plugins/presets.

Daniel15:
@xtuc - Does `bundle.run` support externals yet? It's a great idea to use it for external plugins/presets, but we wouldn't want to bundle a whole other Babel with those plugins/presets :)

yavorsky:
@Daniel15 Sure! You can [check how this pr handles async-generator-functions](https://github.com/babel/website/pull/1415/files#diff-70f5c2fe2441a48519c15305524aef72R26), so we could handle other plugins the same way.

For example, the `babel-plugin-jest-hoist` bundle from the `bundle.run` doesn't include babel-core: https://bundle.run/babel-plugin-jest-hoist@22.0.3

js/repl/PluginConfig.js
yavorsky: > I can see that we can save a few CPU cycles (and thus probably trees 🌍 ) here, plugin- is being camelCase'd every time even if it's a constant.

The other way is `_babel_plugin${capitalize(camelCase(pluginName))}`.
I don't think it is more performant and need an additional dependency.

xtuc:
> Does bundle.run support externals yet?

We actually discussed about that a few months ago (https://github.com/Rich-Harris/packd/issues/27) but I forgot about it.

Your point is still valid, I agree. I will work on it when I get a chance.

yavorsky:
I hope this is ready 🙄
screen shot 2018-01-05 at 15 32 14

Would be cool if someone will help to test it better

xtuc:
Sorry I haven't followed this PR so much, what is the status of babel-standalone having the incorrect version of Babel?

yavorsky:
@xtuc We can [fix only publishing process](https://github.com/babel/babel/pull/7039), but it won't solve the problem at all because all versions published before this fix to npm will still return the previous version.

I've created PR for the website, which disables version overriding:
https://github.com/babel/website/pull/1488
There are some questions need to be resolved before we can merge this. For example, should we override non-semver values like (7). Currently, it prevent's overriding only semver values, but won't override `master`, `7`, `6117` (build number).

yavorsky:
This PR isn't blocked by this bug because it already presents on the babeljs.io

js/repl/PluginConfig.js
```diff
@@ -1,11 +1,62 @@
...
-  version: "0",
+  version: "1.6.2",
+  baseUrl: "https://unpkg.com",
+  versionKey: "envVersion",
+  instanceName: "babelPresetEnv",
+};
+
+const stage3Plugins: Array<PluginConfig> = [
+  "proposal-async-generator-functions",
+  // "proposal-object-rest-spread",
```
existentialism: oops, I probably confused you, but I meant to say that we should add a comment describing _why_ we need `stage3Plugins` (i.e. proposal-async-generator-functions not available >= beta.36)

js/repl/PluginConfig.js
```diff
@@ -58,20 +112,26 @@ const replDefaults: ReplState = {
   isPresetsTabExpanded: false,
   isSettingsTabExpanded: true,
   lineWrap: true,
+  presets: "es2015,react,stage-2",
```
existentialism: Nit: why was this moved? I kinda prefer this being alpha sorted :) (no big deal)

js/repl/Repl.js
```diff
@@ -98,18 +104,23 @@ class Repl extends React.Component {
...
-      builtIns: persistedState.builtIns,
+      babel: persistedStateToBabelState(persistedState, babelConfig),
       code: persistedState.code,
       compiled: null,
       compileErrorMessage: null,
       debugEnvPreset: persistedState.debug,
       envConfig,
       envPresetDebugInfo: null,
-      envPresetState: configToState(
+      envPresetState: persistedStateToEnvState(
```
existentialism: Not a blocker, but we should step back and re-think/clean up the state shape now that we have all a number of things that may or may not be loaded _and_ depending on version, etc

js/repl/ReplOptions.js
```diff
@@ -57,6 +60,22 @@ type Props = {
...
+  className?: string,
+  children: React$Element<any> | string,
+  section: string,
+};
+
+const LinkToDocs = ({ className, children, section }: LinkProps) => (
+  <a
+    className={className}
+    target="_blank"
+    href={`https://github.com/babel/babel/tree/master/experimental/babel-preset-env#${section}`}
```
existentialism: This link needs updating

js/repl/ReplOptions.js
```diff
@@ -304,17 +405,36 @@ class ExpandedContainer extends Component {
...
 
   _onDebugChange = (event: SyntheticInputEvent) => {
     this.props.onSettingChange("debugEnvPreset", event.target.checked);
   };
 
   _onElectronChange = (event: SyntheticInputEvent) => {
+    this.props.onEnvPresetSettingChange("electron", event.target.value);
+  };
+
+  _onShippedProposalsChange = (event: SyntheticInputEvent) => {
```
existentialism: Nit: we could replace these with something more generic for the checkboxes:

```js
_onEnvSettingChange = (event: SyntheticInputEvent) => {
  this.props.onEnvPresetSettingChange(event.target.name, event.target.checked)
}
```

js/repl/loadBundle.js
```diff
@@ -0,0 +1,136 @@
+import semver from "semver";
+import { loadBuildArtifacts, loadLatestBuildNumberForBranch } from "./CircleCI";
+import { BabelState } from "./types";
+import WorkerApi from "./WorkerApi";
+
+const SCOPED_CORE_VERSION_FROM = "7.0.0-beta.4";
```
existentialism: The things we do to ourselves :)

xtuc: aha exactly!

js/repl/PluginConfig.js
yavorsky: Oh, I just merged crookedly. Will fix, thanks :)

yavorsky:
@existentialism Thanks for the review! 🎉

PR #1420, opened 2017-10-15
Changed NODE_ENV to BABEL_ENV in .babelrc.md

Johannestegner:
The `BABEL_ENV` env variable was mistakenly written as `NODE_ENV` in the babelrc.md doc file.

### DIFF
--- a/docs/usage/babelrc.md
+++ b/docs/usage/babelrc.md
@@ -65,7 +65,7 @@ BABEL_ENV=production YOUR_COMMAND_HERE
 Or as a separate command:
 
 ```sh
-export NODE_ENV=production
+export BABEL_ENV=production
 ```
 
 ```sh


xtuc:
Thanks for the PR.

Actually both are correct, see [source](https://github.com/babel/babel/blob/0189b387026c35472dccf45d14d58312d249f799/packages/babel-core/src/config/helpers/environment.js#L4).

I would prefer `NODE_ENV` since it's the standard way of defined your env in Node.

Johannestegner:
Aye, I read that both are correct (well, babel_env seems to be checked before, but both works as you said) but at every other place in the file the `BABEL_ENV` is used.  
The reason for the PR was just that it was a thorn in my eye! ;)

xtuc:
Yes, you are right. 'BABEL_ENV` and `NODE_ENV` are the files that are inconsistent.

I think we should rename them to `NODE_ENV`. What do you think @existentialism?

Johannestegner:
Want me to swap the var to `NODE_ENV` in the PR?  
If so, want the `BABEL_ENV` to be mentioned as an override or rather leave it out for deprecation? :)

Johannestegner:
Ping @xtuc, want me to remove this PR or change it? :)

xtuc:
I would probably leave this as it is for now.

hzoo:
we can just make it consistent and change it later.

Issue #1263, opened 2017-06-12
Enhanced Examples


hzoo:
Ref https://github.com/babel/babel.github.io/issues/59

Maybe this is just weird, but it would be pretty cool to see real world examples of specific syntax out out on github. So if you are learning about arrow functions we can provide some basic examples but maybe you'd like to see how it's used in an actual codebase?

- [x] examples we can come up with (simple, easy to learn) #59
- [ ] examples from our unit tests (basically "syntax" coverage, but not as helpful to developers to learn from)
- [ ] examples from babel/github itself (no context but more real world usage)
  - [ ] hardcode a script to find kinds of syntax, or create an AST/Babel grep kind of tool that prints a codeframe of the relevant syntax
  - [ ] Use GitHub Archive/Google BigQuery's database to search github for uses of a certain syntax/pattern (filter JS repositories, filter by stars, date)? (may need to pin the commit/file)
  - [ ] populate examples?

PR #1421, opened 2017-10-15
Add AST panel to repl

danez:
This adds tabs to switch between babel output and AST.

I fixed also an issue, that an empty source was still compiled. I changed it to skip compilation if source is empty.

I had to change how the codemirror is rendered because of performance issues. The AST output can be quite large and it took 5 seconds to switch tabs. With the change from `textarea` to `div` and positioning the div absolute it is instant. (I found this behaviour on astexplorer)

There is one bug still, that I haven't figured out yet. In mobile when the options are uncollapsed and then get collapsed the editor is not using the complete height.

Design wise I am open to suggestions, I'm really not good with making design choices :)

bildschirmfoto 2017-10-15 um 12 36 28

### DIFF
--- a/js/repl/CodeMirror.js
+++ b/js/repl/CodeMirror.js
@@ -1,5 +1,6 @@
 // @flow
 
+import { css } from "emotion";
 import { injectGlobal } from "emotion";
 import CodeMirror from "codemirror";
 import React from "react";
@@ -39,22 +40,27 @@ export default class ReactCodeMirror extends React.Component {
   };
 
   _codeMirror: any;
-  _textAreaRef: HTMLTextAreaElement;
+  _container: HTMLDivElement;
+
+  shouldComponentUpdate() {
+    return false;
+  }
 
   componentDidMount() {
-    this._codeMirror = CodeMirror.fromTextArea(this._textAreaRef, {
+    this._codeMirror = CodeMirror(this._container, {
+      value: this.props.value || "",
+      placeholder: this.props.placeholder,
+      autofocus: this.props.autoFocus,
       ...DEFAULT_CODE_MIRROR_OPTIONS,
       ...this.props.options,
     });
     this._codeMirror.on("change", this._onChange);
-    this._codeMirror.setValue(this.props.value || "");
   }
 
   componentWillUnmount() {
-    // is there a lighter-weight way to remove the cm instance?
-    if (this._codeMirror) {
-      this._codeMirror.toTextArea();
-    }
+    const container = this._container;
+    container.removeChild(container.children[0]);
+    this._codeMirror = null;
   }
 
   componentWillReceiveProps(nextProps: Props) {
@@ -93,15 +99,7 @@ export default class ReactCodeMirror extends React.Component {
   }
 
   render() {
-    return (
-      <textarea
-        autoComplete="off"
-        autoFocus={this.props.autoFocus}
-        defaultValue={this.props.value}
-        ref={this._setTextAreaRef}
-        placeholder={this.props.placeholder}
-      />
-    );
+    return <div className={styles.editor} ref={c => (this._container = c)} />;
   }
 
   _updateOption(optionName: string, newValue: any) {
@@ -117,11 +115,18 @@ export default class ReactCodeMirror extends React.Component {
       this.props.onChange(doc.getValue());
     }
   };
-
-  _setTextAreaRef = (ref: HTMLTextAreaElement) => {
-    this._textAreaRef = ref;
-  };
 }
+
+const styles = {
+  editor: css({
+    overflow: "auto",
+    position: "absolute",
+    top: "0",
+    left: "0",
+    right: "0",
+    bottom: "0",
+  }),
+};
 
 injectGlobal({
   ".CodeMirror": {
--- a/js/repl/compile.js
+++ b/js/repl/compile.js
@@ -10,6 +10,7 @@ import type { BabelPresetEnvResult, CompileConfig } from "./types";
 
 type Return = {
   compiled: ?string,
+  ast: ?any,
   compileErrorMessage: ?string,
   envPresetDebugInfo: ?string,
   sourceMap: ?string,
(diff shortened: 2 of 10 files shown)


xtuc:
Really cool, nice job.

It seems that it added another scrolling bar because of the new header which is taking some space.

danez:
In which browser?

xtuc:
I'm running `Chrome version 58.0.3029.110 (64-bit)`.

This is what I can see:
screenshot from 2017-10-15 13-34-03
screenshot from 2017-10-15 13-34-14

danez:
Okay found it and fixed it, thanks @xtuc 

Also added the time the transformation took to the toolbar now.

xtuc:
Fixed on my side, thanks @danez

bvaughn:
Where's the Netlify preview build comment? 😄 

https://deploy-preview-1421--babel.netlify.com

bvaughn:
I like having the AST output handy, but I don't like losing a line of vertical space fro the tab bar.

bvaughn:
I like having the AST output handy, but I don't like losing a line of vertical space fro the tab bar.

I wonder if we could add this as a togglable option in the sidebar's "Settings" panel somehow? Maybe something like...

screen shot 2017-10-15 at 2 54 47 pm

hzoo:
@bvaughn has a good point about the vertical space, and I guess it would mostly make sense to reuse the menu bar we already have on the left in some way?

existentialism:
Yea... I'm fine with it in the sidebar. Guess we can solve the examples stuff later if we ever even get to it.

bvaughn:
Alternately we could maybe let people show/hide all 3 for more horizontal space too? Maybe this is inventing a feature that no one wants though 😁 

screen shot 2017-10-15 at 2 59 57 pm


>  Guess we can solve the examples stuff later if we ever even get to it.

What "examples stuff"? 😄

hzoo:
regarding examples: wanted a way to pre-fill in some input code to test out syntax etc for new users https://github.com/babel/website/issues/1263, ideally from real code though

bvaughn:
> regarding examples: wanted a way to pre-fill in some input code to test out syntax etc for new users #1263, ideally from real code though

I may be misunderstanding you, but isn't this already doable just via the query params?

hzoo:
@bvaughn Oh it is, just need a UI for it / the data/code snippets themselves

bvaughn:
> @bvaughn Oh it is, just need a UI for it / the data/code snippets themselves

Wouldn't the UI just be links (eg "Try this in the REPL") mixed throughout the non-JS portion of the site?

Or were you imagining some menu within the REPL to view pre-installed "recipes"?

hzoo:
> Imagining some menu within the REPL to view pre-installed "recipes"?

If that would be useful yes, but you are right it's already in those places as well true

bvaughn:
I kind of like keeping the REPL simple- kind of like a browser-based IDE. I think an external recipes section (eg somewhere else on the website) with links _to_ the REPL for common code patterns could be nice though.

hzoo:
We should definetely land this in some form now that babylon is in the monorepo, and it would be nice to have it remove loc or other ways to filter.

Although now that I think about it we could always just use this url to test haha

bvaughn:
I'm not 100% sold on the need for AST output here. Seems like it overlaps with [AST Explorer](https://astexplorer.net/), which is focused solely on AST exploration (and as such is much better IMO).

That being said, I still have the same concern I raised previously about the loss of vertical space with this change.

If you like either of the sidebar options I mentioned in previous comments, I can probably implement one of them on the flight to London tomorrow. Just let me know which you prefer.

hzoo:
Yeah, astexplorer does it much better for sure: has highlighting, filtering of properties, etc. We just want to be able to see the ast output per PR, but that can be alternatively dealt with by just running this PR's link: https://deploy-preview-1421--babel.netlify.com/repl/build/master or getting astexplorer itself to read any babel-standalone version just like babel repl is doing?

bvaughn:
> but that can be alternatively dealt with by just running this PR's link

Ha! I understand what you meant by that comment now 😆 Yeah, assuming Netlify never cleans up "preview" URLs, you're right.

> or getting astexplorer itself to read any babel-standalone version just like babel repl is doing?

Yup! That could also be nice! 😁

danez:
For me it is an essential feature missing from our repl, especially when working on pull requests to see what the AST looks like is something I do all the time. Right now I create dummy tests and change them, run tests, change test, run tests, ... and in the end have to remember not to commit this stuff, which is not a nice workflow. Also reproducing bugs with the latest master is something i do regularly and ASTexplorer always being outdated doesn't help with that.

I personally don't care where the option for the AST lives even if its a tab row, as screens are big enough imho and nobody is probably seriously using the repl on a phone. Also if we want to integrate other stuff the top bar could be useful.
On the other hand if we want to have options to disable output of locations for example then they might live in the sidebar, but splitting options is also not good I guess, so probably everything should be moved to the sidebar.
I also like the idea of being able to see the output and the AST at the same time.

I guess the main points of the repl is to let people see what babel is doing (where examples could come handy), let people reproduce bugs in an easy way (source code as well as AST) and for maintainers to investigate bugs.

When I have time I  will remove the tabbar and move the AST option to the sidebar and add an option to remove locations.

danez:
I changed it now to be a separate panel instead of tabs.

bvaughn:
I like this change, Daniel!

js/repl/CodeMirror.js
```diff
@@ -39,22 +40,27 @@ export default class ReactCodeMirror extends React.Component {
...
   }
 
   componentWillUnmount() {
-    // is there a lighter-weight way to remove the cm instance?
-    if (this._codeMirror) {
-      this._codeMirror.toTextArea();
-    }
+    const container = this._container;
+    container.removeChild(container.children[0]);
+    this._codeMirror = null;
```
bvaughn: I guess we didn't actually need to call `toTextArea()` ? 😄

js/repl/compile.js
```diff
@@ -69,15 +96,30 @@ export default function compile(code: string, config: CompileConfig): Return {
...
     compiled = transformed.code;
 
+    if (!config.astLocations) {
+      transformed.ast = filterLocation(transformed.ast);
+    }
+    if (!config.astTokens) {
+      delete transformed.ast.tokens;
+    }
+
+    ast = JSON.stringify(transformed.ast, null, 2);
```
xtuc: I'm just wondering, the AST shows the emitted code, wouldn't it more practical to show the input?

For example JSX: [REPL](https://deploy-preview-1421--babel.netlify.com/repl/#?astLocations=false&astTokens=false&babili=false&browsers=&build=&builtIns=false&code_lz=DwYQ9gtgDmB2CmsAuACA9APiA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&outputAST=true&outputCode=false&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0)

js/repl/CodeMirror.js
danez: As the source is not a textarea anymore, it doesn't make sense to convert to to textarea back. I stole this from somewhere, but cannot recall from where.

js/repl/compile.js
danez: Good point, you can of course by hand disable all presets, but that might be forgotten.

So I see to things to do here:
* Either ensure that no presets and plugins are supplied to babel when outputing the AST.
* Have a separate compiler that uses babylon directly. 

The first one is probably easier. And in any case all other options (presets, env) should be disabled in the UI when selecting AST.

xtuc: I would take the second option.

You have two possible outputs:
- A Babel transform (as currently implemented)
- An AST, Babylon could directly parse it using a few Babylon plugin from the plugins/presets.

hzoo:
You're free to land this on master and `old-site` if you want

PR #1424, opened 2017-10-18
Update mocha `--compiler` deprecation syntax

huchenme:
https://github.com/mochajs/mocha/wiki/compilers-deprecation
Mocha will deprecate `--compiler` syntax

### DIFF
--- a/_includes/tools/mocha/usage.md
+++ b/_includes/tools/mocha/usage.md
@@ -1,3 +1,9 @@
+#### Mocha 4
+
+Updated in Mocha v4.0.0
+
+`--compilers` is deprecated as of Mocha v4.0.0. See [further explanation and workarounds](https://github.com/mochajs/mocha/wiki/compilers-deprecation).
+
 In your `package.json` file make the following changes:
 
 ```json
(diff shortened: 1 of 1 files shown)


xtuc:
Thanks for your PR.

In which version of Mocha has the syntax changed? I'm wondering if we shouldn't keep both syntax just in case. What do you think?

huchenme:
@xtuc I see the deprecation warning in mocha version 4

xtuc:
IMO I would keep both versions of the argument and mention the version.

huchenme:
@xtuc how do you think the updated version?

_includes/tools/mocha/usage.md
```diff
@@ -1,9 +1,15 @@
+#### Mocha 4
+
+Updated in Mocha v4.0.0
```
xtuc: Could you maybe remove the first sentence, it's redundant with the rest to me.

huchenme: Thanks, done

hzoo:
Great, thanks @huchenme !

PR #1433, opened 2017-10-25
Remove mention of babel-plugin-undeclared-variables-check

vp2177:
It doesn't work properly, so remove mention of this plugin.

For example, it results in the following error:
```
    ERROR in ./demo/react/main.jsx
    Module build failed: BabelLoaderError: /home/ubuntu/workspace/demo/react/main.jsx: Reference to undeclared variable "props"
    
      11 | class Demo extends Component { 
      12 |     constructor(props) {
    > 13 |         super(props)
         |               ^
      14 |         this.state = {
```

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -184,7 +184,6 @@ These plugins are in the minify repo.
 ## Misc Plugins
 
 - [external-helpers](/docs/plugins/external-helpers/)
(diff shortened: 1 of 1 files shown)


hzoo:
Thanks! yeah we don't really intend to support this anyway because you can use ESLint for that

PR #1440, opened 2017-10-26
Update FAQ, babel-preset-es2015 → babel-preset-env

lachlanjc:
As far as I understand [babel-preset-es2015 is deprecated](http://babeljs.io/env), so we should not mention installing it anywhere in the docs. Please let me know if I'm misunderstanding :)

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -63,12 +63,12 @@ get all the same transforms that were in Babel 5:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


hzoo:
👌 no misunderstanding, just an out of date doc (and weird since we are going to make a v7). Maybe we should just remove it entirely haha?

Thanks!

Issue #1402, opened 2017-10-04
REPL: remove parts of the url that are falsy/empty since unncessary


hzoo:
So for a url like

`https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=IZA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015-loose&targets=&version=7.0.0-beta.0`

Some of the fields that are either `&key=` or `&key=false` can be removed and it should work the same?
`babili=false&browsers=&build=&builtIns=false&debug=false&circleciRepo=&evaluate=false&targets=`

This is because we don't have default true/false so not setting false should be ok. Basically then we can have a shorter url

(unless we create a button somewhere to shorten the url or something)

Daniel15:
Instead of stripping `false`, we should strip params that have the default value. It's possible we already have params that default to `true`, or might in the future.

hzoo:
Much better idea ^, thanks Daniel 😄

rajzshkr:
I can pick this.

rajzshkr:
has PR #1441

bvaughn:
> Some of the fields that are either &key= or &key=false can be removed and it should work the same?

> This is because we don't have default true/false so not setting false should be ok. Basically then we can have a shorter url

I have concerns about this proposed change, as mentioned [in this comment](https://github.com/babel/website/pull/1441/#discussion_r147421106).

PR #1441, opened 2017-10-26
REPL - remove parts of the url that are falsy/empty #1402

rajzshkr:


rajzshkr:
@Daniel15 I'll update and keep posted here.

rajzshkr:
Ping @Daniel15

js/repl/UriUtils.js
```diff
@@ -4,20 +4,22 @@ import LZString from "lz-string";
...
+const URL_DEFAULTS = {
+  babili: false,
+  browsers: "",
+  build: "",
+  builtIns: false,
+  circleciRepo: "",
+  code: "",
+  debug: false,
+  evaluate: false,
+  lineWrap: true,
```
Daniel15: This looks okay to me, but I'd like @bvaughn to take a look too (if he gets time to). It seems like [replUtils.js](https://github.com/babel/website/blob/f34ae222d3f7edf70711e1c849f13418da95b4cb/js/repl/replUtils.js#L52) is also doing some defaulting, so I wonder if we could unify the two such that they both share the same set of defaults.

rajzshkr: That make sense. I'll make it generic and update the PR.

js/repl/UriUtils.js
```diff
@@ -84,7 +86,7 @@ const updateQuery = (state: PersistedState) => {
       return null;
     } else if (key === "code") {
       return `${key}_lz=` + compress(state.code);
-    } else {
+    } else if(state[key] !== URL_DEFAULTS[key]){
       return key + "=" + encode(state[key]);
```
bvaughn: This isn't quite right.

js/repl/UriUtils.js
```diff
@@ -84,7 +86,7 @@ const updateQuery = (state: PersistedState) => {
       return null;
     } else if (key === "code") {
       return `${key}_lz=` + compress(state.code);
-    } else {
+    } else if(state[key] !== URL_DEFAULTS[key]){
```
bvaughn: This isn't quite right. The way the REPL is setup is this:
* If a flag is specified in the URL, we use its value
* Else we check `localStorage` for the most recently-used value and use it,
* Else we fallback to a (hard-coded) default

bvaughn: If folks feel that the convenience of using localStorage to remember last-used values isn't important, then we could drop it and move forward with this change.

Alternately we could change the REPL to only use localStorage if there's a special flag in the URL (which we could pass in from the REPL link on the website). That's the most useful use case of it really.

Maybe it's better to only read/write localStorage when we have explicit permission anyway? This would prevent a shared link from overriding our last explicitly chosen settings?

existentialism: FWIW, I'd like to keep localstorage... just need to iron out a few bugs and I think it's fine.

bvaughn: Bugs with our localStorage use? Hm. I wasn't aware of any. :)

rajzshkr: @bvaughn @existentialism Thanks for the clarification. So I'm gonna Make the priority in this order **URL > LOCAL STORAGE > DEFAULTS**.

I'll keep posted once the change is done.

existentialism: @rajzshkr 👍

Daniel15:
What if we only use LocalStorage if there are no URL parameters? Would that
solve the issue?

Sent from my phone.

On Oct 27, 2017 8:10 AM, "Brian Ng" <notifications@github.com> wrote:

*@existentialism* commented on this pull request.
------------------------------

In js/repl/UriUtils.js
<https://github.com/babel/website/pull/1441#discussion_r147437102>:

> @@ -84,7 +86,7 @@ const updateQuery = (state: PersistedState) => {
       return null;
     } else if (key === "code") {
       return `${key}_lz=` + compress(state.code);
-    } else {
+    } else if(state[key] !== URL_DEFAULTS[key]){

@rajzshkr <https://github.com/rajzshkr> 👍

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/babel/website/pull/1441#discussion_r147437102>, or mute
the thread

bvaughn:
> What if we only use LocalStorage if there are no URL parameters? Would that
solve the issue?

That's kind of similar to what I mentioned in [my comment above](https://github.com/babel/website/pull/1441#discussion_r147423058).

Essentially I think the `localStorage` fallbacks are useful only when you're coming to the REPL as a direct link (eg [https://babeljs.io/repl/](https://babeljs.io/repl/)). So whether it's the lack of all URL parameters, or a special one that signifies "it's okay to restore and update default settings", I think it's fine. 👍

This would have an added benefit of addressing something that's always nagged me a little- which is to say, that viewing a link shared by someone else can clobber your `localStorage` values, which isn't always obvious or expected.

Daniel15:
> This would have an added benefit of addressing something that's always nagged me a little- which is to say, that viewing a link shared by someone else can clobber your localStorage values

The opposite is also true - Your localStorage can affect a REPL link if the variable is not in the URL, as it'll use the variable from localStorage. I really think localStorage and URL parameters need to be handled totally separately - Maybe ignore localStorage if using a URL? Not sure.

bvaughn:
> The opposite is also true - Your localStorage can affect a REPL link if the variable is not in the URL

True, but unless someone manually edits the URL to remove variables, this would not occur.

Still, I think it sounds like we're moving toward consensus in changing the behavior of localStorage, which I dig.

rajzshkr:
Ping @Daniel15 @bvaughn

js/repl/replUtils.js
```diff
@@ -33,27 +33,28 @@ export const loadPersistedState = (): PersistedState => {
   const storageState = StorageService.get("replState");
   const queryState = UriUtils.parseQuery();
   const merged = {
+    ...persistedStateDefaults,
     ...storageState,
```
bvaughn: Values in `localStorage` will override defaults from `persistedStateDefaults `

js/repl/UriUtils.js
```diff
@@ -84,7 +85,9 @@ const updateQuery = (state: PersistedState) => {
       return null;
     } else if (key === "code") {
       return `${key}_lz=` + compress(state.code);
-    } else {
+      // Default state can be empty string or a boolean value
+      // URL State should not be empty string and default state
+    } else if(state[key] !== "" && state[key] !== persistedStateDefaults[key]){
```
bvaughn: I raised concerns about this initially. Removing values from the URL could result in a shared link behaving differently between 2 users b'c of `localStorage` values (see my comment below). If we remove values from the URL (eg your changes in `UriUtils`) we need to also change the override/restore behavior elsewhere. I think for now, we should revert the changes to `UriUtils`.

rajzshkr:
> I'm a little confused why commits from PR #1442 are showing up in this diff. Could you rebase to make sure this isn't accidentally overriding any changes from that PR?

I did rebase again, But I don't know why the other commits are showing in the PR

>It also looks like CI is failing for what looks to be several linting and formatting errors. (See below.) Have you ran Prettier?

I run **npm run precommit** every time I commit. it didn't show these error for me.

rajzshkr:
Closing this in favour of #1450

bvaughn:
> I run npm run precommit every time I commit. it didn't show these error for me.

Hm. Interesting. It's possible then that those syntax issues were caused by the commit that made its way into the PR mysteriously, since `precommit` runs `lint-staged` which only covers files that are staged (about to be committed).

PR #1449, opened 2017-10-31
Add more videos

zsoltime:
I've just added some more videos. 🤓


hzoo:
Based on this I was thinking we should have topics similar to the team page?

Not sure how we want to implement/display but there is a differentiation between videos that explain how babel itself works (ASTs, etc) vs. what it is/how to use it as a user, and I guess maintainer/team talks vs. community. There are other talk types like explaining TC39, proposals, maintenance/oss

PR #1455, opened 2017-11-05
Add babel-core to browserify install guide

### LINKED ISSUE #1452: Babelify missing install of babel-core
1. Go to https://babeljs.io/docs/setup/
2. Choose Browserify

Under the section "Installation" it says:
```
npm install --save-dev babelify
```
While I believe it should be as in https://github.com/babel/babelify#installation
where it says the following:
```
npm install --save-dev babelify babel-core
```

azz:
Fixes #1452

https://deploy-preview-1455--babel.netlify.com/docs/setup/#installation

### DIFF
--- a/docs/tools/browserify/install.md
+++ b/docs/tools/browserify/install.md
@@ -1,3 +1,3 @@
 ```sh
-npm install --save-dev babelify
+npm install --save-dev babelify babel-core
 ```


hzoo:
This is only for v7 though, so not sure unless we have a separate doc for v6 and v7

hzoo:
Ok another one we should do after moving to versioned docs in https://github.com/babel/website/issues/1485, maybe you can help with that 😄

PR #1456, opened 2017-11-05
Added babel core and env to the gulp install guide

FabioAntunes:


FabioAntunes:
nevermind it seems thats for gulp-babel beta 8

PR #1461, opened 2017-11-13
Adding .mjs to watching file list

rajzshkr:
https://github.com/babel/babel/issues/6800


hzoo:
I think we should make this kind of change after https://github.com/babel/website/issues/1485, and also we should move that information into the readme itself not in this repo?

PR #1462, opened 2017-11-14
Add flow and typescript presets to repl

existentialism:
Supersedes/Closes https://github.com/babel/website/pull/1360.

Only real difference is presets marked with `isPreLoaded` inside `presetPluginConfigs` are now  checked against `availablePresets` from `@babel/standalone` to prevent the invalid/missing preset errors (i.e. trying to load Flow/TypeScript preset with 6.x)

Examples:

[6.26.0](https://deploy-preview-1462--babel.netlify.com/repl#?babili=false&browsers=&build=&builtIns=false&code_lz=PTAEAEDMBsHsHcBQA3AhgJ1ADwFygM4Au6AlgHYDmoAvKAOSSywBGGdA3EA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&targets=&version=6.26.0) (no flow/ts)

[7.0.0-alpha.15](https://deploy-preview-1462--babel.netlify.com/repl#?babili=false&browsers=&build=&builtIns=false&code_lz=PTAEAEDMBsHsHcBQA3AhgJ1ADwFygM4Au6AlgHYDmoAvKAOSSywBGGdA3EA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&targets=&version=7.0.0-alpha.15) (no flow/ts)

[7.0.0-alpha.20](https://deploy-preview-1462--babel.netlify.com/repl#?babili=false&browsers=&build=&builtIns=false&code_lz=PTAEAEDMBsHsHcBQA3AhgJ1ADwFygM4Au6AlgHYDmoAvKAOSSywBGGdA3EA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&targets=&version=7.0.0-alpha.20) (flow/ts)

### DIFF
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -114,7 +114,8 @@ class Repl extends React.Component {
       isSidebarExpanded: persistedState.showSidebar,
       lineWrap: persistedState.lineWrap,
       plugins: configArrayToStateMap(pluginConfigs, defaultPlugins),
-      presets: configArrayToStateMap(presetPluginConfigs, defaultPresets),
+      // Filled in after Babel is loaded
+      presets: {},
       runtimePolyfillState: configToState(
         runtimePolyfillConfig,
         persistedState.evaluate
@@ -122,7 +123,7 @@ class Repl extends React.Component {
       sourceMap: null,
     };
 
-    this._setupBabel();
+    this._setupBabel(defaultPresets);
   }
 
   render() {
@@ -199,9 +200,22 @@ class Repl extends React.Component {
     );
   }
 
-  async _setupBabel() {
+  async _setupBabel(defaultPresets) {
     const babelState = await loadBabel(this.state.babel, this._workerApi);
-    this.setState(babelState);
+
+    // Filter the list of preloaded presets with those available with
+    // the loaded version of @babel/standalone.
+    const availablePresetConfigs = babelState.isLoaded
+      ? presetPluginConfigs.filter(
+          ({ label, isPreLoaded }) =>
+            !isPreLoaded || babelState.availablePresets.indexOf(label) > -1
+        )
+      : presetPluginConfigs;
+
+    this.setState({
+      babel: babelState,
+      presets: configArrayToStateMap(availablePresetConfigs, defaultPresets),
+    });
 
     if (babelState.isLoaded) {
       this._compile(this.state.code, this._checkForUnloadedPlugins);
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -16,7 +16,6 @@ import PresetLoadingAnimation from "./PresetLoadingAnimation";
 import {
   envPresetConfig,
   pluginConfigs,
-  presetPluginConfigs,
   runtimePolyfillConfig,
 } from "./PluginConfig";
 import {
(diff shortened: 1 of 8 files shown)


js/repl/Worker.js
```diff
@@ -23,6 +23,13 @@ registerPromiseWorker(message => {
         return null;
       }
 
+    case "getAvailablePresets":
+      try {
```
Andarist: why can it throw?

js/repl/Repl.js
```diff
@@ -199,9 +200,22 @@ class Repl extends React.Component {
...
+  async _setupBabel(defaultPresets) {
     const babelState = await loadBabel(this.state.babel, this._workerApi);
-    this.setState(babelState);
+
+    // Filter the list of preloaded presets with those available with
+    // the loaded version of @babel/standalone.
+    const availablePresetConfigs = babelState.isLoaded
+      ? presetPluginConfigs.filter(
+          ({ label, isPreLoaded }) =>
+            !isPreLoaded || babelState.availablePresets.indexOf(label) > -1
```
bvaughn: I'm just glancing at this PR while I'm waiting on a build to complete, so I'm probably misreading something silly, but this line doesn't make sense to me:
```js
!isPreLoaded || babelState.availablePresets.indexOf(label) > -1
```

If we're explicitly validating that the preset is available (loaded) then why are we even checking `isPreLoaded` (let alone treating `!isPreLoaded` the same as one that's actually been verified and loaded)?

existentialism: Originally, I actually replaced `presetPluginConfigs` with one generated from `babelState.availablePresets` but decided (perhaps foolishly) that maybe it was better to preserve the ability to have presets listed in the config that weren't preloaded and would get parsed later in `_checkForUnloadedPlugins`.

At least thats how I've been interpreting the intent of that bool for the entries in `presetPluginConfigs`.

js/repl/Worker.js
Daniel15: +1

Are you guarding against `Babel` being `undefined` (ie. Babel hasn't loaded yet)? You should probably just check `!!Babel` in that case.

js/repl/Repl.js
bvaughn: If I recall correctly, I actually added the bool thinking that we might want to add more presets eventually that required runtime loading (similar to how we load eg Prettier). Maybe we should just ditch it though, since they're all true now.

Either way, it seems odd to me that we filter this way. 😄  Seems like the `isPreLoaded` isn't important if we're explicitly checking the plugins that standalone Babel's reporting it has.

existentialism: I guess I was thinking that the filtered list should include the `isPreLoaded` presets that _are_ available via babel/standalone and any additional non preloaded presets to pass down to the next step in the process (at that point it can eventually runtime load stuff, etc).

Maybe we just simplify it altogether and generate `presetPluginConfigs` based on available presets from babel/standalone and tackle runtime loading differently (similar to preset-env).

bvaughn: I think that would be a nice simplifying change for the time being 😄

existentialism:
Updated!

bvaughn:
Nice work 😄

bvaughn:
Maybe we could `.sort()` the new dynamic plugins list though? It used to be alpha-sorted in the config. Now it looks like "react" drops to the bottom.

PR #1466, opened 2017-11-19
Re-enable prettier v1.6.1

bvaughn:
Follow up for PR #1395

Looks like Prettier v1.7.0 broke for the browser with the error "_cosmiconfig requires at least version 4 of Node_" and v1.8.0 broke with the error "_os$2.homedir is not a function_". Version 1.6.1 was working before though, so I'm re-enabling and pinning us to that version for the time being.

Note that I'm seeing 3 Flow errors in this branch when I run `yarn flow` locally, but I also see those 3 errors on master. ☹️  Seems like this got broken at some point?

### DIFF
--- a/js/repl/CircleCI.js
+++ b/js/repl/CircleCI.js
@@ -22,7 +22,7 @@ async function sendRequest(repo: ?string, uri: string): Promise<Object> {
 export async function loadBuildArtifacts(
   repo: ?string,
   build: string,
-  cb: (url: string, error?: string) => void
+  cb: (url: string, error?: string) => void // eslint-disable-line no-unused-vars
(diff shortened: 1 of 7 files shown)


existentialism:
@bvaughn yeah, IIRC flow broke when https://github.com/babel/website/pull/1411 landed.

This looks fine to me, although it kinda sucks that there's now the possibility for supported syntax mismatches due to two versions of Babylon, nbd I guess since prettier is opt in. 

Any thoughts @hzoo?

bvaughn:
Once prettier/prettier/issues/3296 is resolved then we can get back on the latest release 😄

PR #1473, opened 2017-11-28
Add the MIT license

### LINKED ISSUE #1251: LICENSE file?
It seems that this repository does not contain a LICENSE. I'd like to use a subset of the code, but I need to make sure we're license-compatible. (We are planning to use Apache 2.)

j-f1:
This is the same one as the one in babel/babel. Closes #1251.


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2014–2017 Sebastian McKenzie <sebmck@gmail.com>
```
j-f1: Should this include @hzoo too?

_includes/footer.html
```diff
@@ -1,6 +1,15 @@
 <div class="footer text-center">
-  <p><a href="https://github.com/babel/babel">{{ site.title }}</a> · <a href="https://github.com/babel/babel/blob/master/LICENSE">Distributed under MIT License</a> · <a href="https://github.com/babel/babel/blob/master/CODE_OF_CONDUCT.md">Code of Conduct</a></p>
-  <p><a href="http://henryzoo.com/babel.github.io/">Looking for Babel 5.x docs?</a> · Found an issue with the docs? Report it <a href="https://github.com/babel/babel.github.io/issues/new">here</a>.</p>
+  <p>
+    <a href="https://github.com/babel/babel">{{ site.title }}</a>
+    · <a href="https://github.com/babel/babel/blob/master/LICENSE">Babel</a>
+      and <a href="https://github.com/babel/website/blob/master/LICENSE">this website</a>
```
j-f1: Should this say “this website” or “the Babel website” or something else?

xtuc:
To be honest, we (or only I) are not lawyers, and I'm not sure what to do with this.

Initial credits goes to @kittens and without his agreement it can't be modified by us.

j-f1:
cc @xtuc Does this look better? This is the license currently specified in `package.json`.

kittens:
👍 MIT 👎 ISC

kittens:
ISC is probably in `package.json` because it's the default value of `npm init`.

kittens:
Thanks @j-f1 :heart:

PR #1478, opened 2017-12-08
Fix link to Wikipedia

sonicdoe:
The previous link led to a disambiguation page on Wikipedia.


existentialism:
@sonicdoe thanks!

PR #1479, opened 2017-12-09
Updated transform-class-properties package link

bodaz:


existentialism:
@bodaz thanks for the PR, but the website currently represents 6.x documentation. we'll be updating the docs with a version switcher and 7.x docs asap!

bodaz:
@existentialism  thanks for the response! btw how do i escape the @ sign?? Netlify deploy failed because of that

hzoo:
The package folder name is still without the `@` so it's correct

Hopefully we can just fix this issue with https://github.com/babel/website/issues/1485, thanks for the pr

PR #1488, opened 2017-12-17
 Don’t override babel version if it’s explicit.

yavorsky:
This is a workaround for https://github.com/babel/babel/issues/7016 and https://github.com/babel/babel/issues/6174
Each babel-standalone bundle includes the previous version, so if, for example, we are loading `@babel/standalone@7.0.0-beta.34`, then `Babel.version` will return `7.0.0-beta.33`

So, we are overriding version each time we reload the page and change some query params. 
It's easy to reproduce it, just by reloading the page several times and compare version in the query or right left corner.

[Try](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=true&code_lz=IwbiA&debug=false&circleciRepo=&evaluate=true&lineWrap=false&presets=&prettier=true&targets=&version=7.0.0-beta.34).

After publishing process will be fixed, we still will have old bundles with the previous version, so proposed solution is to don't override version if it's explicit semver value, and override for versions from the bundle for builds from CI.


Daniel15:
Interesting idea. It also means that we won't override the version when a range is passed (eg. `6` will stay as `6` rather than changing to `6.26.0`)... Is that expected? That actually seems better, as currently going to https://babeljs.io/repl/version/6/ will end up with two version numbers in the URL (`6` in the path and `6.26.0` in the querystring).

Let's still fix the version number in the standalone package, of course,. 😄

yavorsky:
> That actually seems better, as currently going to https://babeljs.io/repl/version/6/ will end up with two version numbers in the URL (6 in the path and 6.26.0 in the querystring).

On the other hand, we couldn't exactly know what version we are using for range values like 6. The only way is by opening network tab in the inspector.

Currently, it won't override version for range values, and will just for valid semver values, but we can also add range values support if showing just `6` as a version is ok.

xtuc:
What's the status @yavorsky?

yavorsky:
@xtuc I think it's ok to merge. And according to the babeljs.io it wasn't fixed with other updates.

PR #1492, opened 2017-12-20
fix dead link to emotion/macro docs

tomByrer:


xtuc:
The change is https://github.com/emotion-js/emotion/blob/master/docs/babel-macros.md -> https://github.com/emotion-js/emotion/blob/master/docs/babel.md#babel-macros

PR #1495, opened 2017-12-26
Fix typo

reod:


xtuc:
Good catch!

I changed a bit the sentence, what do you think? Otherwise lgtm.

Issue #1485, opened 2017-12-15
Switch to Docusaurus


hzoo:
EDIT: https://github.com/babel/website/pull/1521

https://twitter.com/fbOpenSource/status/941370631357997056

Docusaurus was released: https://docusaurus.io/blog/2017/12/14/introducing-docusaurus.html

Used on a few FB projects like prettier, jest, react-native, etc.

I mostly am in favor because other projects are using that have transitioned to it easily (doesn't seem hard since it's markdown but also can use react for custom stuff) and it has 2 features we are missing (and leaves the two we already have which is algolia search and markdown).

We need translation + site versioning. Translation is definetely a want, and site versioning is literally an issue right now while we are developing Babel 7 but everyone getting confused at the documentation being v6 while the repo is in v7.

- Translations: https://github.com/babel/website/issues/135
- Whether to add new docs: https://github.com/babel/website/issues/1257
- Update links/repos https://github.com/babel/website/pull/1496
---

Also https://twitter.com/gatsbyjs which is what react is using, and what I was thinking about earlier. It does seem like the jekyll -> docusaurus doesn't seem that hard so we could switch to that later anyway?

Always looking for help for anyone that would like to make this happen!

JoelMarcey:
@hzoo Hi. I am a member of the team that helped create Docusaurus. Definitely let us know if you have any questions. We tried to keep the initial launch simple in terms of format support. 

- markdown (which gets generated to HTML)
- static assets (CSS, images, html)

We are looking at other support moving forward though.

KyleAMathews:
Happy to answer questions about Gatsby as well + help with transition as well if that's the decision!

maaz93:
Just took a very quick look at Docusaurus. I think it'd be a fun experience to give it a spin using the Babel docs.

@hzoo can I give it a shot? If yes, anything specific you'd like me to check while attempting the migration?

hzoo:
Nothing in particular I guess we just need to check the REPL still can work, etc. I'm ok if we have to change styles too if it's more structured. Maybe want to land in another branch so can iterate too

QC-L:
I think switch is necessary. I'm the maintainer of [babeljs.cn](https://babeljs.cn/).

xtuc:
@QC-L Ideally we should have a way to translate it easily in this repo (like a `en` and `cn` folder).

hzoo:
[...]

PR #1496, opened 2017-12-28
Update download-readmes.js -> Babylon

QC-L:
Babylon has been moved into the main Babel mono-repo.
Change download uri to /babel/babel/master/packages/babylon/README.md.

### DIFF
--- a/scripts/download-readmes.js
+++ b/scripts/download-readmes.js
@@ -45,7 +45,7 @@ Promise.all([getDirectoryListing('babel', '6.x'), getDirectoryListing('minify')]
       },
       {
         name: 'babylon',
-        uri: '/babel/babylon/master/README.md',
+        uri: '/babel/babel/master/packages/babylon/README.md',
       },
     ];
 


QC-L:
@yavorsky Not at all.

yavorsky:
Hmm.. I just remembered why we haven't done that before...
Actually, we haven't released 7 yet, so the website doesn't refer to any 7 stuff.

Look at the babel-preset-env docs uri: https://github.com/QC-L/babeljs.io/blob/b3e134dc6fb068ae6be04236afe1d7a1b9718f9d/scripts/download-readmes.js#L44

So we need to be consistent and update preset-env also since we've moved it to the monorepo like Babylon, or use Babylon's 6.x branch because preset-env refers to 1.x.
I'd prefer the 2 option unless Babel 7 released.

Sorry for the initial confusion.

QC-L:
I see. Thank you. Look forward to the release of Babel 7.

yavorsky:
@QC-L You can update this PR with a branch name change to `6.x`. 🙂

QC-L:
Nothing, no hurry. Wait for you to update.

hzoo:
Yeah with https://github.com/babel/website/issues/1485 - we should have versioned docs so it won't be as confusing

hzoo:
Can close now since we have the new docs

PR #1500, opened 2017-12-30
Add Statusbar  on bottom of editor to display status 

samundrak:
This PR allows you to see the size of the code of both raw and compiled so whenever people want to transpile there code they will also see how much bytes going to be added on there file. This might be helpful
as if we need to see minification size then we have to go another website, it will be better if we get this information in one place.

babel-status-bar


hzoo:
Very cool, something like being able to see the size/gzip size sounds cool to me! cc @developit?

bvaughn:
First thought: I'm not sure how I feel about having _both_ a `yarn.lock` and a `package-lock.json` file. How do we ensure they remain in sync? Which is the source of truth if they ever disagree? I think this new file should be removed.

bvaughn:
Third thought- in retrospect it should have been the first. This is really cool! 😄

hzoo:
Like if you just inline it within the pane of the code or something on the bottom right, absolutely positioned?

bvaughn:
For example, what about CSS something more like this:
```css
    position: absolute;
    top: 1rem;
    right: 1rem;
    border-radius: 0.5rem;
    padding: 0.5rem 0.5rem 0.5rem 0.25rem;
    background-color: rgba(100,100,100,0.2);
    border: 0;
    color: rgba(0,0,0,0.35);
```

screen shot 2017-12-30 at 8 39 45 am

bvaughn:
Or yeah, like Henry said, they could float bottom+right instead.

js/repl/compile.js
```diff
@@ -111,5 +115,6 @@ export default function compile(code: string, config: CompileConfig): Return {
     compileErrorMessage,
     envPresetDebugInfo,
     sourceMap,
+    meta,
```
bvaughn: tiny nit: preserve alpha-sorting

js/repl/CodeMirrorPanel.js
```diff
@@ -13,10 +14,18 @@ type Props = {
...
 export default function CodeMirrorPanel(props: Props) {
-  const { className = "", errorMessage, info, onChange } = props;
+  const {
+    className = "",
+    errorMessage,
+    info,
+    onChange,
+    status,
+    options,
+  } = props;
```
bvaughn: tiny nit: preserve alpha-sorting

js/repl/PluginConfig.js
```diff
@@ -57,6 +57,12 @@ const replDefaults: ReplState = {
   isPresetsTabExpanded: false,
   isSettingsTabExpanded: true,
   lineWrap: true,
+  statusBar: false,
+  status: {
+    compiled: {},
+    raw: {},
+  },
```
bvaughn: Not sure how I feel about this nesting structure. Do we think we'll include things here _other_ than size? If not, do we _need_ to nest this 3 levels deep? (Couldn't we just do `rawSize` and `compiledSize`?)

js/repl/StatusBar.js
```diff
@@ -0,0 +1,38 @@
...
+    maxHeight: "33%",
+    overflow: "auto",
+    margin: 0,
+    padding: "0.5rem 0.75rem",
+    whiteSpace: "pre-wrap",
+    "-webkit-overflow-scrolling": "touch",
+  }),
+  statusItem: css({
+    marginLeft: "5px",
+  }),
```
bvaughn: I left a comment on the main thread, but I think it might be nicer to float this rather than show it as a bar. This way it wouldn't take up any vertical space from the text layout. 😄

package.json
```diff
@@ -1,78 +1,76 @@
...
+            "lint-list-item-indent": false,
+            "lint-no-blockquote-without-caret": false,
+            "lint-no-literal-urls": false,
+            "lint-no-duplicate-headings-in-section": true,
+            "lint-no-empty-url": true
+        }
+    },
+    "author": "",
+    "license": "ISC",
+    "devDependencies": {
```
bvaughn: Please revert all of the white-space only changes to this file if possible. 😄 

So far as I can see, the only 2 _actual_ changes to this file are:
* The addition of the `prettysize` module
* An (unnecessary?) formatting change to the `lint-staged` block

webpack.config.js
```diff
@@ -18,13 +18,12 @@ const config = {
     rules: [
       {
         test: /\.js$/,
-        exclude: /node_modules/,
+        exclude: /node_modules\/(?!(prettysize)\/).*/,
```
bvaughn: Why is this change necessary?

samundrak:
The benefit of the bar being at the bottom is we can add other information in future like execution time a code is taking or any other which may come in future. 
And Sorry for package-lock.json

bvaughn:
> The benefit of the bar being at the bottom is we can add other information in future like execution time a code is taking or any other which may come in future.

If we decide to add other info in the future, we could always reconsider the placement of this bar. For now though, while there's only one thing being shown, (and nothing else being actively planned), I think it's best to optimized for showing as much text as we can on small screens. 😄

js/repl/PluginConfig.js
samundrak: I think with new changes nested object won't be necessary :smile:

webpack.config.js
samundrak: Actually, this module is not already transpiled and Uglify is throwing an error because babel ignores it while building, any suggestion for this?

bvaughn: Hm. The problem seems to be with this line:
```js
module.exports = (size, nospace, one, places) => {
```

But suppressing the Uglify warning is just masking the symptom. It will still leave an arrow function in the built bundle, which will [break IE 11 (and older)](https://caniuse.com/#feat=arrow-functions).

Taking a look at `prettisize`, the amount of work it's doing is trivial- so it's probably easier to just rewrite this logic ourselves in a util function. 😄

samundrak: Actually here babel ignores node_modules but not prettysize  so no arrow function will be in the bundle but yes I think it's not a good idea to be specific for a module. :smile:

bvaughn: Ah! I misread. My apologies. 😄

samundrak:
any update on this pr, i would love to see this feature at babel/repl :smile:

js/repl/Repl.js
```diff
@@ -317,7 +329,14 @@ class Repl extends React.Component {
         sourceMap: runtimePolyfillState.isEnabled,
         useBuiltIns: state.builtIns,
       })
-      .then(result => this.setState(result, setStateCallback));
+      .then(result => {
+        const { raw, compiled } = result.meta;
+        this._updateStatus({
+          rawSize: prettySize(raw.size),
+          compiledSize: prettySize(compiled.size),
+        });
```
bvaughn: The `_updateStatus` method does nothing but call `setState` and it's only called here. Just combine this with the below `setState(result)` call?

js/repl/Repl.js
```diff
@@ -447,6 +467,12 @@ class Repl extends React.Component {
     // This prevents frequent updates while a user is typing.
     this._compileToState(code);
   };
+
+  _updateStatus(status: Object) {
+    this.setState({
+      status: Object.assign({}, this.state.status, status),
```
bvaughn: This should use the [updater-function format of `setState`](https://reactjs.org/docs/react-component.html#setstate) to avoid potential stale values in `this.state.status`.

js/repl/Utils.js
```diff
@@ -0,0 +1,42 @@
+const sizes = ["Bytes", "kB", "MB", "GB", "TB", "PB", "EB"];
+export function prettySize(
+  size: number,
+  nospace: boolean,
+  one: boolean,
+  places: number
+) {
```
bvaughn: Tiny nit, but it looks like this (forked) implementation is more complex than is needed for this project, since we only ever call it with a single (`size`) argument.

Can we drop things like `nospace`?

js/repl/compile.js
```diff
@@ -75,9 +79,9 @@ export default function compile(code: string, config: CompileConfig): Return {
       presets: config.presets,
       sourceMap: config.sourceMap,
     });
-
     compiled = transformed.code;
-
+    meta.compiled.size = new Blob([compiled], { type: "text/plain" }).size;
+    meta.raw.size = new Blob([code], { type: "text/plain" }).size;
```
bvaughn: Kind of a nit, but we could also remove the `compiled` and `raw` wrapper objects here since we aren't bundling anything other than `size`.

If we inlined them more, we could also get rid of the unnecessary work being done by `_updateStatus` to massage their format before calling `setState` (and the `this.setState(result, setStateCallback)` would be sufficient)

package.json
bvaughn: This still needs to be reverted.

js/repl/Repl.js
bvaughn: Actually, based on my comment on `js/repl/Utils.js` (below) this logic might be able to go away entirely.

js/repl/CodeMirrorPanel.js
```diff
@@ -31,6 +39,9 @@ export default function CodeMirrorPanel(props: Props) {
           preserveScrollPosition={onChange == null}
           value={props.code}
         />
+        {options.fileSize && (
+          <div className={styles.fileSize}> Size: {fileSize} </div>
```
bvaughn: One final thought: I wonder if we could get rid of the "Size:" prefix entirely? Given the units (eg KB, MB) it seems kind of unnecessary.) This would make the overlay even smaller, which would be nice.

```js
<div className={styles.fileSize}>{fileSize} </div>
```

samundrak:
Thank you @bvaughn for feedback, I have changed pull request as per comments :+1:

js/repl/Repl.js
```diff
@@ -317,7 +329,11 @@ class Repl extends React.Component {
         sourceMap: runtimePolyfillState.isEnabled,
         useBuiltIns: state.builtIns,
       })
-      .then(result => this.setState(result, setStateCallback));
+      .then(result => {
+        result.meta.compiledSize = prettySize(result.meta.compiledSize);
+        result.meta.rawSize = prettySize(result.meta.rawSize);
```
samundrak: can do `{...result,  meta: {...result.meta, rawSize:0, compiledSize:0 }}` if some additional bytes won't matter by not mutating param ? :smile:

js/repl/compile.js
```diff
@@ -101,6 +98,7 @@ export default function compile(code: string, config: CompileConfig): Return {
       //   compiled = prettier.__debug.formatAST(transformed.ast, DEFAULT_PRETTIER_CONFIG);
       // } else {
       compiled = prettier.format(compiled, DEFAULT_PRETTIER_CONFIG);
+      meta.compiledSize = new Blob([compiled], { type: "text/plain" }).size;
```
bvaughn: Can't we only do the `meta.compiledSize` calculation once, by just moving it to after the if-prettier check?

bvaughn:
I dig it

bvaughn:
On mobile so having trouble viewing CI logs, but added a comment

existentialism:
Looks good, will land once CI finishes!

js/repl/Utils.js
```diff
@@ -0,0 +1,25 @@
+const sizes = ["Bytes", "kB", "MB", "GB", "TB", "PB", "EB"];
```
montogeek: It would be great to specify where was this code inspired from: https://github.com/davglass/prettysize
Something like:
`Forked from https://github.com/davglass/prettysize`

PR #1508, opened 2018-01-11
change the babel-plugin-macros example

kentcdodds:
Because people keep conflating babel-plugin-macros with preval and that turns people off from both tools because they think it means code in strings which is not true.

I'm actually going to make another PR with a different example and we can talk about which we like better. Standby...


kentcdodds:
Yeah, I think I like the other one better: https://github.com/babel/website/pull/1509

Andarist:
This one was great too, maybe you could incorporate both examples? The other one seems more focused on preval than macros - by including both examples you'd let people know that those 2 separate tools, it only happens that preval has a macro variant.

kentcdodds:
Sure. I'll update the blog post to have two examples.

PR #1509, opened 2018-01-11
change example to use preval.require

kentcdodds:
Because people think that `babel-plugin-macros` === `code in strings` && `preval` === `code in strings` and that turns people off to both tools.


hzoo:
K gave merge access

PR #1511, opened 2018-01-12
Fix mongodb svg image

hswolff:
I accidentally butchered the image and need to just update the svg content. Apologies :(


hswolff:
Actually hold one sec, just noticed I duped this.

hswolff:
Ok, that's what I meant originally. D'oh!

existentialism:
Oh dang... good catch!

PR #1518, opened 2018-01-21
Add community plugins

xtuc:
> test at https://deploy-preview-1518--babel-new.netlify.com/repl.html

This PR adds a "Plugins" section in the sidebar. You can search for plugins on npm and activate them.

Notes:
- some syntax plugins are not pushed to the parser (pipeline operator for example)
- the API is hosted on `now` for now, we could have it under `some-api.bundle.run` if needed.

Next steps:
- Store the plugins list into the url

Example:
kapture 2018-01-21 at 19 14 29


SaraVieira:
The code for the GraphQL server is here: https://github.com/SaraVieira/babel-plugins, it can be deployed anywhere you all prefer 🚀

Andarist:
Wow, this looks awesome!

xtuc:
@hzoo you mentioned keeping the `babel/plugin` prefix but it's taking so much space: 
screenshot from 2018-01-28 19-54-57

I think we should remove them.

hzoo:
Don't have to do it because most plugins are `babel-plugin` but some are not is all so there isn't a way to get them unless you introduce a new syntax to allow it?

xtuc:
Yes we use `babel-plugin` and `@babel/plugin` for the search on npm. Scanning the entire registry is possible but probably overkilled.

We can maybe indicate it somewhere, like:

screenshot from 2018-01-29 09-21-51

js/repl/ReplOptions.js
```diff
@@ -677,6 +749,11 @@ const styles = {
     maxWidth: "4rem",
     paddingLeft: "0.75rem",
   }),
+  pluginName: css({
+    width: "100%",
+    padding: "0.75rem",
+    marginBottom: "1rem",
```
nicolo-ribaudo: This two sizes might be need to be updated when this PR is included in the docusaurus site (I had to change some sizes in the repl - https://github.com/babel/website/pull/1533)

xtuc: Yeah, I need to change to branch of this PR against the new website.

xtuc:
Netlify finally managed to deploy the website, you can try it live here https://deploy-preview-1518--babel-new.netlify.com/repl.html.

I switched the search to Algolia with the help of @haroenv (thanks!).

@haroenv
I'm wondering why my plugin `babel-plugin-transform-scala-lambda` doesn't show up when you search `scala` in the REPL because on the [yarn website it does](https://yarnpkg.com/en/packages?q=babel-plugin%20scala).

Haroenv:
There's a bug still, I think the `onChange` of the checkbox and search input are switched or something. 



Otherwise, your plugin isn't yet shown, because we didn't do the reindex to add the tag on all packages (next Friday at the latest). 

One last thing, just so we don't forget, there needs to be a `<PoweredBy />` component still, you can use the one from React InstantSearch; or make your own link, as long as it says search or powered by Algolia and has a link to algolia.com

SaraVieira:
Hi @Haroenv,

I will add this tomorrow and also take a look at that :)

SaraVieira:
Hello !

I wasn't able to do this yesterday because the internet on the train was a lie but the logo has been added , let's wait for netifly 🤞

xtuc:
Thanks @SaraVieira!

What about something like:

- screenshot from 2018-02-08 08-44-16
- screenshot from 2018-02-08 08-44-01

SaraVieira:
But aligned right? 😝

Haroenv:
It can also be at the bottom of the results, providing that it’s not hidden behind a scrollbar

SaraVieira:
screen shot 2018-02-08 at 15 36 14


I also added some padding around the checkbox :)

cc @xtuc @Haroenv

Haroenv:
Bug I described earlier is still happening on Safari BTW

giuseppeg:
> Scanning the entire registry is possible but probably overkill

Would you be open to add keywords search too?

Haroenv:
@guiseppeg, not sure what you mean

xtuc:
I'm suspecting a bug in the way we merge the old keywords with the one inferred from the name (I can add a test for that later).

PR #1519, opened 2018-01-22
all the yearly presets are deprecated now.

graingert:


hzoo:
Sounds good, but I think I would just remove the entries in that list entirely?

graingert:
@hzoo removed.

PR #1520, opened 2018-01-22
Add Webpack usage

kroko:
Related to https://github.com/babel/babel/issues/7254

### DIFF
--- a/docs/usage/polyfill.md
+++ b/docs/usage/polyfill.md
@@ -24,7 +24,7 @@ npm install --save babel-polyfill
 
 > Because this is a polyfill (which will run before your source code), we need it to be a `dependency`, not a `devDependency`
 
-## Usage in Node / Browserify / Webpack
+## Usage in Node / Browserify / webpack
 
 To include the polyfill you need to require it at the top of the **entry point** to your application.
 
@@ -42,13 +42,13 @@ polyfills are loaded first:
 import "babel-polyfill";
 ```
 
-When using Webpack there are multiple ways to include Babel Polyfill. 
+When using webpack there are multiple ways to include Babel Polyfill. 
 
 * In case [`@babel/preset-env`](https://github.com/babel/babel/tree/master/packages/babel-preset-env) preset is used 
 
 	* If `useBuiltIns: 'usage'` is specified in `.babelrc` then do not include Babel Polyfill neither in `webpack.config.js` entry array nor source. Babel Polyfill still has to be installed though.
 	* If `useBuiltIns: 'entry'` is specified in `.babelrc` then include Babel Polyfill at the top of the entry point to your application via `require` or `import` as discussed above.
-	* If `useBuiltIns` key is not specified or it is explicitly set to false `useBuiltIns: false` in `.babelrc` then add Babel Polyfill to Weback entry array within `webpack.config.js`.
+  * If `useBuiltIns` key is not specified or it is explicitly set with `useBuiltIns: false` in your .babelrc, add Babel Polyfill to the entry array in your `webpack.config.js`.
 
 > babel-loader 7.x | babel 6.x
 
@@ -66,7 +66,7 @@ module.exports = {
 };
 ```
 
-* In case `@babel/preset-env` is not used then add Babel Polyfill to Weback entry array as discussed above. It can still be added at the top of the entry point to application via `require` or `import`, but this is not recommended.
+* In case `@babel/preset-env` is not used then add Babel Polyfill to webpack entry array as discussed above. It can still be added at the top of the entry point to application via `import` or `require`, but this is not recommended.
 
 ## Usage in Browser
 
--- a/docs/usage/polyfill.md
+++ b/docs/usage/polyfill.md
@@ -42,13 +42,31 @@ polyfills are loaded first:
 import "babel-polyfill";
 ```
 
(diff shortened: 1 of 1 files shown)


docs/usage/polyfill.md
```diff
@@ -42,14 +42,32 @@ polyfills are loaded first:
...
 
+> babel-loader 8.x | babel 7.x
+
+```js
+module.exports = {
+  entry: ["@babel/polyfill", "./app/js"]
+};
+```
+
+* In case `@babel/preset-env` is not used then add Babel Polyfill to Weback entry array as discussed above. It can still be added at the top of the entry point to application via `require` or `import`, but this is not recommended.
```
Andarist: I'd switch to "`import` or `import`" - let's put import always as first.

existentialism:
Nice work!

One thing we need to keep in mind that none of the docs are updated for v7 yet (we still need to support switching between versions, etc), so referencing both the scoped preset and `useBuiltIns: usage` might be premature?

docs/usage/polyfill.md
```diff
@@ -42,14 +42,32 @@ polyfills are loaded first:
 import "babel-polyfill";
 ```
 
-With `webpack.config.js`, add `babel-polyfill` to your entry array:
+When using Webpack there are multiple ways to include Babel Polyfill.
```
existentialism: Let's convert "Babel Polyfill" to `babel-polyfill` here and everywhere else.

docs/usage/polyfill.md
```diff
@@ -42,14 +42,32 @@ polyfills are loaded first:
...
 ```
 
-With `webpack.config.js`, add `babel-polyfill` to your entry array:
+When using Webpack there are multiple ways to include Babel Polyfill. 
+
+* In case [`@babel/preset-env`](https://github.com/babel/babel/tree/master/packages/babel-preset-env) preset is used 
+
+	* If `useBuiltIns: 'usage'` is specified in `.babelrc` then do not include Babel Polyfill neither in `webpack.config.js` entry array nor source. Babel Polyfill still has to be installed though.
+	* If `useBuiltIns: 'entry'` is specified in `.babelrc` then include Babel Polyfill at the top of the entry point to your application via `require` or `import` as discussed above.
+	* If `useBuiltIns` key is not specified or it is explicitly set to false `useBuiltIns: false` in `.babelrc` then add Babel Polyfill to Weback entry array within `webpack.config.js`.
```
existentialism: > then add Babel Polyfill to Weback entry

Webpack -> Webpack

docs/usage/polyfill.md
existentialism: Also, Weback -> Webpack

docs/usage/polyfill.md
existentialism: Also the false `useBuiltins: false` bit reads a bit weird, maybe:

> * If `useBuiltIns` key is not specified or it is explicitly set with `useBuiltIns: false` in your `.babelrc`, add `babel-polyfill` to the entry array in your `webpack.config.js`:

docs/usage/polyfill.md
Andarist: actually I remember that they prefer lower cased `webpack`

docs/usage/polyfill.md
kroko: Disagree.  

_Babel Polyfill_ is the name of the game.  
Once you change it to `babel-polifill` it looks like name of the package to install.  
Reader comes to the docs, sees `babel-polifill` and just hits `npm i babel-polifill` which in near future will be wrong thing to do.  Thus it would need yet another change again to `@babel/polyfill`.

If the name *Babel Polyfill* was left then we could avoid all that. The reader would just refer to https://babeljs.io/docs/usage/polyfill/#installation where the correct package name is noted. I suggest not mixing "name" with "current package name".

kroko:
I'm totally ok with changes introduced by @existentialism  
See my comments https://github.com/babel/website/pull/1520#discussion_r163548518 on using _Babel Polyfill_ though

kroko:
Can this be resolved and merged (accepting all @existentialism proposed changes)?

hzoo:
Deploy preview for *babel-new* failed.

Built with commit 03d84ac3229539d56b86e5c77a2aa1f26dbfd09d

https://app.netlify.com/sites/babel-new/deploys/5b14e32e11b73b6c9d21109e

docs/usage/polyfill.md
```diff
@@ -42,14 +42,32 @@ polyfills are loaded first:
 import "babel-polyfill";
 ```
 
-With `webpack.config.js`, add `babel-polyfill` to your entry array:
+When using webpack there are multiple ways to include Babel Polyfill. 
+
+* In case [`@babel/preset-env`](https://github.com/babel/babel/tree/master/packages/babel-preset-env) preset is used
```
ev1stensberg: Does `modules: false` have any relevance here to compiling successfully?

hzoo:
looks good, I don't think "Babel Polyfill" is consistent. We tend to refer to the package name itself `@babel/polyfill` instead of just those words. Especially since we are using scoped packages now and we don't want people to try to google it and use `babel-polyfill` instead. I would refer to Babel as the project and the package names for everything else?

hzoo:
Merged as part of https://github.com/babel/website/pull/1858

Issue #1522, opened 2018-01-25
New Site: Videos page


hzoo:
Part of https://github.com/babel/website/pull/1521
> aside: maybe we need a resources page which links to podcasts as well or we rename it to media? I guess we could do a similar thing for articles/blog posts/books/etc.

We need to port the [videos](https://babeljs.io/docs/community/videos/) page in the new site since the old one used jekyll (`{% for category in site.data.videos %}`) and bootstrap css

This would be targeting the [`docusaurus`](https://github.com/babel/website/tree/docusaurus) branch, not master.

> I would read through some of the docs: https://docusaurus.io/docs/en/installation.html + and check out the branch

- [ ]  we need to basically convert the [videos](https://github.com/babel/website/blob/docusaurus/videos.html) page into a react component and put it into [website/pages/en](https://github.com/babel/website/tree/docusaurus/website/pages/en) as `videos.js`
- [ ] should probably move the [data](https://github.com/babel/website/blob/docusaurus/_data/videos.yml) into it's own file or just in the video.js file itself (doesn't matter to me right now).
- [ ] will need to re-create/create styling for that page: current styling at https://github.com/babel/website/blob/docusaurus/_sass/components/_video.scss. Totally fine if you can improve on what's there as well so it doesn't "have" to look the same

- So move https://github.com/babel/website/blob/docusaurus/videos.html + https://github.com/babel/website/blob/docusaurus/_data/videos.yml into a `https://github.com/babel/website/blob/docusaurus/website/pages/en/videos.js`

cc @zsoltime since you made the page originally if you want to do this or we can ask someone new!

QC-L:
@hzoo Hi,  I would like to take up this issue, will be submitted as soon as possible.

hzoo:
Via https://github.com/babel/website/pull/1524

PR #1524, opened 2018-01-26
feat(videos): convert the videos page into react component

QC-L:
I have completed the following functions:

* convert the videos page into react component
* change _data/videos.yml -> website/data/videos.yml
* add a few styles

See #1522


QC-L:
https://deploy-preview-1524--babel.netlify.com/videos.html

abiduzz420:
Nice work @QC-L 
I have few suggestions which can be taken into account relating to styling:
- [ ] In the `CategoryInfo`, there is lot of empty space. We can either reduce the empty space or increase the font sizes. I prefer the latter option (Decreasing the bottom margin)
- [ ] Make the `VideoItem` component center aligned. They don't even out well on the right.
I like the idea of `text-muted` for links. Looks sweet!

hzoo:
Awesome, feel free to remove the corresponding old video files that I referenced in the issue

QC-L:
@hzoo It has been deleted.

hzoo:
Can you check https://babeljs.io/docs/community/videos/, it looks like it looks different when you resize (not center aligned on mobile) - I think that's the same point made ^

QC-L:
@hzoo OK, I'm going to amend it

QC-L:
@abiduzz420 OK, let me try.

QC-L:
@hzoo Please try again.

website/static/css/video.css
```diff
@@ -0,0 +1,94 @@
+.babel-video {
+  border: 1px solid rgba(0, 0, 0, .2);
+  border-radius: .25rem;
+  height: 100%;
+  overflow: hidden;
+}
+
+.page-header {
```
hzoo: One issue I'm seeing is that you have styles that could be re-used across the other pages like in https://github.com/babel/website/pull/1530 - I would think video.css would only apply to that page (or that docusarus isolates each css but that's not the case)

QC-L: I initially felt that the style would be adjusted, so the style was not added to babel.css. I'm going to amend it.

Issue #1528, opened 2018-01-26
New Site: Users page


hzoo:
> Current page is at https://babeljs.io/users/
> New site is http://localhost:3000/users.html if you run locally
> https://deploy-preview-1521--babel.netlify.com/users.html from the preview in https://github.com/babel/website/pull/1521

Looks like the styles are kind of weird, so this task is to fix those to be similar to the current version. Free free to make other changes if it seems more suitable though.

This would be targeting the [`docusaurus`](https://github.com/babel/website/tree/docusaurus) branch, not master.

> I would read through some of the docs: https://docusaurus.io/docs/en/installation.html + and check out the branch

- Again check `website/pages/en/users.js` and then modify the styles in static/css/users.css (make a new file)?

ajaymathur:
cool, can I please do it 🙂

hzoo:
Done by https://github.com/babel/website/pull/1529

PR #1529, opened 2018-01-27
added styles for user

ajaymathur:
updated the users styles.  https://github.com/babel/website/issues/1528

Addition:

added a hover effect, please drop review for that too. Also let me know if needs to reverted. 🙂

### DIFF
--- a/website/static/css/users.css
+++ b/website/static/css/users.css
@@ -35,6 +35,9 @@
   
   .mainContainer .showcaseSection .babel-user-link:hover {
     background: #fff;
+    -webkit-transform: scale3d(1.1, 1.1, 1);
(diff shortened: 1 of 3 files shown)


website/static/css/users.css
```diff
@@ -0,0 +1,52 @@
+.mainContainer .showcaseSection .logos {
```
hzoo: does it need to be nested like this?

ajaymathur: I can remove mainContainer but I think we should keep `.showcaseSection .logos` so that styles don't effect other pages.

hzoo:
I'll land as is and you/someone can make another pr

Issue #1523, opened 2018-01-25
New Site: Team page


hzoo:
Basically the same idea as in https://github.com/babel/website/issues/1522

We need to port the [team](https://babeljs.io/team) page in the new site since the old one used jekyll (`{% for category in site.data.team %}`) and bootstrap css

This would be targeting the [`docusaurus`](https://github.com/babel/website/tree/docusaurus) branch, not master.

> I would read through some of the docs: https://docusaurus.io/docs/en/installation.html + and check out the branch

- [ ]  we need to basically convert the [team](https://github.com/babel/website/blob/docusaurus/team.md) page + https://github.com/babel/website/blob/docusaurus/_includes/member.html into a react component and put it into [website/pages/en](https://github.com/babel/website/tree/docusaurus/website/pages/en) as `team.js`
- [ ] should probably move the [data](https://github.com/babel/website/blob/docusaurus/_data/team.yml) into a different file or just in the team.js file itself (doesn't matter to me right now).
- [ ] will need to re-create/create styling for that page: current styling at https://github.com/babel/website/blob/docusaurus/_sass/pages/_team.scss. Totally fine if you can improve on what's there as well so it doesn't "have" to look the same

cc @rajzshkr since you had worked on this originally but open to other help

EDIT: would be nice to include https://github.com/babel/website/issues/1457 too (an area for each team member to say something)

abiduzz420:
Hi @hzoo This is a wonderful opportunity for contribution to babel ecosystem. I would like to take up this issue. Right now, just figuring out different things in the new branch Docusaurus.

hzoo:
Closed by https://github.com/babel/website/pull/1530

Issue #1457, opened 2017-11-07
Updates to the team page


hzoo:
https://babeljs.io/team

I think we continue to improve on this page, and take from some things from [vue](https://vuejs.org/v2/guide/team.html).

We don't have translated docs yet (we should figure that out too), and also versioned docs 😛 
but I do like the links to location/language?

aside: I would like to add a blurb from everyone who wants to say something - like how they got involved, etc

hzoo:
Also add @angus-c and @jdan as part of this

PR #1530, opened 2018-01-27
Team page designed for the new website

abiduzz420:
##### Changes Made:
- Added `team.js` in `website/pages/en/`
- Added `team.css` in `website/static/css/`
- In `team.yml` the keys are changed from kebab case to camel case
- Moved `_data/team.yml` -> `website/data/team.yml`
- Moved `icons` directory: `_includes/icons` -> `website/static/img/icons`
- Included new team members

This PR solves issue #1523 

_Please let me know if I need to make any further modifications or changes._


abiduzz420:
Changes can be viewed here: https://deploy-preview-1530--babel.netlify.com/team.html

abiduzz420:
@hzoo I have taken note of the issue #1457. I like the idea of contributors sharing their first PR stories and memorable contributions to babel.
The current PR does not included these changes. I shall work on the mockups and present you some designs, by taking inspiration from other testimonial websites. 

What would be the upper limit of number of characters in the message ?

website/static/css/team.css
```diff
@@ -0,0 +1,137 @@
+.page_header {
```
hzoo: Looks like there are similar styles from video.css https://github.com/babel/website/pull/1524/files#r164283741

maybe should move to another css like 'page' or to the babel.css file? Not sure the best way

website/static/css/team.css
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/* 
+  ##Device = Desktops
+  ##Screen = 1281px to higher resolution desktops
+*/
+
+@media (min-width: 1281px) {
+  .team_member {
+    width: 25%;
```
hzoo: Does flexbox itself not cover this kind of thing already?

website/static/css/team.css
abiduzz420: Yeah. We can have it in babel.css file. I feel we should create `Header` component, and re-use them on both videos and team page.

website/static/css/team.css
abiduzz420: Yes. You can achieve the same thing using flex-box as well. I have used media queries to set specific styles based on the screen width. Flexbox helps in layout settings without knowing explicit width, height values. Also, we can change flex layouts within media queries as well. Shall I do do the same using flexbox ?

abiduzz420: To be a bit more clear, flex box has more to do with how space is being used and media queries tells how content is being viewed.

rajzshkr:
screen shot 2018-01-28 at 8 12 34 pm

Can you check this?

abiduzz420:
Thanks @rajzshkr for looking into it. I didn't catch that. I have fixed the issue now. Please let me if we need to modify anything further :)

website/static/css/team.css
abiduzz420: I have shifted common styles from video.css & team.css to babel.css

hzoo:
I think we should can probably just move to `emotion` to get around the naming issues like we are doing in the repl?

abiduzz420:
Yes I agree @hzoo . emotion would be the right choice.

Issue #1525, opened 2018-01-26
New Site: REPL page


hzoo:
Part of https://github.com/babel/website/pull/1521

> Current REPL is at: https://babeljs.io/repl

TL;DR is that we need to load the REPL page in docusaurus somehow.
Now since the repl is just a react component, it seemed like we could just add it to `website/pages/en` but because the current implementation using `codemirror` uses browser specific code like `navigator` etc it won't work when trying to render via docusarus on the server side.

Options: use a different module like react-codemirror (too much work), only load that part of the code on component mount (might be too much work), or just load the bundled version into a separate html page (like the current repl does).

Current JS code for it is at https://github.com/babel/website/tree/master/js/repl
HTML is at https://github.com/babel/website/blob/master/repl.html
webpack file is https://github.com/babel/website/blob/master/webpack.config.js
makefile at https://github.com/babel/website/blob/master/Makefile (might want to keep the webpack commands)

It's simple enough to convert these to script tags:

https://github.com/babel/website/blob/master/repl.html#L6-L20

For this part:

https://github.com/babel/website/blob/50e09bdceeee658a621ce9a85ddc7ca4fa15c38f/repl.html#L21-L22

We'll have to look at https://github.com/babel/website/blob/50e09bdceeee658a621ce9a85ddc7ca4fa15c38f/_includes/scripts.html#L35-L53

nicolo-ribaudo:
I'd like to work on this (it would be my first PR to the website I think)

hzoo:
Ooh ok, let me know if you need context on the repl still - or just message in our slack

hzoo:
Closed by https://github.com/babel/website/pull/1533

PR #1533, opened 2018-01-28
Docusaurus repl

nicolo-ribaudo:
The options bar doesn't look exactly like in the old repl, because the base font size changed (it was 10px in the old site, while it is 16px in the new one).
I think that less compact it looks nicer, but if you want I can scale it down like it was.

I'd like to prerender the repl component with docusaurus, but since it will require some more work I opened this PR so that we have a working implementation in the meantime.

Fixes https://github.com/babel/website/issues/1525

---

https://deploy-preview-1533--babel.netlify.com/repl.html


website/pages/en/repl.js
```diff
@@ -0,0 +1,44 @@
+"use strict";
+
+const React = require("react");
+
+const styles = [
+  "https://unpkg.com/codemirror@5.30.0/lib/codemirror.css",
+];
+
+const scripts = [
```
nicolo-ribaudo: Do you know if there is a way to define some scripts in `siteConfig.js#scripts` which are only included in this page, instead of doing like this?

js/repl/Repl.js
```diff
@@ -553,12 +553,13 @@ const styles = {
     flex: "0 0 auto",
   }),
   repl: css({
-    height: "100%",
+    height: "calc(100vh - 50px)", // 50px is the header's height
```
xtuc: I'm not sure if `calc` has a good browser support?

rajzshkr: From IE11+ calc is supported.

nicolo-ribaudo: Is there a way with the emotion library to do something like this?

```css
.repl {
  heigh: 100%;
  height: calc(100vh - 50px);
}
```

xtuc: You mean having two times the `height` attribute? Like a fallback in case the `calc` doesn't work?

We can try to ping @tkh44.

nicolo-ribaudo: > You mean having two times the height attribute? Like a fallback in case the calc doesn't work?

Yes!
Anyway, later I will try to make it work without using `calc`.

tkh44: Seems to work fine 😄

xtuc: Perfect, thanks @tkh44! We need to remove the object and use the template string instead.

hzoo:
Awesome work @nicolo-ribaudo !!

PR #1539, opened 2018-01-29
Amathur/user follow up

ajaymathur:
follow up to : https://github.com/babel/website/pull/1529

removed nested styles where not needed.

### DIFF
--- a/website/static/css/users.css
+++ b/website/static/css/users.css
@@ -1,27 +1,27 @@
-.mainContainer .showcaseSection .logos {
+.showcaseSection .logos {
     background: #eee;
     border-radius: 5px;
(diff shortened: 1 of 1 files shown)


website/static/css/users.css
```diff
@@ -1,52 +1,52 @@
...
-  .mainContainer .showcaseSection .babel-user-link {
-    display: block;
-    position: relative;
-    height: 120px;
-    padding: 10px;
-    margin: 15px 0;
-    background: #fff;
-    border-radius: 5px;
-  }
+.mainContainer .showcaseSection .babel-user-link {
```
ajaymathur: need main container here as default styles are more specific than this.

Issue #1526, opened 2018-01-26
New Site: 404 page


hzoo:
Part of https://github.com/babel/website/pull/1521

> Should be simple enough to just make a 404.js in `website/pages/en`
> Ref https://github.com/facebook/Docusaurus/issues/301

This would be targeting the [`docusaurus`](https://github.com/babel/website/tree/docusaurus) branch, not master.

> I would read through some of the docs: https://docusaurus.io/docs/en/installation.html + and check out the branch

Might be fun to make it show the ascii guy fieri or something but feel free to just port https://github.com/babel/website/blob/master/404.html for now and we can move forward

GregBorrelly:
Hello, I would like to take over this issue.

hzoo:
@GregBorrelly still planning on tackling this?

GregBorrelly:
@hzoo pushing a PR request tonight. Got caught up with work on the weekend.

GregBorrelly:
@hzoo so far I created the 404.js page with the following code in it 
`const React = require('react');


const ErrorPage = props => {

    return (
        <div className="container">
            <div className="row">
                <div className="col-md-8 col-md-push-2">
                    <script type="text/javascript">
                        var GOOG_FIXURL_LANG = 'en';
                        var GOOG_FIXURL_SITE = 'https://babeljs.io/'
                     </script>
                    <script type="text/javascript" src="https://linkhelp.clients.google.com/tbproxy/lh/wm/fixurl.js"></script>
                </div>
            </div>
        </div>
    )
}`

How can I check out how this page looks?  Where is the routing implemented in docusaurus ?

GregBorrelly:
@hzoo  yes I'm.  I created a 404.js page, but I'm having trouble pushing.  Should I fork the repo before creating a new branch?  I apologize for the noob questions, first big open source project I contribute too.

GregBorrelly:
https://github.com/babel/website/pull/1540

PR #1540, opened 2018-01-30
Adding 404 page

GregBorrelly:
Fixes https://github.com/babel/website/issues/1526

So not sure if I'm missing a step. I created a class base component that renders the HTML as it was on the legacy page.   Do we need a class base component? or is it ok if its pure;y functional? 

I haven't hooked up the component into anything, need your feedback to what the next steps should be.
``

### DIFF
--- a/website/pages/en/404.js
+++ b/website/pages/en/404.js
@@ -1,27 +1,18 @@
 const React = require("react");
 
-
 class ErrorPage extends React.Component {
-    render() {
-        return (
-
-            <div className="error-page">
-                <div className="error-message">
-                    <div className=" error-message-container container">
-                        <span>404 </span>
-                        <p>Page Not Found.</p>
-                    </div>
-
-                </div>
-
-
-         
-
-            </div>
-
-        );
-    }
+  render() {
+    return (
+      <div className="error-page">
+        <div className="error-message">
+          <div className=" error-message-container container">
+            <span>404 </span>
+            <p>Page Not Found.</p>
+          </div>
+        </div>
+      </div>
+    );
+  }
 }
-
 
 module.exports = ErrorPage;
--- a/website/pages/en/404.js
+++ b/website/pages/en/404.js
@@ -1,0 +1,23 @@
+const React = require("react");
+
+
+class ErrorPage extends React.Component {
+    render() {
+        return (
+            <div className="container">
+                <div className="row">
+                    <div className="col-md-8 col-md-push-2">
+                        <script type="text/javascript">
+                            var GOOG_FIXURL_LANG = 'en';
+                             var GOOG_FIXURL_SITE = 'https://babeljs.io/'
+                        </script>
+                        <script type="text/javascript" src="https://linkhelp.clients.google.com/tbproxy/lh/wm/fixurl.js"></script>
+                    </div>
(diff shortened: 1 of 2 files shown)


hzoo:
Not sure what fun thing we should do for the 404 page - maybe add play https://medium.com/@angustweets/hallelujah-in-praise-of-babel-977020010fad haha

GregBorrelly:
so I have the live server running but when I type a url that's not there for example: http://localhost:3000/not-here I get `Cannot GET /not-here/index.html`

How does docusaurus know to respond with my 404 page? 

I do a lot of CSS animations at work. I can have an animation that on click expands to display the "praise to babel" I think that would be pretty cool, no?

hzoo:
Oh ok I think the 404 is specifically from netlify itself :stuck:. I would just rename it to be a regular filename or directly check http://localhost:3000/404

GregBorrelly:
makes sense,  I'll rename it, style it, and then name it back to 404.js

what did you think about the animation idea for the 404 page? too much?

GregBorrelly:
Just added the styling to match the previous page

website/pages/en/404.js
```diff
@@ -0,0 +1,36 @@
...
+
+
+class ErrorPage extends React.Component {
+    render() {
+        return (
+
+            <div className="error-page">
+                <div className="container">
+                    <div className="row">
+                        <div className="col-md-12 error-message">
```
hzoo: Some of these styles won't mean anything anymore like `col-md-12` because we don't have bootstrap anymore

GregBorrelly: I'll remove them, are we using another framework?

hzoo: It's just whatever docusaurus provides and our own styles - The black heading was what we had before but I don't think we need that

GregBorrelly: roger that, I want to work in something eye-catching as you mentioned before. I removed the bootstrap classes

hzoo:
Ok I merged `docusaurus` into the `master` branch so you'll need to switch the target branch now (you can do this in github ui at the top)

GregBorrelly:
Ok, I just switched branches to master.  What other tasks need to be done? or should I continue to expand on this 404 page?

GregBorrelly:
@hzoo  IBM has a nice 404 page message
"Oops — that's not right!
Sorry, unless you really were hoping to find our 404 message, the page you were looking
for has been moved or deleted"

https://www.ibm.com/;www-01.ibm.com/events/wwe/indexconf/indexconf18.nsf/Registration.xsp?open

hzoo:
Yeah now I'm thinking we should just have something simple, the google thing might not make that much sense to me given we have a search bar anyways

hzoo:
If we can just have the footer at the bottom, and the contents taking up the whole space of the screen, and then a message like "not found" kind of thing, then we can update that it later

GregBorrelly:
Sure, I will set it up

GregBorrelly:
@hzoo made the update, there seem to be no conflicts but the netlify build failed.

hzoo:
All good, I updated the build steps so it would fail if you didn't rebase. I'l push it up

Issue #1527, opened 2018-01-26
New Site: Setup/Tools page


hzoo:
Basically the same idea as in https://github.com/babel/website/issues/1522

> This one might be a little more involved

We need to port the [setup](https://babeljs.io/docs/setup) page in the new site since the old one used jekyll (`{% for category in site.data.tools %}`) and bootstrap css

This would be targeting the [`docusaurus`](https://github.com/babel/website/tree/docusaurus) branch, not master.

> I would read through some of the docs: https://docusaurus.io/docs/en/installation.html + and check out the branch

We can probably just redo how this works? Not sure the time investment though.

---

If we port:

- [ ]  we need to basically convert the setup page - https://github.com/babel/website/blob/docusaurus/docs/setup.md and https://github.com/babel/website/blob/docusaurus/_includes/tools/items.md
- [ ] should probably move the [data](https://github.com/babel/website/blob/docusaurus/_data/tools.yml) into a different file or just in the tools.js file itself (doesn't matter to me right now). Same with https://github.com/babel/website/blob/docusaurus/_includes/tools
- [ ] will need to re-create/create styling for that page: current styling at https://github.com/babel/website/blob/docusaurus/_sass/pages/_tools.scss. Totally fine if you can improve on what's there as well so it doesn't "have" to look the same

QC-L:
Let me try, but I didn't think about how to design it. I realized it first.

QC-L:
@hzoo How do you feel like that?
![](http://oypz34fc0.bkt.clouddn.com/babel-setup.png)

xtuc:
@QC-L that looks good. I'm not sure about the yellow background on the buttons. Maybe could we change it to white?

QC-L:
White is the original effect. How do we make some changes?

hzoo:
I think it's all good now, we'll probably want to redo this in the future anyway (a lot of the packages aren't really updated or maintained by us either)

QC-L:
@hzoo I found that docusaurus doesn't seem to support `onClick`

xtuc:
Ok, fine to me

> @hzoo I found that docusaurus doesn't seem to support onClick

Outch, we could maybe ask @JoelMarcey

QC-L:
Hey, @JoelMarcey. I try to add the `onClick` event in the `<a>` tag, but didn't take effect. Is Docusaurus not supported? I searched through all the documents and found nothing.

JoelMarcey:
[...]

PR #1541, opened 2018-01-30
feat(tools): complete setup page

QC-L:
Edit by hzoo: Ref https://github.com/babel/website/issues/1527

Error: 
* `onClick` no response


QC-L:
https://deploy-preview-1541--babel.netlify.com/en/setup.html

hzoo:
Cool! can also remove https://github.com/babel/website/blob/docusaurus/_data/tools.yml later

QC-L:
@hzoo There are still some bugs to be changed and later changes

hzoo:
Ok I merged `docusaurus` into the `master` branch so you'll need to switch the target branch now (you can do this in github ui at the top)

website/pages/en/setup.js
```diff
@@ -0,0 +1,80 @@
...
+  render() {
+    const tools = siteConfig.tools;
+    const showCase = tools.map((types, i) => {
+      const showTools = Object.keys(types.items).map((tool, j) => {
+        const isActive = false;
+        return (
+          <a
+            key={j}
+            href={`#${tool}`}
+            className={isActive ? "tools-button active" : "tools-button"}
```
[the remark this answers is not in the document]
QC-L: I've changed it.

QC-L:
@hzoo I complete, but it can't be previewed.

hzoo:
Awesome! nice work on moving over everything - using jquery is ok given we probably want to redo how this works later anyways but this will bring us back to what we had

website/pages/en/setup.js
```diff
@@ -0,0 +1,110 @@
...
+const setupBabelrc = siteConfig.setupBabelrc;
+const toolsMD = siteConfig.toolsMD;
+
+const StepFour = () => {
+  return (
+    <div className="step-hidden step-setup">
+      <h2>
+        <span className="step-no">4</span> Create <code>.babelrc</code> configuration file
+      </h2>
+      <MarkdownBlock>{setupBabelrc}</MarkdownBlock>
```
hzoo: I guess we'd want to translate the setup too? Not sure how this will work with https://docusaurus.io/docs/en/translation.html#enabling-translations-on-pages unless we wrap in translate but it's also markdown

hzoo: Or maybe we should make individual pages that are translatable and leave this as english?

website/static/scripts/tools.js
```diff
@@ -20,14 +22,16 @@
     $currentNav = $(this);
     $currentNav.addClass("active");
 
-    var name = $currentNav.attr("data-name");
+    var name = $currentNav.attr("data-title");
     location.hash = name;
 
-    $currentItem = $("[data-name=" + name + "]:not(.btn)").show();
+    $currentItem = $("[data-title=" + name + "]:not(.tools-button)").show();
+    console.log($currentItem);
```
hzoo: unnecessary now?

website/static/scripts/tools.js
```diff
@@ -2,15 +2,17 @@
...
   function reset() {
     if ($currentItem) $currentItem.hide();
     if ($currentNav) $currentNav.removeClass("active");
     $currentItem = $currentNav = null;
     $steps.removeAttr("style");
+    $stepHidden.removeAttr("style");
   }
 
-  $(".btn-group .btn").click(function () {
+  $(".tools-group .tools-button").click(function () {
```
hzoo: It should be easy to convert these to non-jquery so we don't have to load it right?

QC-L:
I want to submit quickly, the code quality may be worse, follow-up I will optimize it.

hzoo:
yep +1 me too

JoelMarcey:
> Going to land for now but we should figure out the setup page is going to be translated (or if we should do something else about that) @JoelMarcey

Hi @hzoo - Since the setup page is a `.js` file within `pages`, it can be translated.

https://docusaurus.io/docs/en/translation.html#enabling-translations-on-pages

Is that what you were asking?

PR #1543, opened 2018-01-30
First pass at adding plugin/preset docs

existentialism:
There are a number of improvements we can/will make here... but this gets us up and running for now.

* We might want to improve how we remove the initial heading/description from the readme? Might not be a huge deal if we're strict about how standardized they are... but still seems brittle.

* Not ideal since it commits the readmes, but I don't think docusaurus will let us do what we were doing in jekyll here (maybe we can request?). Could also use custom pages and `CompLibrary.MarkdownBlock`, but then translations might be more difficult.

* The plan is to get things generally decent for v6, and then use docusaurus' versioning to start getting things updated to v7.


existentialism:
@JoelMarcey would def love your input on this! Our current website has [a script](https://github.com/babel/website/blob/master/scripts/download-readmes.js) that grabs all the READMEs from the separate packages in babel/babel and then we have matching jekyll pages that import them ([example](https://raw.githubusercontent.com/babel/website/master/docs/plugins/transform-es2015-destructuring.md)). Ideally we'd be able to not commit the README content but still take advantage of docusaurus' translation support.

JoelMarcey:
@existentialism Hi! So, I am thinking you could still run the script and get to where you want to be. 

I want to understand the flow a little bit better...

As I see it, the script grabs the `.md` files for the README. Now, if `.html` files could be generated via the script instead, you could place those files as static assets - https://docusaurus.io/docs/en/custom-pages.html

If you are just getting the `.md` files, it might be a little trickier. But could you add the `.md` files to the `docs` directory and programmatically inject each file into the `sidebars.json` file so that they show up in the documentation sidebar?

hzoo:
If we create html files, - isn't the translation method different if it's a doc vs a custom page?

I guess this is an issue with wanting to keep the documentation in the readme of each npm package itself vs. moving it to the docs page but wanting the translations to be easily done. We need to add the `id/title` fields for docusaurus (or jekyll), but I guess we might not want those in the readme of the packages themselves (I might be ok with this).

existentialism:
@JoelMarcey 👋 

Yeah, this PR does what the latter, it currently pulls down the README files and copies them into `docs`, adds id and sidebar_label front matter, etc.

One kind-of-nice thing our old jekyll setup let us do, was avoid committing all the individual README changes since the build script would import them from the repos, and include them at build time.

We _could_ likely accomplish this with HTML and the `CompLibrary.MarkdownBlock` component, but as @hzoo mentioned, would how would translations work with this?

hzoo:
It's like impossible to see the diff since I deleted the extra folder we committed.

We will just iterate more but for now, I modified https://github.com/babel/website/blob/f8b4ef3f08d9d964b673916cca9f02ffd3def061/scripts/download-readmes.js#L89
so that it actually renames the files to `docs/z_filename.md` and then in gitignore ignored those files

https://github.com/babel/website/blob/f8b4ef3f08d9d964b673916cca9f02ffd3def061/.gitignore#L28

PR #1552, opened 2018-02-03
Fix snippets cut off in mobile

### LINKED ISSUE #1549: [new site] Code snippets not scrollable on mobile
The snippets are cut off on mobile. And I'm not able to scroll them.

smartselectimage_2018-02-02-06-55-39

I'd like to fix this if that's ok. 😄

maaz93:
Fix #1549 

The code snippets were cut off on small screens. This is not a bug with Docusaurus, since removing the custom CSS for `.hljs` at https://github.com/babel/website/blob/3ec8a278bb9416bd1ebf604d6b9fc96fc1f77346/website/static/css/code-blocks.css#L2 and https://github.com/babel/website/blob/3ec8a278bb9416bd1ebf604d6b9fc96fc1f77346/website/static/css/code-blocks.css#L13 brings in a scroll bar. I believe the reason for the snippets being cut off is `.h1js::before`, which is unique to the babel website. Also, other sites like [jest](https://facebook.github.io/jest/) use Docusaurus and they automatically get scroll bars.

Because we want a custom header like thing for the snippets, we'll have to make do with wrapped text and won't be able use a scroll bar. So, just adding a way to handle `white-space`.

Some screenshots*:

## Before
js_before
shell_before

## After
js_after
shell_after

_*The screenshots were taken using the Chrome devtools by simulating mobile view using Responsive - Mobile S_

### DIFF
--- a/website/static/css/code-blocks.css
+++ b/website/static/css/code-blocks.css
@@ -4,6 +4,7 @@
   padding: 1em;
   position: relative;
   overflow-x: visible;
+  white-space: pre-wrap;
   border-left: 0;
   /* base 20px + header height */
   margin: 54px 0 20px;


hzoo:
Nice! also realized the preview's aren't working so will fix

SavePointSam:
Nice Catch!

PR #1559, opened 2018-02-05
fix(setup): rm jquery, use javascript

QC-L:
I completed the following functions:

* component optimization
* fix code snippets style
* rm jquery, use javascript
* add translate

See #1527
@hzoo Please review.

### DIFF
--- a/website/static/scripts/tools.js
+++ b/website/static/scripts/tools.js
@@ -1,36 +1,58 @@
 (function() {
-  var $currentItem;
-  var $currentNav;
-  var $steps = $(".step");
-  var $stepHidden = $(".step-hidden");
+  let currentItem;
+  let currentNav;
+  const stepHidden = document.getElementsByClassName("step-hidden");
 
-  function reset() {
-    if ($currentItem) $currentItem.hide();
-    if ($currentNav) $currentNav.removeClass("active");
-    $currentItem = $currentNav = null;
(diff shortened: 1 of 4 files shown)


website/static/scripts/tools.js
```diff
@@ -1,36 +1,58 @@
...
+    }
+    currentItem = currentNav = null;
+    for (let i = 0; i < stepHidden.length; i++) {
+      stepHidden[i].removeAttribute("style");
+    }
+  }
 
-  if (location.hash && location.hash !== "#") {
-    $(".tools-button[data-title=" + location.hash.slice(1) + "]").click();
+  const buttons = document.querySelectorAll(".tools-group .tools-button");
```
xtuc: jQuery looks simpler for that part. Why do you want to remove jQuery on our website?

QC-L: Here's jQuery import by me. Extra request too much trouble.

hzoo: If you think it's ok to keep jquery since it's just our docs page that's fine too, but I suggested we could remove. From a code maintenance perspective I guess it is easier

QC-L:
This page finished.

website/static/scripts/tools.js
QC-L: What do you think should be done here?

Issue #1538, opened 2018-01-29
New site: Main Page


hzoo:
I guess we should just port what we currently have?

> but we'll also probably want to redo this page later/soon (or just move the contents into an about page)

Similar to https://github.com/babel/website/issues/1522 and https://github.com/babel/website/issues/1523

---
This would be targeting the [`docusaurus`](https://github.com/babel/website/tree/docusaurus) branch, not master.

> I would read through some of the docs: https://docusaurus.io/docs/en/installation.html + and check out the branch

- [x] Convert https://github.com/babel/website/blob/master/index.md into js and put it into https://github.com/babel/website/blob/docusaurus/website/pages/en/index.js (You can probably throw away most of what is in that file since it's just a placeholder). Remember the main site is at https://babeljs.io, the mini repl code is at https://github.com/babel/website/blob/master/scripts/index.js
- [x] old styling code might be in https://github.com/babel/website/blob/docusaurus/_sass/, can delete unused code later
- [x] if it's difficult to port I would skip the mini repl at first and we can do that in a new pr https://github.com/babel/website/pull/1564
- [x] Would like to get some of our open collective sponsors up there at least the gold sponsors like on vuejs.org https://github.com/babel/website/pull/1564

hzoo:
Ok I just removed the old text contents and moved them into a new index/about page.

microbouji:
Hi, I'd like to work on this.

hzoo:
Mostly done in https://github.com/babel/website/pull/1564 so closing!

PR #1564, opened 2018-02-10
Add minirepl and sponsors to main page

microbouji:
This is for #1538. Changes overview:

- Minirepl
  - The same old minirepl code but now linted and bundled together with lodash.debounce ( if we ran webpack with -p it would go from ~40k to ~10k)
  - babel-standalone and ace editor still loaded separately through script tags
  - minirepl code fired on window load, after algolia's docsearch has had a chance to load zepto. I saw a discussion on other issues on whether to keep jquery or not. As long as we have search enabled on every page it means we also have zepto, so might as well use it. Probably not a reliable loading strategy though, so let me know if I should change it to vanilla js.

- Welcome & Donate sections
Instead of an "as easy as 1, 2, 3" usage demo, I thought "optimizing" for potential contributors and donors would be more helpful here. We should make a splendid getting started page to accommodate new folks.
Also, let me know about wording - couldn't think of anything better than "Friends of Open Source" for example.


hzoo:
Awesome start! I moved the current info into the main docs page https://5a7f5ad00b79b773ce8e38f5--babel-new.netlify.com/docs/en/index.html and we also have a setup page (although will probably be re-done later as well)

hzoo:
I think the special sponsors and probably just be the platinum ones, and maybe we can put the people each company is sponsoring (me for behance and @jridgewell on amp)?

website/pages/en/index.js
```diff
@@ -59,62 +49,222 @@ const PromoSection = props => (
...
+  return (
+    <div className="container paddingTop paddingBottom">
+      <div className="wrapper productShowcaseSection">
+        <div className="support-the-team">
+          <h2>Support the team</h2>
+          <p>
+            Babel is helping shape the future of the JavaScript language itself,
+            being used at companies like Facebook, Google, Netflix, and{" "}
+            <a href={pageUrl("users.html", language)}>hundreds more</a>. Your
+            donation will help cover expenses like attending TC39 ( the
```
hzoo: remove the `( the` space here, and `meettings` is `meetings`.

Yeah don't know how many things we should put here since I can think of more like attending/speaking at events/meetups, doing contributor workshops. Or we should probably make a new page that explains more of this actually..

website/pages/en/index.js
```diff
@@ -59,62 +49,222 @@ const PromoSection = props => (
...
+  );
+};
+
+const OpenCollectiveSponsors = props => {
+  const language = props.language || "en";
+  return (
+    <div className="container paddingTop paddingBottom">
+      <div className="wrapper productShowcaseSection">
+        <div className="support-the-team">
+          <h2>Support the team</h2>
```
hzoo: Team capitalized?

Makefile
```diff
@@ -2,6 +2,7 @@
 
 bootstrap:
 	./scripts/download-readmes.js
+	node ./scripts/download-sponsors.js
```
hzoo: I think you can just put `#!/usr/bin/env node` at the top like in the file above and you don't need the `node` part

azz: Yep + `chmod +x scripts/download-sponsors.js`

hzoo:

Built with commit df699a720cc967ff47971d22abe70fc69b827ad1

https://deploy-preview-1564--babel-new.netlify.com

website/pages/en/index.js
microbouji: For now I think a separate page would just become a duplicate of the opencollective page the donate button links to. Unless the new page will also have paypal or other ways to donate?

On the current homepage section I could use some help in writing a better summary to highlight the top expenses.

hzoo:
Awesome!! We can continue to iterate but this is a great start, thanks again and congrats on the first PR!

PR #1483, opened 2017-12-14
Upgrade to babel 7, and include Prettier optimization

azz:
Closes #1312

At the moment the build fails with 

```
TypeError: Property declaration of ExportNamedDeclaration expected node to be of a type ["Declaration"] but instead got "ExpressionStatement"
```

Maybe related to https://github.com/facebook/regenerator/pull/319?

<details>
<summary>Build Error</summary>

```
ERROR in ./js/repl/loadBabel.js
Module build failed: TypeError: Property declaration of ExportNamedDeclaration expected node to be of a type ["Declaration"] but instead got "ExpressionStatement"
    at Object.validate (/Users/azz/code/babel-website/node_modules/@babel/types/lib/definitions/utils.js:88:13)
    at Object.validate (/Users/azz/code/babel-website/node_modules/@babel/types/lib/validators/validate.js:15:9)
    at NodePath._replaceWith (/Users/azz/code/babel-website/node_modules/@babel/traverse/lib/path/replacement.js:169:7)
    at NodePath.replaceWith (/Users/azz/code/babel-website/node_modules/@babel/traverse/lib/path/replacement.js:153:8)
    at Object.replaceWithOrRemove (/Users/azz/code/babel-website/node_modules/regenerator-transform/lib/util.js:49:10)
    at PluginPass.<anonymous> (/Users/azz/code/babel-website/node_modules/regenerator-transform/lib/visit.js:156:16)
    at PluginPass.<anonymous> (/Users/azz/code/babel-website/node_modules/regenerator-transform/lib/util.js:27:17)
    at newFn (/Users/azz/code/babel-website/node_modules/@babel/traverse/lib/visitors.js:223:21)
    at NodePath._call (/Users/azz/code/babel-website/node_modules/@babel/traverse/lib/path/context.js:64:19)
    at NodePath.call (/Users/azz/code/babel-website/node_modules/@babel/traverse/lib/path/context.js:38:17)
 @ ./js/repl/Repl.js 26:40-62
 @ ./js/repl/index.js
 @ multi (webpack)-dev-server/client?http://localhost:34512 webpack/hot/dev-server ./js/repl/index.js
```
</details>

Any ideas?

PR #1600, opened 2018-03-08
[WIP] Initial Sandpack Proof of Concept

CompuIves:
This is my initial POC for Sandpack, it's reaaaally rough right now because I don't have more time today. One thing I'd like to do is wrap the existing CodeMirror panels in `SandpackConsumer`s, because those are already styled perfectly. I'm also thinking of exposing the `Error` field on the Sandpack state.

### DIFF
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -5,6 +5,7 @@ import "regenerator-runtime/runtime";
 import { css } from "emotion";
 import debounce from "lodash.debounce";
 import React from "react";
+import {SandpackProvider, CodeMirror, TranspiledCodeView} from 'react-smooshpack';
 import { prettySize } from "./Utils";
 import ErrorBoundary from "./ErrorBoundary";
 import CodeMirrorPanel from "./CodeMirrorPanel";
@@ -34,6 +35,7 @@ import {
 import WorkerApi from "./WorkerApi";
 import scopedEval from "./scopedEval";
 import { colors, media } from "./styles";
+
 
 import type {
   BabelPresets,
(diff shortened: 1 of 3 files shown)


hzoo:
Deploy preview for *babel-new* failed.

Built with commit 0aba0222b68ce2c1eb55c91742bb53fb75674ed7

https://app.netlify.com/sites/babel-new/deploys/5aa1b5c5fd0efa49cf68403c

CompuIves:
Ah I went ahead and started to reuse the existing components, was simpler than I thought!

hzoo:
Dang it feels super fast!?

xtuc:
> Dang it feels super fast!?

wow it's really super fast, Ives :heart_eyes:

js/repl/Repl.js
```diff
@@ -243,16 +255,25 @@ class Repl extends React.Component {
...
+                  && sandpack.managerState.transpiledModules['/index.js:'].source.compiledCode}
+                errorMessage={state.evalErrorMessage}
+                fileSize={state.meta.compiledSize}
+                info={state.debugEnvPreset ? state.envPresetDebugInfo : null}
+                options={options}
+                placeholder="Compiled output will be shown here"
+              />
+            )}
+          </SandpackConsumer>
+        </SandpackProvider>
```
xtuc: That is really looking super nice. We can even conserve our styling, right?

CompuIves: Yes! It's set up pretty flexible, in this case we can just give Babel the state of our bundler and that can be used in the UI. 😄

CompuIves:
There are some things that still need to be done and I could get help with:

- [ ] Generate a `.babelrc` based on the sidebar state.
- [ ] Add a File Explorer in the sidebar to view `.babelrc` and `/index.js`.
- [ ] Remove the existing Babel worker.
- [ ] Move things like File Size etc. over to the compiled code coming from Sandpack.
- [ ] In babel standalone, externalize `fs` and expose `processOptions` as `Babel.processOptions`. This way we can use `fs` in Babel transpilation.
- [ ] Add `Open in CodeSandbox` button somewhere, the component for this is already exported by `Sandpack`.

And these are some things that should still come to Sandpack, I can look at the first one today:

- [ ] Add an `errors` array field on `managerState`.
- [ ] Create a Babel template (I'll do this later).

If there's anyone interested in helping, I'd really appreciate it 😄. Just take something from the list of todos if you're interested!

CompuIves:
I got the bundle from 3.2MB to 1.7MB.

I think we can still improve the bundle size, I believe that the Netlify build server doesn't have `NODE_ENV=production`, which causes the bundle to include source maps from `emotion`. If `NODE_ENV=production` we'll get a build that's 424KB (not gzipped).

hzoo:
If we do https://github.com/babel/website/pull/1483 or update to v7 you can use `<>` instead of React.Fragement altogether it's really not a big deal.

existentialism:
TODO items for me:

- [ ] Generate `babel-transpiler.json` based on incoming query params to support previewing PRs, etc.
- [ ] Fix 3rd party plugins loading
- [ ] Fix eval
- [ ] Fix prettier
- [ ] Fix minifier

PR #1604, opened 2018-03-14
Fix on-page nav scroll

microbouji:
This fixes a css bug with the height of the on-page nav so that it's visible even when the nav is too short and the page is scrolled to the bottom.

We can wait for this to be upstreamed to docusaurus, or push now and remove it with next docusaurus version bump.


hzoo:

Built with commit ff30f9f4ee49117ee4d3d15a22835ee5c7d9cb80

https://deploy-preview-1604--babel-new.netlify.com

PR #1625, opened 2018-04-13
Clean docs frontmatters to remove warnings

microbouji:
Invalid fields in the docs front matter like `layout` and `description` cause warnings when building or running docusaurus.

I removed the fields from current version docs, and updated the download-readmes script to place comment outside of front matter in https://github.com/babel/website/commit/62218c459d28ed82676892c4b374e0374d0523df.

Old version docs don't cause any warnings but I still removed them for consistency. Manually edited the 7 files in https://github.com/babel/website/commit/eaea5cc72dd763317f52ce756d997e84636f9aac, and did a find and replace in https://github.com/babel/website/commit/ca23dc32da4b7ce00185c02a0d9c2ee4f849971e.

### DIFF
--- a/website/versioned_docs/version-6.x/generator.md
+++ b/website/versioned_docs/version-6.x/generator.md
@@ -1,5 +1,4 @@
 ---
-# Don't edit this file directly, it was copied using scripts/download-readmes.js: 
 id: version-6.x-babel-generator
 title: babel-generator
 sidebar_label: babel-generator
--- a/docs/babelrc.md
+++ b/docs/babelrc.md
@@ -1,7 +1,5 @@
 ---
 title: .babelrc
-description: How to use a .babelrc
-
 id: babelrc
 ---
 
--- a/docs/caveats.md
+++ b/docs/caveats.md
@@ -1,7 +1,5 @@
 ---
 title: Caveats
-description: Just some things to keep in mind when using Babel.
-
 id: caveats
 ---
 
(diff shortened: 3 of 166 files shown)


hzoo:

Built with commit ca23dc32da4b7ce00185c02a0d9c2ee4f849971e

https://deploy-preview-1625--babel-new.netlify.com

website/versioned_docs/version-6.x/generator.md
```diff
@@ -1,5 +1,4 @@
 ---
-# Don't edit this file directly, it was copied using scripts/download-readmes.js:
```
hzoo: I guess these are versioned/ anyway so no need to do have anything here?

microbouji: Yes, so I removed the comment and only left the stuff required by docusaurus like id, original_id etc. Or do you mean we should leave the comment there as well?

hzoo: The purpose of the comments are just so people don't try to modify the files that are imported from other repos. However because if this issue we were thinking we should just move all the docs to this repo and just link to the website instead as the readmes

microbouji: That would certainly make things easier. The way it currently is can also work though. Current version in /docs dynamic with the "don't edit" comment, past versions in /versioned committed to this repo. A script can be written to strip the comments in /versioned after a new version is cut. A little more work, but maybe worth it to avoid that extra click from the npm readme page.

hzoo: yeah the issue is that the linking and everything just doesn't work when the file is generated so it seems like it makes more sense to have generic information in the readmes in the babel/babel repo and then link to the "real" docs on this repo instead.

PR #1655, opened 2018-05-11
WIP: Add a sourcemap interactive mapping to the REPL

giles-v:
Closes https://github.com/babel/website/issues/1291.

This is intended to provide a visual view of the sourcemap generated by the compile, so that REPL users can see exactly what transformations map to which generated output.

One immediate impression, having done this work so far, is that the sourcemap is fairly sparse for this purpose: it will correlate specific symbols in the source, but the helper chunks are unrepresented. Having those linkable to the source would be 💯 .


js/repl/Repl.js
```diff
@@ -474,10 +504,57 @@ class Repl extends React.Component {
...
         this.setState(result, setStateCallback);
       });
   };
 
+  processSourcemap = (rawSourceMap: string) => {
+    const smc = new sourceMap.SourceMapConsumer(rawSourceMap);
+
+    const mappings = [];
+    smc.eachMapping(function (m) {
+      if (m.name) mappings.push(m);
```
loganfsmyth: This logic isn't quite what I'd recommend. If we're going to show mappings, we shouldn't exclude things, especially not based on if there is a name. Along with that:

* Mappings have both a start and an end
* Each original mappings can have multiple generated mappings.

Here is how I would read the mappings:

```
const originalPositions = [];
smc.eachMapping(
  function (m) {
    if (m.originalLine === null) return;

    const last = originalPositions.length > 0 ? originalPositions[originalPositions.length - 1] : null;

    // Ignore multiple matching original locations.
    if (last && last.line === m.originalLine && last.columnStart === m.originalColumn) return;

    // Set the _end_ of the previous mapping to the start of the new one.
    if (
      last &&
      last.line === m.originalLine
    ) {
      last.columnEnd = m.originalColumn;
    }

    originalPositions.push({
      source: m.source,
      name: m.name,
      line: m.originalLine,
      columnStart: m.originalColumn,
      columnEnd: Infinity,
    });
  },
  null,
  sourceMap.SourceMapConsumer.ORIGINAL_ORDER
);

// Compute the spans on the generated file.
smc.computeColumnSpans();

const results = originalPositions.forEach(origPos => {
  const ranges = smc.allGeneratedPositionsFor({
    source: origPos.source,
    line: origPos.line,
    column: origPos.columnStart,
  });

  const generatedRanges = ranges.map(range => ({
    line: range.line,
    columnStart: range.column,
    columnEnd: range. lastColumn + 1
  }));

  results.push({
    original: origPos,
    generated: generatedRanges,
  });
});
```
which will give you an array of the original ranges, and for each one of those, the _multiple_ generated ranges they reference.

loganfsmyth:
> hovering in the source or compiled views to highlight

That'd definitely be my preference UX-wise, and it would also allow us to remove the lower pain and remove the `evaluate` question. Do we know if that would be hard to do with the code editor we use?

giles-v:
@loganfsmyth it’s not hard from what I’ve seen. I just didn’t have time today. Coming soon...

Issue #1627, opened 2018-04-16
"Try it out" not working on IE11


darkylmnx:
On the website, the page "try it out" does not work on IE11

http://babeljs.io/repl/

![image](https://user-images.githubusercontent.com/1426357/38811463-d764ed94-4189-11e8-9cfc-d9c88e82cfc1.png)

xtuc:
Yeah Promise is not available on IE11. We should probably use `babel-preset-env`, and more likely on the new website.

hzoo:
Same issue as https://github.com/babel/babel/issues/7607 I believe, we aren't using transform runtime or we should add the polyfill

RusinovAnton:
I'm currently working on the fix. Its pretty trivial for the old site but in the new site there is another error arise:
```
Object doesn't support property or method 'forEach'
```
Which is forEach method of NodeList.  
The work on polyfill for this is in progress:
https://github.com/zloirock/core-js/pull/325

hzoo:
Should be ok

CoeJoder:
Does not work; getting same error.  IE 11.

balzzac:
Try it out doesn't work, I get the same error in Chrome Canary Version 74.0.3698.0 (Official Build) canary (64-bit) and Chrome Version 72.0.3626.96 (Official Build) (64-bit)

TrevorBurnham:
I believe this issue should be reopened. Currently, loading the "Try it out" page in IE11 yields a blank page (apart from the header) and these console errors:

<img width="343" alt="Screen Shot 2019-11-27 at 9 41 42 AM" src="https://user-images.githubusercontent.com/224895/69732795-442eaf00-10fa-11ea-80f9-5036b4396cb0.png">

PR #1662, opened 2018-05-22
Fix REPL for IE11 at new site

RusinovAnton:
refs: https://github.com/babel/website/issues/1627

I've added "babel-polyfill" to make REPL work as Henry suggested.

There is another error is still being thrown but it doesn't affect REPL page:
https://github.com/babel/website/issues/1627#issuecomment-391074217

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -11,6 +11,7 @@
     [
       "env",
       {
+        "useBuiltIns": true,
         "modules": false,
         "targets": {
           "browsers": ["last 3 versions"]
--- a/.babelrc
+++ b/.babelrc
@@ -13,7 +13,7 @@
       {
         "modules": false,
         "targets": {
-          "browsers": ["last 3 versions", "ie >= 11"]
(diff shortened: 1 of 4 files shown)


hzoo:

Built with commit b082d0197cb83ef6ccc7f2e72aac623c93aae95c

https://deploy-preview-1662--babel-new.netlify.com

.babelrc
```diff
@@ -13,7 +13,7 @@
       {
         "modules": false,
         "targets": {
-          "browsers": ["last 3 versions"]
+          "browsers": ["last 3 versions", "ie >= 11"]
```
existentialism: IIRC, isn't `ie >=11` included in `last 3 version`?

RusinovAnton: Yeah, you're right actually. Silly me assumed that edge is the next version of IE. Going to fix it now

hzoo:
We can still add `useBuiltIns: true` in preset-env for the polyfill?

PR #1673, opened 2018-05-28
[WIP] File Name Extension Add

AkshatSrivastava2:
File Name Field Add.


hzoo:

Built with commit 5694645237c3360490ae1f20361097d3fc94d396

https://deploy-preview-1673--babel-new.netlify.com

js/repl/compile.js
```diff
@@ -81,7 +81,7 @@ export default function compile(code: string, config: CompileConfig): Return {
   try {
     const babelConfig = {
       babelrc: false,
-      filename: "repl",
+      filename: config.fileName,
```
loganfsmyth: I think I'd prefer we name this `config.filename` in lowercase so it would be more consistent with Babel's own option name.

js/repl/ReplOptions.js
```diff
@@ -215,6 +217,15 @@ class ExpandedContainer extends Component<Props, State> {
               />
               File Size
             </label>
+            <label className={styles.settingsLabel}>
+            File Name
+              <input
```
loganfsmyth: At least on the published sample, it doesn't seem like I'm actually able to type in this box?

js/repl/compile.js
loganfsmyth: On the sample page, this appears to not be set. Not sure what the issue is. Maybe it's not getting passed along somewhere?

alexzherdev:
@AkshatSrivastava2 are you still working on this? I can pick up if necessary, otherwise you can check out #1676 (which I think is similar) for hints. cc @loganfsmyth

AkshatSrivastava2:
@alexzherdev If you could help me with where am I doing wrong would be great. It will be my first contribution in the open source community.
Otherwise, you may wish to proceed with the issue. :)

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,36 @@
+<component name="ProjectCodeStyleConfiguration">
```
alexzherdev: All these files in `.idea/` should not be committed

docs/parser.md
```diff
@@ -154,6 +154,7 @@ require("@babel/parser").parse("code", {
 > NOTE: When a plugin is specified multiple times, only the first options are considered.
 
 - `decorators`:
+  - `decoratorsBeforeExport` (`boolean`)
```
alexzherdev: Hmm why is this change here?

js/repl/Repl.js
```diff
@@ -71,6 +71,7 @@ type State = {
   pluginSearch: ?string,
   showOfficialExternalPlugins: boolean,
   loadingExternalPlugins: boolean,
+  filename: ?string,
```
alexzherdev: You also need to add the setting to the state initializer similar to https://github.com/babel/website/pull/1676/files#diff-13734ac24709204bbae97c4b659d1ba1R149

js/repl/ReplOptions.js
```diff
@@ -213,6 +215,15 @@ class ExpandedContainer extends Component<Props, State> {
               />
               File Size
             </label>
+
+            <label className={styles.settingsLabel}>
```
alexzherdev: This label is missing a closing tag after the input, hence the broken build.

js/repl/ReplOptions.js
```diff
@@ -213,6 +215,15 @@ class ExpandedContainer extends Component<Props, State> {
...
+            File Name
```
alexzherdev: This should be indented one level deeper

js/repl/ReplOptions.js
```diff
@@ -213,6 +215,15 @@ class ExpandedContainer extends Component<Props, State> {
...
+              <input
+                type= "text"
```
alexzherdev: Space after `=` isn't necessary

js/repl/ReplOptions.js
```diff
@@ -213,6 +215,15 @@ class ExpandedContainer extends Component<Props, State> {
...
+                className={styles.inputFileBoxLeft}
+                size={fileNameBoxSize}
+                value={filename}
```
alexzherdev: For this controlled component to work properly, it needs to have an `onChange` handler to report the changes happening to it. You can probably follow how I did it https://github.com/babel/website/pull/1676/files#diff-ecb6fb48984cd148bfc45fb63ca3f34aR225

js/repl/types.js
```diff
@@ -80,6 +80,7 @@ export type PluginState = LazyLoadedState & {
 
 export type PluginStateMap = { [name: string]: PluginState };
 
+export type filename = "file.js";
```
alexzherdev: You don't need this type, I believe what this is saying is that a filename can only be equal to `file.js`, which isn't what we want.

PR #1702, opened 2018-06-22
docs: styling babel-parser

Aladdin-ADD:
align to middle looks so strange.
image

### DIFF
--- a/docs/parser.md
+++ b/docs/parser.md
@@ -4,7 +4,7 @@ title: babel-parser
 sidebar_label: babel-parser
 ---
 
-<p align="center">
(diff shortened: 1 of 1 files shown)


hzoo:

Built with commit 3fd234be030a1ae88f113dab2c16ddc7ef0bf3a5

https://deploy-preview-1702--babel-new.netlify.com

existentialism:
@Aladdin-ADD thanks!

hzoo:
Maybe we should just get rid of the use of html in that 😛. I think this is because we had the image of "babylon" before

PR #1705, opened 2018-06-23
Publishing ES6+ Post 

hzoo:
Rendered: https://deploy-preview-1705--babel.netlify.com/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages

Purpose: bring up a discussion on how library authors should publish ES6+ and ESM as well as how this is dealt on the consumer side (whether or not you need to compile deps)

- [x] I will move this to https://github.com/babel/website eventually
- [x] Should I move the headings around (maybe a different order is better)? This could probably be better organized
- [x] Am I missing something?
- [x] Should I split the post up, it's getting long?
- [x] Check if I repeated myself too much in between sections
- [x] Anything better with a visual? (either code, picture, etc)
- [x] Pitch v7 release/post, mention patreon/oc?
- [x] mention creating a standard for publishing (latest syntax, ESM) for lib authors to use, create a way to validate, npm should promote, tooling should use.

### DIFF
--- a/website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
+++ b/website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
@@ -59,9 +59,9 @@ Is it as straightforward as just running Babel over `node_modules`?
 
 ### Compiler Complexity
 
-All code (compilers are no different) has bugs. Although it shouldn't deter us from making this a common practice, we should be aware that compiling dependencies does increase the surface area of issues and complexity.
+Although it shouldn't deter us from making this a common practice, we should be aware that compiling dependencies does increase the surface area of issues and complexity.
 
-- It was never in Babel's scope to compile dependencies: we actually got issue reports that people would actually accidently do it, making the build slower.
+- Compilers are no different than other programs and has bugs. 
 - `preset-env` itself could have because we use [`compat-table`](https://kangax.github.io/compat-table/es6/) for our data vs. [test262](https://github.com/tc39/test262) (the official test suite).
 - Browsers themselves can have issues with running native ES2015+ code vs. ES5.
 - There is still a question of determining what is "supported": https://github.com/babel/babel-preset-env/issues/54 is an example of an edge case.
@@ -96,15 +96,17 @@ An example of an issue is how [`this` gets converted to `undefined`](https://git
 
 This was [changed in v7](https://github.com/babel/babel/pull/6280) so that it won't auto inject the `"use strict"` directive unless it actually is a `module`.
 
+It was also not in Babel's original scope to compile dependencies: we actually got issue reports that people would accidently do it, making the build slower. So there may be a lot of defaults and documentation in the tooling that purposely not allow compiling `node_modules`.
+
 ### Using Non-Standard Syntax
 
 There are many issues with *shipping* uncompiled proposal syntax (this post was inspired by [Dan's concern](https://twitter.com/dan_abramov/status/1009179550134296577) about this).
 
 #### Staging Process
 
-- The [TC39 staging process](https://tc39.github.io/process-document/) does not always move forward: there is the possibility for a proposal to move to any point in the process: even moving backwards from Stage 3 to Stage 2 as the case with Numeric Separators (`1_000`) or dropped entirely (`Object.observe()`, and others we may have forgotten 😁)
-- Summary: Stage 0 has no criteria and is just an idea, Stage 1 is accepting that the problem is worth solving, Stage 2 is about describing a solution in spec text, Stage 3 means the specific solution is thought out, and Stage 4 means that it is ready for inclusion with tests, multiple browser implementations, and in-the-field experience.
-- The committee would advise that Stage 0-2 is still experimental.
+The [TC39 staging process](https://tc39.github.io/process-document/) does not always move forward: a proposal can move to any point in the process: even moving backwards from Stage 3 to Stage 2 as the case with Numeric Separators (`1_000`),  dropped entirely (`Object.observe()`, and others we may have forgotten 😁), or just stall like function bind (`a::b`) or decorators until recently.
+
+- Summary of the Stages: Stage 0 has no criteria and is just an idea, Stage 1 is accepting that the problem is worth solving, Stage 2 is about describing a solution in spec text, Stage 3 means the specific solution is thought out, and Stage 4 means that it is ready for inclusion with tests, multiple browser implementations, and in-the-field experience.
 
 #### Using Proposals
 
@@ -112,21 +114,21 @@ There are many issues with *shipping* uncompiled proposal syntax (this post was 
 
 We already recommend that people should be careful when using proposals lower than Stage 3, let alone publishing them.
 
-But telling people not to use Stage X goes against the whole purpose of Babel in the first place. A big reason why proposals gain improvements and move forward are because of the feedback the committee gets from real-world usage (whether in production or not) based on using it via Babel.
+But only telling people not to use Stage X goes against the whole purpose of Babel in the first place. A big reason why proposals gain improvements and move forward are because of the feedback the committee gets from real-world usage (whether in production or not) based on using it via Babel.
 
-There is certainly a balance to be had here. We don't want to scare people away from using new syntax (that is a hard sell 😂), but we also don't want people to get the idea that "once it's in Babel, the syntax is official or immutable". Ideally people look into the purpose of a proposal and make the tradeoffs for their use case.
+There is certainly a balance to be had here: we don't want to scare people away from using new syntax (that is a hard sell 😂), but we also don't want people to get the idea that "once it's in Babel, the syntax is official or immutable". Ideally people look into the purpose of a proposal and make the tradeoffs for their use case.
 
 #### Removing the Stage Presets in v7
 
-One of the most common things people do is use the Stage 0 or Stage 2 presets. We plan to remove the stage presets in v7. We thought at first it would be convienent, that people would make their own unofficial ones anyway, or it might help with "JavaScript fatigue". We find now that it just causes more of an issue: people continue to copy/paste configs with out understanding what goes into a preset in the first place. After all, seeing `"stage-0"` says nothing. My hope is that in making the decision to use proposal plugins explicit, people will have to learn what non-standard syntax they are opting into.
+Even though one of the most common things people do is use the Stage 0 preset, we plan to remove the stage presets in v7. We thought at first it would be convienent, that people would make their own unofficial ones anyway, or it might help with "JavaScript fatigue". It seems to cause more of an issue: people continue to copy/paste configs with out understanding what goes into a preset in the first place. 
+
+After all, seeing `"stage-0"` says nothing. My hope is that in making the decision to use proposal plugins explicit, people will have to learn what non-standard syntax they are opting into. More intentionality should lead to a better understanding of not just Babel but JavaScript the language and it's development instead of just it's usage.
 
 ### Publishing Non-standard Syntax
 
 As a library author, publishing non-standard syntax is setting our users up for possible inconsistencies, refactoring, and breakage of their projects. Because a TC39 proposal (even at Stage 3) has a possibility of changing, it means we will inevitability have to change the library code.
 
 At least if we ship the compiled version, it will still work and the library maintainer can change the output so that it works the same as before. Shipping the uncompiled version means that anyone consuming a package will necessarily have to have a build step to use it, and will have to have the same configuration of Babel as us. This is in the same bucket as using TS/JSX/Flow: we wouldn't expect consumers to configure the same compiler environment just because we used them.
-
-TODO: Or a proposal might just stall because people are busy, etc. Decorators are a great example of this.
 
 ### Conflating ESM and ES2015+
 
@@ -147,14 +149,15 @@ Some tools like Rollup/Webpack will also read from another field called `module`
 
 This was introduced so that users could consume ES2015 modules (ESM).
 
-However, the sole intention of this field is ESM, not anything else. The [Rollup docs](https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features) make it clear that it's not intended for future JavaScript syntax.
+However, the sole intention of this field is ESM, not anything else. The [Rollup docs](https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features) that specify the `module` field make it clear that it's not intended for future JavaScript syntax.
 
 Despite this warning, package authors invariably conflate the use of ES modules with the JavaScript language level they authored it in.
+
 As such, we may need another way to signal the language level.
 
 #### Non-scalable Solutions?
 
-A yearly `package.json` key: a common suggestion is for libraries to start publishing ES2015 under another field like `es2015`, e.g. `"es2015": "es2015/redux.js"`.
+A common suggestion is for libraries to start publishing ES2015 under another field like `es2015`, e.g. `"es2015": "es2015/redux.js"`.
 
 ```js
 // @angular/core package.json
@@ -169,30 +172,33 @@ A yearly `package.json` key: a common suggestion is for libraries to start publi
 }
 ```
 
-This works for ES2015 but the question next is what about ES2016? Are we supposed to create a new folder for each year and a new field in `package.json`? That seems unsustainable, and will continue to produce larger `node_modules`.
+This works for ES2015, but it begs the question of what we should do about ES2016? Are we supposed to create a new folder for each year and a new field in `package.json`? That seems unsustainable, and will continue to produce larger `node_modules`.
 
-This was an issue with Babel itself: we had intended to continue to publish yearly presets (`preset-es2015`, `preset-es2016`..) until `preset-env` removed that need.
+> This was an issue with Babel itself: we had intended to continue to publish yearly presets (`preset-es2015`, `preset-es2016`..) until we realized that `preset-env` would remove that need.
 
-Publishing it based on specific environments/syntax would seem to be even worse as the amount of combinations only increases.
+Publishing it based on specific environments/syntax would seem to be just as untenable as the amount of combinations only increases (`"ie-11-arrow-functions"`).
 
-What about providing just the source? That may have similar problems if we used non-standard syntax as mentioned earlier.
+What about distributing just the source itself? That may have similar problems if we used non-standard syntax as mentioned earlier.
 
-Having a `esnext` field may not be helpful either. The "latest" version of JavaScript changes depending on the point in time we authored the code.
-
-One suggestion is to provide an ES5 version but also publish a version that includes the latest standard syntax so it can be compiled with `preset-env`.
+Having a `esnext` field may not be entirely helpful either. The "latest" version of JavaScript changes depending on the point in time we authored the code.
 
 ### Dependencies May Not Publish ES2015+
 
-- Due to complexity and tooling, it may be difficult for projects to publish ES2015/ESM without more setup. This is probably the biggest issue to get right, even docs aren't enough. We should add some feature requests to `@babel/cli` to make this easier, and maybe make the `babel` package do this by default? Tools like @developit's [microbundle](https://github.com/developit/microbundle) may help a lot with this.
-- How do we deal with polyfills (this will be an upcoming post)? What would it look like for a library author not to have to think about polyfills (or the user).
+This effort will only be standard if it becomes straightfoward to do as a library author. It will be hard to argue for the significance of this change if both new and popular libraries aren't able to ship the latest syntax.
 
-With that said, how does Babel help with all this?
+Due to the complexity and tooling setup, it may be difficult for projects to publish ES2015+/ESM. This is probably the biggest issue to get right, and more documentation just isn't enough.
+
+For Babel, we may need to add some feature requests to `@babel/cli` to make this easier, and maybe make the `babel` package do this by default? Or we should integrate better with tools like @developit's [microbundle](https://github.com/developit/microbundle).
+
+And how do we deal with polyfills (this will be an upcoming post)? What would it look like for a library author not to have to think about polyfills (or the user).
+
+With all that said, how does Babel help with all this?
 
 ## How Babel v7 Helps
 
-As I've discussed, compiling dependencies in Babel v6 can be pretty painful. Babel v7 will address some of these pain points. 
+As we've discussed, compiling dependencies in Babel v6 can be pretty painful. Babel v7 will address some of these pain points. 
 
-One issue (which we have fixed in v7) is around config lookup. Babel currently runs per file so when compiling a file, it tries to find the closest config to know what to compile against. If it doesn't find it in the same directory, it keeps looking up directories (this is why if we don't specify a config but have one in our home directory it might try to load that instead).
+One issue is around configuration lookup. Babel currently runs per file so when compiling a file, it tries to find the closest config ([`.babelrc`](https://babeljs.io/docs/en/next/babelrc)) to know what to compile against. It will keep looking up the directory tree if it doesn't find it in the current folder.
 
 ```
 project
@@ -206,7 +212,7 @@ project
 
 We made a [few changes](https://github.com/babel/babel/pull/7358):
 
-- One is to stop lookup at the package boundary (stop when we find a `package.json`). This makes sure Babel won't try to load a config file outside.
+- One is to stop lookup at the package boundary (stop when we find a `package.json`). This makes sure Babel won't try to load a config file outside the app, the most surprising being with it finds one in your home directory.
 - If we use a monorepo, we'll may want to have a `.babelrc` per-package that extends some other central config.
 - Babel itself is a monorepo, so instead we are using the new `babel.config.js` which allows us to resolve all files to that config (no more lookup).
 - We added an [`"overrides"`](https://github.com/babel/babel/pull/7091) option which allows us to basically create a new config for any set of paths.
@@ -239,16 +245,24 @@ module.exports = {
 }
 ```
 
-## Proposals to Discuss
+## Recommendation to Discuss
 
-Potential recommendation: package authors should also publish a version compiled down to latest syntax (no experimental proposals) under a new key we can standardize on (I don't believe `module` should be that key) but continue to publish ES5 under `main`. Consumers can use `preset-env` and opt-in into running on `node_modules`.
+My recommendation is that package authors should continue to publish ES5/CJS under `main` for backwards compat with current tooling and workflow but also publish a version compiled down to latest syntax (no experimental proposals) under a new key we can standardize on like `main-es` (I don't believe `module` should be that key).
 
-Maybe we should decide on another key in `package.json`, maybe `"es"`? Reminds me of a poll I made for [babel-preset-latest](https://twitter.com/left_pad/status/758429846594850816).
+> Maybe we should decide on another key in `package.json`, maybe `"es"`? Reminds me of a poll I made for [babel-preset-latest](https://twitter.com/left_pad/status/758429846594850816).
+
+Compiling dependencies isn't just something for one project/company to take advantage of: it requires a push by the whole community to move forward. Even though this effort will be natural, it might require some sort of standardization: we can implement a set of criteria for how librarys can opt-in to publishing ES2015+ and verify this via CI/tooling/npm itself.
+
+Documentation needs to updated to mention the benefits of compiling `node_modules`, how to do so for the library authors, and how to consume it in bundlers/compilers.
+
+With Babel 7, consumers can more safely use `preset-env` and opt-in into running on `node_modules` with new config options like `overrides`.
 
 ## Let's Do This!
 
 Hopefully this is an encouraging call to action for looking into moving forward to make compiling dependencies more first class. It's not just about the specific ES2015/ES5 distinction.
 
 Babel v7 should be out soon. This post goes into some of the ways it should help with this effort but we'll need everyone's help to change the ecosystem: more education, published packages that do this, and better tooling and sustainability.
+
+> Thanks to the [many](https://twitter.com/left_pad/status/1010280464840617984) folks who offered to review this post including [@chrisdarroch](https://twitter.com/chrisdarroch), [@existentialism](https://twitter.com/existentialism), [@betaorbust](https://twitter.com/betaorbust), [@_developit](https://twitter.com/_developit), [@jdalton](https://twitter.com/jdalton), [@bonsaistudio](https://twitter.com/bonsaistudio).
 
 <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script>
--- a/website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
+++ b/website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
@@ -1,41 +1,40 @@
 ---
 layout: post
-title:  "On Consuming (and Publishing) ES6+ Packages"
+title:  "On Consuming (and Publishing) ES2015+ Packages"
 author: Henry Zhu
 authorURL: https://twitter.com/left_pad
 date:   2018-06-25 12:00:00
 categories: announcements
-share_text: "On Consuming (and Publishing) ES6+ Packages"
+share_text: "On Consuming (and Publishing) ES2015+ Packages"
 ---
 
 How can we make compiling dependencies not just possible, but normal?
 
 <!--truncate-->
 
-This post is inspired by [Dan's tweet](https://twitter.com/dan_abramov/status/1009179550134296577) about how we can conflate shipping source code, ES6+, ESM
-
-This is a pretty enabling feature request for the whole ecosystem, so I'm glad we've tried to make this easier in Babel v7 (I just realized this whole post is a good pitch for using it). Hopefully it can be made more standard moving forward (if we're collectively able to figure out some things I outline below).
-
-## Why
-
-Why is compiling your dependencies desirable in the first place (vs. just your own code)?
-
-We can ship less code to users, since JavaScript has a [cost](https://medium.com/dev-channel/the-cost-of-javascript-84009f51e99e).
+Being able to compile dependencies is an enabling feature request for the whole ecosystem. With some of the changes we made in Babel v7, hopefully it can be made more standard moving forward.
 
 ## Assumptions
 
 - You can ship to modern browsers (don't have to support IE) or are able to send multiple types of bundles (one way being [checking for script`type=module`](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/)).
-- And/or the dependencies actually publish ES6+ instead of ES5/ES3.
+- Your dependencies actually publish ES2015+ instead of ES5/ES3.
 
-## Backing up a Bit: Talking About `preset-env`
+## Why
 
-If you think about it, the argument for why this would be helpful is the same for why we eventually introduced `@babel/preset-env`.
+Why is compiling your dependencies desirable in the first place (vs. your own code)?
 
-Babel used to be `6to5`, since it only converted from ES6 to ES5. Back then, the browser support for ES6 was almost non-existent so the idea of a JS compiler was an amazing contribution to our ecosystem.
+- You can ship less code to users, since JavaScript has a [cost](https://medium.com/dev-channel/the-cost-of-javascript-84009f51e99e).
+- You are able to make the tradeoff of where your code is able to run vs. the library.
 
-I went through with implementing the preset even though there was so much complexity since it seemed to solve multiple issues for us. I didn't realize how much it would move us forward now that it's finally [surpassed `preset-es2015`](https://npm-stat.com/charts.html?package=babel-preset-es2015&package=babel-preset-env&from=2016-11-01&to=2018-06-22) which is a nice accomplishment.
+## The Ephemeral JavaScript Runtime
 
-When I think about Babel's role not just in moving the language itself forward but helping us develop websites better, I saw that we would eventually want to move past only compiling to ES5. Creating `preset-env` helps us align with the browsers since if we only compiled to ES5, no one would ever run native code in the browsers.
+When thinking about Babel's role not just in moving the language itself forward, we saw that developers would eventually want to move past only compiling to ES5. 
+
+Thus, the argument for why compiling dependencies would be helpful is the same for why we eventually introduced `@babel/preset-env`.
+
+Babel used to be `6to5`, since it only converted from ES2015 to ES5. Back then, the browser support for ES2015 was almost non-existent, so the idea of a JavaScript compiler was both novel and useful: you could write modern code and have it work for all of your users. But what about the browser runtimes themselves?
+
+Browsers will eventually catch up to the standard (and have with ES2015). Creating `preset-env` helps us align with the browsers since if we only compiled to ES5, no one would ever run native code in the browsers.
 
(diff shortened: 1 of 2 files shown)


hzoo:

Built with commit 24add6a46f5c8b38bef2c9ceec9101ae6613189e

https://deploy-preview-1705--babel-new.netlify.com

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+There is certainly a balance to be had here. We don't want to scare people away from using new syntax (that is a hard sell 😂), but we also don't want people to get the idea that "once it's in Babel, the syntax is official or immutable". Ideally people look into the purpose of a proposal and make the tradeoffs for their use case.
+
+One of the most common things people do is use the Stage 0 or Stage 2 presets. We plan to remove the stage presets in v7 after much back and forth. I thought at first it would be convienent, that people would make their own unofficial ones anyway, or it might help with "JavaScript fatigue". I find now that it just causes more of an issue: people continue to copy/paste configs with out understanding what goes into a preset in the first place. After all, seeing `"stage-0"` says nothing. My hope is that in making the decision to use proposal plugins explicit, people will have to learn what non-standard syntax they are opting into.
```
jdalton: You switch voice here from "we" to "I". You can drop the "after much back and forth" bit too. "My hope" -> "Our hope".

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+#### Staging Process
+
+- The [TC39 staging process](https://tc39.github.io/process-document/) does not always move forward: there is the possibility for a proposal to move to any point in the process: even moving backwards from Stage 3 to Stage 2 as the case with Numeric Separators (`1_000`) or dropped entirely (`Object.observe()`, and others you may have forgotten)
+- It can be hard to differentitate between the stages if you aren't involved: most people would advise that Stage 0-2 is unstable though.
+- Summary: Stage 0 has no criteria and is just an idea, Stage 1 is accepting that the problem is worth solving, Stage 2 is about describing a solution in spec text, Stage 3 means the specific solution is thought out, and Stage 4 means that it is ready for inclusion with tests, multiple browser implementations, and in-the-field experience.
+
+#### Publishing Non-standard Syntax
+
+So actually publishing it in a library may cause even more issues. Because a TC39 proposal (even at Stage 3) has a possibility of changing, it means you will inevitability have to change the code.
```
jdalton: Drop the `So` (it's like starting in the middle of an idea but it's the start of a new paragraph so should be a solid start). "Publishing non-standard syntax in a library is setting your users up for possible inconsistencies, refactoring, and breakage of their projects."

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+However, this isn't necessarily the case for your dependencies (though you _should_ take ownership of them).
+
+## What Then?
+
+The next step is to start thinking "how do we actually make this happen"?
+
+Is it as straightforward as just running Babel over `node_modules` with the same configuration for your application?
+
+## Caveats
```
chrisdarroch: This section could use a different name and a lead-in sentence, though I'm not completely sure what it should be yet.

These are the problems that come with not knowing enough about un-compiled code, right? The reader may not have an appreciation for the considerations and complexities of compiling code. If you're attempting to give them those considerations, this section isn't really about "caveats" (since the reader isn't the one attempting to compile the dependencies; Babel is).

chrisdarroch: Could be called "The problems with compiling code".

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+Every compiler has bugs, although the risk is lower if there are so many users and sufficient tests. We should just be aware that compiling dependencies does increase the surface area of potential bugs, but I don't believe that should deter us from making this a common practice.
+
+- Babel v6 assumed everything that it compiled was a module and thus in strict mode (one could argue this is actually a good thing). Thus if you tried to run Babel on all your `node_modules` and it encountered something that was a `script` like a jQuery plugin it might cause a lot of issues. This is changed in v7 so that it won't always auto inject the `"use strict"` directive unless it actually is a module.
+- React Native has always compiled `node_modules` by default and has probably learned a lot from issues like ^.
+- `preset-env` could have its own bugs because we use `compat-table` vs. test262.
+- Browsers themselves can have issues with running native ES6+ code vs. ES5.
+- Question of how do we determine what is "supported": https://github.com/babel/babel-preset-env/issues/54 example of edge case.
+
+### Help with Babel v7 Config
```
chrisdarroch: This heading needs something before it, because now we're getting in to "what to do about it" rather than this being a further problem with compiling.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+    presets: [
+      ['env', { 
+        targets: { browsers: ["last 2 versions"] },
+      }],
+    ],
+  }],
+}
+```
+
+### Let's Do This!
```
chrisdarroch: Heading level should be `##`

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+The next step is to start thinking "how do we actually make this happen"?
+
+Is it as straightforward as just running Babel over `node_modules` with the same configuration for your application?
+
+## Caveats
+
+### Using Non-Standard Syntax
+
+This is what Dan was warning about. There are many issues with *shipping* uncompiled proposal syntax.
```
chrisdarroch: No need to invoke Dan here.

You can make a short point about non-standard syntax == assumptions the direction and the way the language will evolve. That will allow the sub-headings to expand upon this argument as extra detail.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+#### Staging Process
+
+- The [TC39 staging process](https://tc39.github.io/process-document/) does not always move forward: there is the possibility for a proposal to move to any point in the process: even moving backwards from Stage 3 to Stage 2 as the case with Numeric Separators (`1_000`) or dropped entirely (`Object.observe()`, and others you may have forgotten)
+- It can be hard to differentitate between the stages if you aren't involved: most people would advise that Stage 0-2 is unstable though.
+- Summary: Stage 0 has no criteria and is just an idea, Stage 1 is accepting that the problem is worth solving, Stage 2 is about describing a solution in spec text, Stage 3 means the specific solution is thought out, and Stage 4 means that it is ready for inclusion with tests, multiple browser implementations, and in-the-field experience.
+
+#### Publishing Non-standard Syntax
```
chrisdarroch: This may be a new `##`-level section. Now we're out of "problems with compiling", and in to "problems when code with assumptions in them are published without a way for consumers to easily know or deal with what those assumptions were".

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+So actually publishing it in a library may cause even more issues. Because a TC39 proposal (even at Stage 3) has a possibility of changing, it means you will inevitability have to change the code.
+
+At least if you ship the compiled version, it will still work and the library maintainer can change the output so that it works the same as before. Shipping the uncompiled version means that anyone consuming a package will necessarily have to have a build step to use it, and will have to have the same version of Babel as you. This is no different than using TS/JSX/Flow which you wouldn't publish either.
+
+TODO: Or a proposal might just stall because people are busy, etc. Decorators are a great example of this.
```
chrisdarroch: This looks like a `TODO` for the "Staging process" section, not this one.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+- It can be hard to differentitate between the stages if you aren't involved: most people would advise that Stage 0-2 is unstable though.
+- Summary: Stage 0 has no criteria and is just an idea, Stage 1 is accepting that the problem is worth solving, Stage 2 is about describing a solution in spec text, Stage 3 means the specific solution is thought out, and Stage 4 means that it is ready for inclusion with tests, multiple browser implementations, and in-the-field experience.
+
+#### Publishing Non-standard Syntax
+
+So actually publishing it in a library may cause even more issues. Because a TC39 proposal (even at Stage 3) has a possibility of changing, it means you will inevitability have to change the code.
+
+At least if you ship the compiled version, it will still work and the library maintainer can change the output so that it works the same as before. Shipping the uncompiled version means that anyone consuming a package will necessarily have to have a build step to use it, and will have to have the same version of Babel as you. This is no different than using TS/JSX/Flow which you wouldn't publish either.
```
chrisdarroch: when you say "will necessarily ... have to have the same version of Babel as you", you probably mean the exact same compiler configuration -- it's not just about the literal package version.

chrisdarroch: "This is no different than using TS/JSX/Flow which you wouldn't publish either." -- might be better phrased as "This is in the same bucket as using TS/JSX/Flow: you wouldn't expect consumers to configure the same compiler environment just because you used them."

The explicit point that package authors should not expect consumers to configure Babel could then be made.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+}
+```
+
+This was introduced so that users could consume ES2015 modules (ESM).
+
+However, the sole intention of this field is ESM, not anything else. The [Rollup docs](https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features) make it clear that it's not intended for future JavaScript syntax.
+
+What this means is that we may need another way to handle ES6+ even though some libraries may choose to do this anyway.
+
+A common suggestion is for libraries to start publishing ES2015 under another field like `es2015`, e.g. `"es2015": "es2015/redux.js"`.
```
chrisdarroch: What follows are some non-scalable solutions to the problem described above. As such, you could add a `####`-level heading here to separate the problem from the (bad) solutions.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+  "module": "./fesm5/core.js",
+  "es2015": "./fesm2015/core.js",
+  "esm5": "./esm5/core.js",
+  "esm2015": "./esm2015/core.js",
+  "fesm5": "./fesm5/core.js",
+  "fesm2015": "./fesm2015/core.js",
+}
+```
+
+This works for ES2015 but the question next is what about ES2016? Are we supposed to create a new folder for each year and a new field in `package.json`? That seems unsustainable, and will continue to produce larger `node_modules`.
```
chrisdarroch: 💯
And what about all those non-standard features? Should you create an "esm2015+async-await+decorators" field? The permutation space is untenable.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+  "main": "lib/redux.js", // ES5 + CJS
+  "module": "es/redux.js", // ES5 + ESM
+}
+```
+
+This was introduced so that users could consume ES2015 modules (ESM).
+
+However, the sole intention of this field is ESM, not anything else. The [Rollup docs](https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features) make it clear that it's not intended for future JavaScript syntax.
+
+What this means is that we may need another way to handle ES6+ even though some libraries may choose to do this anyway.
```
chrisdarroch: > Despite this warning, package authors will inevitably conflate use of ES modules with the JavaScript language level they authored it in.
> As such, we may need another way to signal the language level.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
+---
+layout: post
+title:  "On Consuming (and Publishing) ES6+ Packages"
```
chrisdarroch: You switch between use of "ES6" and "ES2015" a few times during the article. Pick one, add the other in brackets after the first occurrence, then find-replace the rest.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+  "esm5": "./esm5/core.js",
+  "esm2015": "./esm2015/core.js",
+  "fesm5": "./fesm5/core.js",
+  "fesm2015": "./fesm2015/core.js",
+}
+```
+
+This works for ES2015 but the question next is what about ES2016? Are we supposed to create a new folder for each year and a new field in `package.json`? That seems unsustainable, and will continue to produce larger `node_modules`.
+
+I feel like this is similar to how we had a Babel preset for each year and ended up creating `preset-latest` which included all the yearly plugins (we removed that soon after in favor of `preset-env`).
```
chrisdarroch: Interesting history, but it's an aside, so could be removed.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+Having a field like `es-latest` or `esnext` may not be helpful either because we want the library versions to be fixed yet changing. And providing only the source may not always be helpful if you use non-standard syntax, TS, Flow, JSX, etc which might have different behavior based on compiler version. Maybe we need some changes on the `npm` side?
```
chrisdarroch: You could clarify what you mean by "fixed yet changing" a bit more.

> Having a field named `es-latest` or `esnext` would not help, either. What, for instance, is the "latest" version of JavaScript? The answer changes depending on the point in time you authored the code!
>
> What about providing just the source? That would have similar problems if you used non-standard syntax, TS, Flow, JSX, etc., since they may yield different output depending on the versions of the compilers you used at the time you authored them.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
chrisdarroch: I may, however, be misunderstanding the target audience. If this is a discussion with fellow compiler authors, then they are indeed considerations to take in to account when moving forward.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+    ],
+  }],
+}
+```
+
+### Let's Do This!
+
+Hopefully this is an encouraging call to action for looking into moving forward to make compiling dependencies more first class. It's not just about the specific ES6/ES5 distinction.
+
+Babel v7 should be out soon, this post goes into some of the ways it should help with this effort but we'll need everyone's help to change the ecosystem (more education, published packages that do this, and better tooling and sustainability).
```
chrisdarroch: > Babel v7 makes some headway in making compilation easier, but it's not enough to change only the tool. We'll need everyone's help...

chrisdarroch: This should be your last paragraph.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+Maybe we should decide on another key in `package.json`, maybe `"es"`? Reminds me of the poll I made for [babel-preset-latest](https://twitter.com/left_pad/status/758429846594850816).
+
+Potential recommendation: package authors should also publish a version compiled down to latest syntax (no experimental proposals) under a new key we can standardize on (I don't believe `module` should be that key) but continue to publish ES5 under `main`. Consumers can use `preset-env` and opt-in into running on `node_modules`.
```
chrisdarroch: Consider moving this + the paragraph above it in to a "proposals for change" section, just before the "let's do it". This will give people one place to look at your recommendations, without needing to hunt through the entire post to find them.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+This works for ES2015 but the question next is what about ES2016? Are we supposed to create a new folder for each year and a new field in `package.json`? That seems unsustainable, and will continue to produce larger `node_modules`.
+
+I feel like this is similar to how we had a Babel preset for each year and ended up creating `preset-latest` which included all the yearly plugins (we removed that soon after in favor of `preset-env`).
+
+Having a field like `es-latest` or `esnext` may not be helpful either because we want the library versions to be fixed yet changing. And providing only the source may not always be helpful if you use non-standard syntax, TS, Flow, JSX, etc which might have different behavior based on compiler version. Maybe we need some changes on the `npm` side?
+
+One suggestion is to provide an ES5 version but also publish a version that includes the latest standard syntax so it can be compiled with `preset-env`.
+
+### Dependencies May Not Publish ES6+
```
chrisdarroch: This section feels out of place in the article. If they don't publish ES6+ and don't use ESM, it's probably using a lowest common denominator language level, so should work everywhere without compilation? If that's the case, it doesn't support the argument of making compilation the norm.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+TODO:
+
+- Due to complexity and tooling, it may be difficult for projects to publish ES6/ESM without more setup. This is probably the biggest issue to get right, even docs aren't enough. We should add some feature requests to `@babel/cli` to make this easier, and maybe make the `babel` package do this by default? Tools like @developit's [microbundle](https://github.com/developit/microbundle) may help a lot with this.
+- How do we deal with polyfills? This could be it's own post (I never finished it). Mention `preset-env` + polyfills. What would it look like for a library author not to have to think about polyfills (or the user).
+
+### Compiler Complexity
```
chrisdarroch: You could move this section between the "What then" and "Caveats" section. Compiler complexity is a problem, but the social problems are more challenging. Lead with the "easy" problem first ;)

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+- Babel v6 assumed everything that it compiled was a module and thus in strict mode (one could argue this is actually a good thing). Thus if you tried to run Babel on all your `node_modules` and it encountered something that was a `script` like a jQuery plugin it might cause a lot of issues. This is changed in v7 so that it won't always auto inject the `"use strict"` directive unless it actually is a module.
+- React Native has always compiled `node_modules` by default and has probably learned a lot from issues like ^.
+- `preset-env` could have its own bugs because we use `compat-table` vs. test262.
+- Browsers themselves can have issues with running native ES6+ code vs. ES5.
+- Question of how do we determine what is "supported": https://github.com/babel/babel-preset-env/issues/54 example of edge case.
+
+### Help with Babel v7 Config
+
+It can be pretty painful to compile `node_modules` in Babel v6. The reason is both that we never thought of it as something we wanted as well as getting issue reports that people would actually accidently do it making the build slower.
```
chrisdarroch: Move the second sentence to where you made this point already in the "Compiler complexity" section.

As an intro to this section, you want to establish what Babel v7 changes. I'd suggest something like:

> As I've discussed, compiling dependencies in Babel v6 can be pretty painful. Babel v7 will address some of these pain points.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
betaorbust: Given that you mention preset-es2015 and publishing under es2015/etc, my vote is for 2015. 
It also lumps all the modern language levels under one title type, which will make the mental model clearer vs es5.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+Thus the need isn't just for `6to5` to be renamed to Babel because it compiles to `7to5`, but for Babel to change the implicit assumption it only targets ES5. With `@babel/preset-env`, you are able to write the latest JavaScript and target whichever browser (environment) you want!
+
+But assuming people even know it exists or have starting using it at all, it's mainly for your **application** code.
+
+## Your Code
+
+You have control over your own code to be able to take advantage of preset-env: write in ES6+ and target ES6+ browsers.
+
+However, this isn't necessarily the case for your dependencies (though you _should_ take ownership of them).
```
betaorbust: The paranthetical here could be omitted or changed to convey that you can’t really do this well today, but you’ll want to, to get the benefits we talked about in your application code.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
betaorbust: Or even 
> [...] package authors invariably conflate [...]

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
betaorbust: This seems like a really good argument for only ever publishing at fully-baked and released ES versions, and not having any proposal work (even S3) in your published artifact.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
betaorbust: 100% agree. Community/external tooling/platform support is going to be really necessary to pull this off, so the call to action based on B7’s work looks like it’s the best place to hand off the charge to the reader.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+Babel used to be `6to5`, since it only converted from ES6 to ES5. Back then, the browser support for ES6 was almost non-existent so the idea of a JS compiler was an amazing contribution to our ecosystem.
+
+I went through with implementing the preset even though there was so much complexity since it seemed to solve multiple issues for us. I didn't realize how much it would move us forward now that it's finally [surpassed `preset-es2015`](https://npm-stat.com/charts.html?package=babel-preset-es2015&package=babel-preset-env&from=2016-11-01&to=2018-06-22) which is a nice accomplishment.
+
+When I think about Babel's role not just in moving the language itself forward but helping us develop websites better, I saw that we would eventually want to move past only compiling to ES5. Creating `preset-env` helps us align with the browsers since if we only compiled to ES5, no one would ever run native code in the browsers.
+
+The real difference is realizing that there will _always_ be a sliding window of support: whether it's the JavaScript language that continues to move forward with new syntax, the browsers catching up with the standard, or the companies/projects themselves.
```
betaorbust: I gave a talk proposing shipping progressive levels of transpilation in 2015, and by far the most positive feedback I got from it was around my section addressing this. 
That we have to come up with a forward-safe paradigm because the world and web are always pushing forward. Shifting the mindset from “this is my target” to “this is my target for now”.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+
+- You (your supported environments)
```
ifyoumakeit: I think you can drop this list, which basically states the same points in the above paragraph. The list could work if each bullet was broken down in the sections below.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
ifyoumakeit: This point should be highlighted more. Right now it's buried within this paragraph and is an important update for v7. It could be a paragraph heading.

> We plan to remove the stage presets in v7 after much back and forth.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,218 @@
...
+layout: post
+title:  "On Consuming (and Publishing) ES6+ Packages"
+author: Henry Zhu
+authorURL: https://twitter.com/left_pad
+date:   2018-06-25 12:00:00
+categories: announcements
+share_text: "On Consuming (and Publishing) ES6+ Packages"
+---
+
+How can we make compiling dependencies not just possible, but normal?
```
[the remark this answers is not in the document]
hzoo: Yeah, good point. I should just move it to the point where it actually talks about the issue that got brought up regarding publishing proposal syntax, not need to reference since it works on it's own!

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
hzoo: Yeah I guess I'm combining "compiling" for both the user and author side which might be confusing. Yeah the audience is all developers even if they aren't a library author because they should understand on the library side why this isn't a thing yet?

hzoo:
Ok I think there's not much to add anymore, probably need to remove a lot since there's so much. Last thing to think about is the ending and making sure the main point comes across and a call to action for everyone.

Tone: open the discussion
Main point: "normalizing dependency compilation" (make everything point back to this)
Key takeaways: Get us to stop thinking about compiling to a fixed target and start thinking of our dependencies like our own code.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+    ],
+  }],
+}
+```
+
+## Recommendations to Discuss
+
+We should shift our fixed view of publishing JavaScript to one that keeps up with the standard which pushes forward.
+
+My recommendation is that package authors should continue to publish ES5/CJS under `main` for backwards compat with current tooling and workflow but also publish a version compiled down to latest syntax (no experimental proposals) under a new key we can standardize on like `main-es` (I don't believe `module` should be that key).
```
bcoe: I'm fascinated by this proposal, there are definitely folks at npm, Inc, like @iarna, @ceejbot, and @zkat who would be interested in working with you around some of these goals -- we've had conversations similar to this proposal.

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+How can we make compiling our dependencies not just possible, but normal?
+
+<!--truncate-->
+
+The ability to compile dependencies is an enabling feature request for the whole ecosystem. Starting with some of the changes we made in Babel v7, we hope to see it standardized moving forward.
+
+## Assumptions
+
+- We believe compiling our own app code is useful.
+- We can and will ship to modern browsers that support ES2015+ [natively](https://kangax.github.io/compat-table/es6/) (don't have to support IE) or are able to send multiple kinds of bundles (i.e. [checking for script`type=module`](https://philipwalton.com/articles/deploying-es2015-code-in-production-today/)).
```
mathiasbynens: typo: `script=module`

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+- Babel (the abstraction layer)
+- TC39/ECMAScript Proposals (and Babel implementations)
+
+Thus, the need isn't just for `6to5` to be renamed to Babel because it compiles to `7to5`, but for Babel to change the implicit assumption it only targets ES5. With `@babel/preset-env`, we are able to write the latest JavaScript and target whichever browser (environment)!
+
+Using Babel and `preset-env` helps us keep up with that constantly changing sliding window. However, even if we use it, it's currently used for *our application code* and not what our code depends upon.
+
+## Who Owns Our Dependencies?
+
+Because we have control over our own code, we are able to take advantage of `preset-env`: both by writing in ES2015+ and targeting ES2015+ browsers.
```
mathiasbynens: and by

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+## The Current Complexities in Compiling Dependencies
+
+### Compiler Complexity
+
+Although it shouldn't deter us from making this a common practice, we should be aware that compiling dependencies does increase the surface area of issues and complexity.
+
+- Compilers are no different than other programs and have bugs. 
+- `preset-env` itself could have bugs because we use [`compat-table`](https://kangax.github.io/compat-table/es6/) for our data vs. [test262](https://github.com/tc39/test262) (the official test suite).
+- Browsers themselves can have issues with running native ES2015+ code vs. ES5.
+- There is still a question of determining what is "supported": https://github.com/babel/babel-preset-env/issues/54 is an example of an edge case.
```
mathiasbynens: “see $URL”

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Publishing Non-standard Syntax
+
+As a library author, publishing non-standard syntax is setting our users up for possible inconsistencies, refactoring, and breakage of their projects. Because a TC39 proposal (even at Stage 3) has a possibility of changing, it means we will inevitability have to change the library code. A "new" proposal doesn't mean the idea is fixed or certain but rather that we collectively want to explore the solution space.
+
+At least if we ship the compiled version, it will still work, and the library maintainer can change the output so that it works the same as before. Shipping the uncompiled version means that anyone consuming a package needs to have a build step to use it and needs to have the same configuration of Babel as us. This is in the same bucket as using TS/JSX/Flow: we wouldn't expect consumers to configure the same compiler environment just because we used them.
```
mathiasbynens: This phrasing is confusing, IMHO. How about:

```
At least if we ship the compiled version, it will still work, and the library maintainer can change the source version so that it compiles into code that works the same as before.
```

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+Although it shouldn't deter us from making this a common practice, we should be aware that compiling dependencies does increase the surface area of issues and complexity.
+
+- Compilers are no different than other programs and have bugs. 
+- `preset-env` itself could have bugs because we use [`compat-table`](https://kangax.github.io/compat-table/es6/) for our data vs. [Test262](https://github.com/tc39/test262) (the official test suite).
+- Browsers themselves can have issues with running native ES2015+ code vs. ES5.
+- There is still a question of determining what is "supported": https://github.com/babel/babel-preset-env/issues/54 is an example of an edge case.
+
+#### Specific Issues in v6
+
+Running a `script` as a `module` either causes a `SyntaxError`, new runtime errors, or unexpected behavior due to the differences in semantics between classic scripts and modules.
```
mathiasbynens: We enumerate these differences in detail here: https://developers.google.com/web/fundamentals/primers/modules#intro Maybe worth a link?

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+    ],
+  }],
+}
+```
+
+## Recommendations to Discuss
+
+We should shift our fixed view of publishing JavaScript to one that keeps up with the standard which pushes forward.
+
+My recommendation is that package authors should continue to publish ES5/CJS under `main` for backwards compat with current tooling and workflow but also publish a version compiled down to latest syntax (no experimental proposals) under a new key we can standardize on like `main-es`. I don't believe `module` should be that key.
```
mathiasbynens: Why not?

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Publishing Non-standard Syntax
+
+As a library author, publishing non-standard syntax is setting our users up for possible inconsistencies, refactoring, and breakage of their projects. Because a TC39 proposal (even at Stage 3) has a possibility of changing, it means we will inevitability have to change the library code. A "new" proposal doesn't mean the idea is fixed or certain but rather that we collectively want to explore the solution space.
+
+At least if we ship the compiled version, it will still work, and the library maintainer can change the output so that it works the same as before. Shipping the uncompiled version means that anyone consuming a package needs to have a build step to use it and needs to have the same configuration of Babel as us. This is in the same bucket as using TS/JSX/Flow: we wouldn't expect consumers to configure the same compiler environment just because we used them.
+
+### Conflating ESM and ES2015+
```
mathiasbynens: FWIW, instead of “ESM”, we’ve been calling them just “JavaScript modules” (or “JS modules”).

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+
+When we write `import foo from "foo"` or `require("foo")` and `foo` doesn't have an `index.js`, it resolves to the `main` field in the `package.json` of the module.
+
+Some tools like Rollup/webpack also read from another field called `module` (previously `jsnext:main`). It uses this to instead resolve to the ESM file.
```
mathiasbynens: “ESM” is not explained here

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+```js
+// redux package.json
+{
+  ...
+  "main": "lib/redux.js", // ES5 + CJS
+  "module": "es/redux.js", // ES5 + ESM
+}
+```
+
+This was introduced so that users could consume ES2015 modules (ESM).
```
mathiasbynens: but then here you have “(ESM)”

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+Some tools like Rollup/webpack also read from another field called `module` (previously `jsnext:main`). It uses this to instead resolve to the ESM file.
+
+- An example with [`redux`](https://github.com/reactjs/redux)
+
+```js
+// redux package.json
+{
+  ...
+  "main": "lib/redux.js", // ES5 + CJS
+  "module": "es/redux.js", // ES5 + ESM
```
mathiasbynens: [This should really be `mjs`.](https://developers.google.com/web/fundamentals/primers/modules#mjs)

mathiasbynens: It says “ES5” but the next paragraph says “ES2015 modules” which might be confusing

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
mathiasbynens: ES2015 is just a version number. I’d just call them “JS modules”

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+```js
+// @angular/core package.json
+{
+  "main": "./bundles/core.umd.js",
+  "module": "./fesm5/core.js",
+  "es2015": "./fesm2015/core.js",
+  "esm5": "./esm5/core.js",
+  "esm2015": "./esm2015/core.js",
+  "fesm5": "./fesm5/core.js",
+  "fesm2015": "./fesm2015/core.js",
```
mathiasbynens: use `.mjs` instead of `.js` for the modules

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
```diff
@@ -0,0 +1,276 @@
...
+
+However, the sole intention of this field is ESM, not anything else. The [Rollup docs](https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features) specify that the `module` field makes it clear that it's not intended for future JavaScript syntax.
+
+Despite this warning, package authors invariably conflate the use of ES modules with the JavaScript language level they authored it in.
+
+As such, we may need another way to signal the language level.
+
+#### Non-scalable Solutions?
+
+A common suggestion is for libraries to start publishing ES2015 under another field like `es2015`, e.g. `"es2015": "es2015/redux.js"`.
```
mathiasbynens: mjs

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
mathiasbynens: Maybe link back to the “Conflating ESM and ES2015+” section or reiterate the take-aways here. (I’d prefer the latter)

website/blog/2018-06-25-on-consuming-and-publishing-es6+-packages.md
hzoo: This is actually from angular's package.json, but I'll change it anyway

website/blog/2018-06-26-on-consuming-and-publishing-es2015+-packages.md
```diff
@@ -0,0 +1,274 @@
...
+{
+  "main": "./bundles/core.umd.js",
+  "module": "./fesm5/core.mjs",
+  "es2015": "./fesm2015/core.js",
+  "esm5": "./esm5/core.mjs",
+  "esm2015": "./esm2015/core.mjs",
+  "fesm5": "./fesm5/core.mjs",
+  "fesm2015": "./fesm2015/core.mjs",
+}
+```
```
jdalton: ☝️ While I get the push by some for `.mjs` the `@angular/core` and `redux` packages doesn't actually use it.

hzoo:
Thanks everyone, https://twitter.com/left_pad/status/1011622065189814274!

https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages

website/blog/2018-06-26-on-consuming-and-publishing-es2015+-packages.md
hzoo: yeah that was my question here https://github.com/babel/website/pull/1705#discussion_r198125693

jdalton: You could solve it by removing the angular.js and redux references (saying instead: some projects do this, others do that)

jdalton: > Going to change it to reflect the actual package e348fc0 then?

Okay.

PR #1723, opened 2018-07-13
Fix flow errors

gyujincho:
/cc @sujinleeme


js/repl/Repl.js
```diff
@@ -94,7 +89,7 @@ function toCamelCase(str) {
     });
 }
 
-class Repl extends React.Component {
+class Repl extends React.Component<Props, State> {
   props: Props;
```
existentialism: Nit: we can remove these, since we're now declaring `Props` and `State` in L92.

gyujincho: Thanks! Removed redundant ones.

PR #1725, opened 2018-07-15
Add blog post about pipeline operator & new config

mAAdhaTTah:


mAAdhaTTah:
Just FYI, ping me before this gets merged so I can update the pub date.

hzoo:
Don't worry bout that, can do that too!

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
```diff
@@ -0,0 +1,153 @@
...
+## Current Status
+
+The pipeline operator was originally introduced by [Gilbert Garza](https://github.com/gilbert), providing a clean syntax for "streamlining chained function calls in a readable, functional manner." The pipeline operator has roots in a number of languages, including F#, Hack, Elm, Elixir, and others, but there were two major points of contention in introducing the new syntax to JavaScript:
+
+* Whether and how to introduce placeholders
+* How to handle async / await in the pipeline
+
+### Placeholders
+
+The first issue was the question of placeholders. This was first raised by [Kevin Smith](https://github.com/zenparsing) in [this issue](https://github.com/tc39/proposal-pipeline-operator/issues/84), suggesting Hack-style pipelining requring placeholders for every function call. Placeholders expand the potential use space of the pipeline operator, as you can now embed arbitrary expressions in the pipeline, with the placeholder representing the value returned from the previous step.
```
hzoo: `suggesting Hack-style pipelining requring placeholders` reads kind of weird to me? maybe "who suggested that function call would require Hack-style placeholders"

should we link to https://docs.hhvm.com/hack/operators/pipe-operator?

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
```diff
@@ -0,0 +1,153 @@
...
+await f(x)
+```
+
+Unfortunately, this is intuitively ambiguous, as it could reasonably desugar to this:
+
+```js
+(await f)(x)
+```
+
+While there was some pushback on the idea of including async handling in the pipeline at all, the committee felt, without a strong reason it wasn't doable, it would oppose a pipeline operator that didn't handle async/await. While there are ways to handle Promise-returning functions without explicit syntax, they are too cumbersome to be useful or require a helper function.
```
hzoo: It is possible to remove some of the extra commas here?

While there was some pushback on the idea of including async handling in the pipeline at all, the committee still chose to oppose a pipeline operator that didn't handle async/await?

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
```diff
@@ -0,0 +1,153 @@
...
+x |> f.b;
+```
+
+If there are any operators, parentheses (including for method calls), brackets, or anything other than identifiers and dot punctuators, then it is in topic style, not in bare style, and a topic token is necessary. This is avoids footguns and eliminate ambiguity in bare style code.
+
+Smart pipelines thus resolve the issue of async in an integrative way, allowing all possible expressions to be emedded in a pipeline; not only `await`, but also `typeof`, `yield`, and another other operator desired.
+
+## Where Babel Comes In
+
+Once all three proposals were fleshed out, we realized discussion and debate were unlikely to resolve the inherent tension between them. We decided the best way to make a decision would be with the feedback of developers, using the proposals in real code. Given babel's role in the community, we approached them about including all three of the above proposals into the pipeline operator plugin.
```
hzoo: `babel's` -> `Babel's` (would use capital when it's not the module itself)

hzoo: Who is we? I think there are multiple we's in the post that could be made clear

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
```diff
@@ -0,0 +1,153 @@
...
+
+We were working under a compressed timeline for this, as we needed to introduce any breaking changes to babel, `@babel/parser`, as well as the pipeline proposal plugin before babel@7 left beta. We also wanted the plugin to be able to eventually default to whichever of the proposals get accepted, so the configuration option becomes obsolete. Given these two constraints, we opted to introduce this new configuration option and make it required, forcing users to decide which proposal they want to use in their project.
+
+Because the pipeline proposal is currently stage-1, both the `stage-1` & `stage-0` plugins include the pipeline operator. An option was added to both of those plugins, `"pipelineProposal"`, to forward the value to the pipeline plugin. This option was also required, and thus breaking, so it caused some frustration for some users. The stage presets have been deprecated and [are being removed](https://github.com/babel/babel/pull/8293), because when those presets are used, consumers often don't know what plugins they're including, so changes like this are surprising and frustrating.
```
hzoo: Yeah in this section we can explain this is the approach we'd like to take moving forward for any kind of proposal that doesn't have a clear semantics moving forward (like Stage 0-2). Where we might force an opt-in to an option vs. some default since that implies it's ready.

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
```diff
@@ -0,0 +1,153 @@
...
+
+Once a particular proposal gets accepted as the canonical behavior of the pipeline operator, we'll deprecate the `"proposal"` option and make the default whichever gets accepted, and the rejected proposals will be removed in the following major version.
+
+## Get Involved
+
+If you're interested in getting involved in the pipeline proposal, all the conversations are public and can be found in the [pipeline operator repository](https://github.com/tc39/proposal-pipeline-operator/). You can also check out the [presentation from the last TC39 meeting](https://docs.google.com/presentation/d/1eFFRK1wLIazIuK0F6fY974OIDvvWXS890XAMB59PUBA/edit#slide=id.p). Lastly, you can find [James DiGioia](https://twitter.com/jamesdigioia), [J. S. Choi](https://twitter.com/__jschoi), or [Daniel Ehrenberg](https://twitter.com/littledan) on Twitter.
+
+But most importantly, once the work is complete, try out the pipeline operator in your own projects! We need feedback and real code in order for this to be useful, so we'd love to hear from you.
```
hzoo: We can mention here we'll be updating the REPL so that you can test each proposal there as options, and then you can experiment online and share/provide feedback. Also can just say you can ping the `@babeljs` twitter maybe as well.

hzoo:
cc @gilbert, @zenparsing, @js-choi, @littledan if ya'll have any comments (reviewing the post or if we missed some context)

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
```diff
@@ -0,0 +1,153 @@
+---
+layout: post
+title:  "What's the Deal with the New Pipeline Plugin Configuration?"
```
hzoo: > What's the Deal with the New Pipeline Plugin Configuration

We could just say it's more about the state of the proposal?

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
littledan: So happy about this approach and creating good documentation for it!

mAAdhaTTah:
@hzoo Thanks! I'll take a look at your comments this evening and make the suggested changes.

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
mAAdhaTTah: I suppose "we" here could be @js-choi and I, specifically, since we're the one's doing the implementation, but it's kind of intentionally ambiguous because there are a lot of people involved in this, and some of them are involved in both TC39 & Babel, so it's not really clear who the "we" actually _is_ here as a group / label.

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
```diff
@@ -0,0 +1,153 @@
...
+  |> doubleSay(#, ', ')
+  |> # || throw new TypeError()
+  |> capitalize
+  |> # + '!'
+  |> new User.Message(#)
+  |> await stream.write(#)
+  |> console.log;
+```
+
+In ["topic style"](https://github.com/js-choi/proposal-smart-pipelines/blob/master/readme.md#topic-style), a placeholder (called a "lexical topic token") is required, and the code will throw an early SyntaxError if it is not included.
```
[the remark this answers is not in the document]
mAAdhaTTah: I think mentioning those names is important for the conversation around Smart Pipelines, so I'd like to get that out there. I cleaned up this section so it's clearer.

mAAdhaTTah:
Updated!

website/blog/2018-07-14-whats-the-deal-with-the-new-pipeline-plugin-configuration.md
hzoo: I just thought it might get confusing with `we approached them` and stuff

mAAdhaTTah: You're right, maybe I rephrase it as: "Given Babel's role in the community, we decided to include all three proposals..." so it reads more like it's the same group? It's still a bit weird, but maybe better?

mAAdhaTTah: Ok, made this change.

website/blog/2018-07-18-whats-happening-with-the-pipeline-proposal.md
```diff
@@ -3,7 +3,7 @@ layout: post
 title:  "What's Happening With the Pipeline (|>) Proposal?"
 author: James DiGioia
 authorURL: https://twitter.com/JamesDiGioia
-date: 2018-07-18 12:00:00
+date: 2018-07-19 12:00:00
```
mAAdhaTTah: The filename also needs to be changed!

website/blog/2018-07-19-whats-happening-with-the-pipeline-proposal.md
```diff
@@ -0,0 +1,168 @@
...
+layout: post
+title:  "What's Happening With the Pipeline (|>) Proposal?"
+author: James DiGioia
+authorURL: https://twitter.com/JamesDiGioia
+date: 2018-07-19 12:00:00
+categories: announcements
+share_text: "What's Happening With the Pipeline (|>) Proposal?"
+---
+
+With the release of babel@7.0.0-beta54, we [introduced](https://github.com/babel/babel/pull/3159) a new required configuration flag to `@babel/plugin-proposal-pipeline-operator`, a breaking change not for the pipeline operator. To clear up the confusion, let's take a look at the pipeline proposal and why we needed to introduce this configuration option.
```
hzoo: Oh this link is going to the original PR for pipeline. I would move this to line 17? And replace this this the release notes for https://github.com/babel/babel/releases/tag/v7.0.0-beta.52?

website/blog/2018-07-19-whats-happening-with-the-pipeline-proposal.md
```diff
@@ -0,0 +1,168 @@
...
+
+Because these proposals parse slightly differently, support needs to be added to `@babel/parser` (formerly `babylon`), and the parser needs to be configured based on which proposal is its current parse target. The pipeline operator plugin itself thus needs the `"proposal"` option, both to configure babylon as well as its own transformation.
+
+We were working under a compressed timeline for this, as we needed to introduce any breaking changes to babel, `@babel/parser`, as well as the pipeline proposal plugin before babel@7 left beta. We also wanted the plugin to be able to eventually default to whichever of the proposals get accepted, so the configuration option becomes obsolete.
+
+Given these two constraints, we opted to introduce this new configuration option and make it required, forcing users to decide which proposal they want to use in their project. Once a particular proposal gets accepted as the canonical behavior of the pipeline operator, we'll deprecate the `"proposal"` option and make the default whichever gets accepted, and the rejected proposals will be removed in the following major version.
```
hzoo: Should we show the example error message here? Probably not necessary but wondering

website/blog/2018-07-19-whats-happening-with-the-pipeline-proposal.md
mAAdhaTTah: Oh, whoops, I copied your link from a comment without checking it. I intended to link to the PR that introduced the change. We could link to the release notes, altho we are saying "with the release of beta54", and then linking to "beta.52"?

website/blog/2018-07-19-whats-happening-with-the-pipeline-proposal.md
mAAdhaTTah: I don't think that's necessary. The error message should stand on its own.

PR #1736, opened 2018-07-27
Add time travel panel in REPL

sujinleeme:
### edit by @hzoo, test here: https://deploy-preview-1736--babel.netlify.com/repl

Hello, We are the team ☀️sunshine (@gyujincho) on RGSoC 2018.
As @hzoo mentioned before, we added new feature, **time travel interface** for babel transformations as following @boopathi 's [babel-time-travel](https://github.com/babel/babel-time-travel/) work.
screen shot 2018-07-27 at 12 30 12 pm

/cc @boopathi @loganfsmyth @existentialism


hzoo:
Amazing! Tested it online at https://deploy-preview-1736--babel.netlify.com/repl. Basically already works which is awesome.

Already thinking of some ways we can improve on the old design haha (like I wonder if we can switch "foreign" for the actual name, and then show the current AST state, or list out the visitors), but we can chat about that in another meeting. Going to 😴now though

sujinleeme:
@xtuc @hzoo 
If there is 'time travel' button in setting, it would be more cool.
Current version is that time travel slider is default view and there isn't an option to turn on/off.
How do you think?

screen shot 2018-08-06 at 9 44 28 pm

sujinleeme:
@hzoo 

[ ] Time travel check box is added. 
[ ] `timeTravel=true` is updated in url.

http://localhost:3000/repl#?babili=false&browsers=&build=&builtIns=false&spec=true&loose=true&code_lz=FDI&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=true&sourceType=module&lineWrap=false&presets=&prettier=false&targets=&version=6.26.0&envVersion=

gyujincho:
- [ ] Figure out plugin name (foreign$1)
- [ ] Add share button
- [ ] Figure out if we need to "explode" visitors that got merged (https://github.com/babel/babel/blob/8c65230258fd8dbda075f178f4843eeb1971c118/packages/babel-traverse/src/visitors.js#L179, https://github.com/babel/babel/blob/8c65230258fd8dbda075f178f4843eeb1971c118/packages/babel-core/src/transformation/index.js#L105)
- [ ] Handle mouse events
- [ ] Show diff between visitors?
- [ ] Output mode (output code, ast, time travel), new tab?
- [ ] Console panel
- [ ] Research what do people already do for time travel/ast
- [ ] Auto install/config generator from repl
- [ ] Track Repl usage (plugin clicks)
- [ ] Redirect url (path/ -> path)

xtuc:
I think that most of the features you listed can be done in other PRs, we can merge this first and iterate on it. What do you think?

existentialism:
@sujinleeme should we add a brighter border between each transition? I worry that people will miss that there's multiple steps you can click on:

image

Maybe something like:

image

js/repl/PluginConfig.js
```diff
@@ -116,6 +116,7 @@ const replDefaults: ReplState = {
   debug: false,
   evaluate: false,
   fileSize: false,
+  timeTravel: false,
```
existentialism: Can we leave a comment describing the difference between `timeTravel` and `showTimeTravel`?

js/repl/TimeTravelSlider.js
```diff
@@ -0,0 +1,125 @@
...
+  componentWillUnmount = () => {
+    this._isMounted = false;
+  };
+
+  render() {
+    const { transitions, currentTransition } = this.props;
+    return (
+      <div className={styles.sliderWrapper}>
+        <div className={styles.sliderRow}>
+          {transitions &&
```
existentialism: Nit: Wondering if we need to check if transitions is truthy, since the `Props` above state that it can't?

js/repl/Repl.js
```diff
@@ -222,26 +229,40 @@ class Repl extends React.Component<Props, State> {
...
-            code={state.compiled}
-            errorMessage={state.evalErrorMessage}
-            fileSize={state.meta.compiledSize}
-            info={state.debugEnvPreset ? state.envPresetDebugInfo : null}
-            options={options}
-            placeholder="Compiled output will be shown here"
-          />
+        <div className={styles.wrapperPanels}>
+          <div
+            className={[
```
existentialism: I think this can be slightly cleaner with emotion's `cx`:

```jsx
import { cs } from 'emotion';

<div className={cx(styles.panels, !state.timeTravel && styles.panelsMax)} />
```

existentialism:
@sujinleeme @gyujincho fantastic work! :)

js/repl/TimeTravelSlider.js
gyujincho: @existentialism sounds right, removed checking

js/repl/PluginConfig.js
gyujincho: @existentialism removed `showTimeTravel` :)

js/repl/Repl.js
gyujincho: @existentialism changed to use `cx`!

existentialism:
@sujinleeme howabout we land this for now and we can follow it up with more changes (including tweaking the design)?

sujinleeme:
@existentialism That's good idea. 🙏

sujinleeme:
@existentialism We have done! 👍

PR #1738, opened 2018-08-01
Update core.md

tbranyen:
Expands `etc` in `filename` description to contain the fact that it enables babel config lookups for source code transformations.

Reference: https://github.com/babel/babel/issues/6296#issuecomment-409758177


hzoo:
cool, thanks!

PR #1740, opened 2018-08-03
Hide big footer in REPL

sujinleeme:
Hello.  I removed big footer on the REPL and fixed cutting off babel version tab in the bottom of left panel.

screen shot 2018-08-03 at 8 38 39 pm


existentialism:
@sujinleeme looks good!

One thing about the version on the sidebar, I think the original intent was to always keep it visible, with this PR, if the plugin content overflows, the version scrolls with it.

untitled-1

sujinleeme:
What about this version? The version tab doesn't overridden always.
ezgif com-video-to-gif

existentialism:
@sujinleeme great work :)

PR #1744, opened 2018-08-08
Change babili to minify in url

sujinleeme:
Hello. 

I changed `https://babeljs.io/en/repl#?babili=true` to `https://babeljs.io/en/repl#?minify=true` not `https://babeljs.io/en/repl#?babili=true` and add some comments for future work when `babel-minify-standalone` is ready to apply.`//TODO switch to babel-minify-standalone.`

### DIFF
--- a/js/repl/PluginConfig.js
+++ b/js/repl/PluginConfig.js
@@ -105,7 +105,7 @@ const pluginConfigs: Array<PluginConfig> = [
 ];
 
 const replDefaults: ReplState = {
-  babili: false,
+  minify: false,
   browsers: "",
   build: "",
   builtIns: false,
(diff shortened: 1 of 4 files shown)


sujinleeme:
Yes. I just changed url literally. :)

xtuc:
What happens with existing URLs? I think that we should maintain the backward compatibility by supporting the old param.

sujinleeme:
@xtuc There is minify opition in left settings panel, however, the url points to babili. Therefore, user can be confused it.

xtuc:
Yes, I meant to support both babili and minify options. That will avoid breaking the existing URLs.

hzoo:
Would the urls still work but just not be minified? That's not too bad in my opinion.

PR #1747, opened 2018-08-10
Add missing parenthesis to template doc

vikr01:
There was a parenthesis missing in the template doc when calling the `template` function.


nicolo-ribaudo:
Actually, that example shows how to use the `template` function as a template tag. Can you remove `)` instead of adding `(`?

vikr01:
Got it, fixed.

nicolo-ribaudo:
Thank you!

PR #1748, opened 2018-08-10
WIP: Fix code-block-buttons.js to work in IE11

ianchobuilds:
- Transpile `code-block-buttons.js` with babel and include `babel-polyfill`
- Minify bundled scripts using `babel-minify-webpack-plugin`

WIP because not tested in an actual IE11 browser yet(currently don't have one, will test with Netlify link in a few days)

### DIFF
--- a/js/code-blocks-buttons.js
+++ b/js/code-blocks-buttons.js
@@ -1,3 +1,5 @@
+import "babel-polyfill";
+
 /* global ClipboardJS */
 /* eslint quotes: ["error", "double", { "avoidEscape": true }] */
 window.addEventListener("load", function() {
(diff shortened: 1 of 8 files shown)


existentialism:
@gyujincho just took at look at this in IE... I think we can avoid transpiling, as the only thing that seems to be breaking is:

```js
document.querySelectorAll(codeBlockSelector).forEach(...)
```

We can probably just copy/paste a polyfill (like [this one](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach)).

(Also looks like we need to fix the css for the copy button for IE11)

danez:
Even if we fix this one file, there are more issues in order to get stuff working in IE11. Foe example docusaurus ships template strings in some files which get directly sent to browser.

https://github.com/facebook/Docusaurus/blob/master/packages/docusaurus-1.x/lib/static/js/codetabs.js

It is probably not worth fixing imho.

PR #1751, opened 2018-08-13
Fix v6 links to presets

likerRr:


likerRr:
I also think it's a better solution. Didn't know we could do so. But that also means we have to change all the remaining links, like [this](https://github.com/babel/website/blame/master/docs/preset-env.md#L109), am I right?

microbouji:
Yep, that could be a different PR though to fix all absolute links that can be converted to this internal doc style.

likerRr:
@microbouji huh, it was a slight delay before I finished changes :smile: Can you please check it? The only thing I still don't understand, why the deploy [preview deploy-preview-1751--babel.netlify.com](https://deploy-preview-1751--babel.netlify.com/docs/en/6.26.3/babel-preset-env) doesn't convert `*.md` links to urls?

microbouji:
Ah, there's currently a docusaurus [bug](https://github.com/facebook/Docusaurus/issues/845) causing that for renamed docs. The affected links on this page are `preset-latest.md`, `plugin-transform-async-generator-functions.md`, and `plugin-transform-object-rest-spread.md` which we have renamed/deleted in the latest version in the docs folder. If they were working for you locally, you probably still had these files in the docs folder.

I guess for now we'll have to go back to absolute urls for these docs like `https://babeljs.io/docs/en/6.26.3/babel-preset-latest` or just the relative `babel-preset-latest`

Other docs that we haven't renamed like `preset-es2015.md` are working fine.

likerRr:
Oh, that's sad. I think relative links are more preferable, are they? 

Also, am I right that relative link `babel-preset-latest` is the result of adding prefix `babel` to a file `preset-latest.md`? So, e.g. relative link for `plugin-transform-object-rest-spread.md` is `babel-plugin-transform-object-rest-spread`

likerRr:
Finally it looks like it's expected

PR #1769, opened 2018-08-25
Fix Code Block Titles

vikr01:
Noticed all code block titles said `CSS`.

I renamed all class names matching `.hljs.*`to `.hljs.language-*` in `code-blocks.css`, since that matches the class names hljs produces.


suchipi:
Thank you!

PR #1770, opened 2018-08-25
Change package titles to be scoped

### LINKED ISSUE #1768: Change the titles for all the packages
@loganfsmyth pointed out that the titles aren't using scoped packages, so sounds like we need to do this for most of the pages

```diff
-title: babel-preset-env
+title: @babel/preset-env
```

notlmn:
Fixes #1768

Changed all current package titles to be scoped, and fixed some doc issues along the way.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1,6 +1,6 @@
 ---
 id: babel-cli
-title: babel-cli
+title: @babel/cli
 sidebar_label: babel-cli
 ---
 
@@ -17,7 +17,7 @@ There are two primary reasons for this.
 
 1. Different projects on the same machine can depend on different versions of
      Babel allowing you to update them individually.
-2. Not having an implicit dependency on the environment you are working in 
+2. Not having an implicit dependency on the environment you are working in
      makes your project far more portable and easier to setup.
 
 We can install Babel CLI locally by running:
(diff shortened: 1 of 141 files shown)


existentialism:
@notlmn thanks!

nicolo-ribaudo:
We also need to update them in the `tooling` section in the left sidebar at https://deploy-preview-1770--babel.netlify.com/docs/en/next

notlmn:
@nicolo-ribaudo Yeah, I also thought of that, but other parts of the docs have `babel-` prefix, and some don't. I was waiting to ask if the prefix should be removed (in `id` and in `sidebar_label`).

I'll submit a PR when I get back home.

notlmn:
I've hit a wall while modifying the other fields, which may be a bug with Docusaursus, where it mixes up strings of all versions in the same file, causing the sidebar labels to be messed up ([test repo](https://github.com/notlmn-dump/doc-test)).

I've reached out to the people at Docusaurus, the new PR may take a while.

notlmn:
@nicolo-ribaudo The sidebar label change cannot be made as of now because of a bug in Docusaurus. I've opened and [issue](https://github.com/facebook/Docusaurus/issues/916) over at the Docusaurus repo.

The change can happen once the issue is resolved, or a workaround is to use different values for `id` (without the `version-XX` part) and `sidebar_label` in these docs to be different.

nicolo-ribaudo:
Ok, we can wait then

PR #1774, opened 2018-08-27
Add a note about preset-env browserlists behavior

ylemkimon:
`@babel/preset-env` only uses `Android`, `Chrome`, `Edge`, `Firefox`, `IE`, `iOS`, `Safari`, `Node` versions for detecting which transforms to include:

https://github.com/babel/babel/blob/e875fc7bc11ffdb8d5176028ae2b336bf7455c6c/packages/babel-preset-env/src/targets-parser.js#L32-L42

and silently ignores other browsers:

https://github.com/babel/babel/blob/e875fc7bc11ffdb8d5176028ae2b336bf7455c6c/packages/babel-preset-env/src/targets-parser.js#L72-L76

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -41,6 +41,8 @@ For browser- or Electron-based projects, we recommend using a [`.browserslistrc`
 
 By default `@babel/preset-env` will use [browserslist config sources](https://github.com/ai/browserslist#queries) _unless_ either the [targets](#targets) or [ignoreBrowserslistConfig](#ignoreBrowserslistConfig) options are set.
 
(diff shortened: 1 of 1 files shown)


existentialism:
@ylemkimon let's update this section to be more explicit:

> Example environments: chrome, opera, edge, firefox, safari, ie, ios, android, node, electron.

https://github.com/babel/website/blob/master/docs/preset-env.md#targets

ylemkimon:
@existentialism I've move the note to `Targets` section.

ylemkimon:
Now, I realize `opera` is missing from `browserNameMap` in `preset-env`. I'll send a PR to babel/babel.

PR #1787, opened 2018-08-28
Add new options.md and config-files.md files to replace current config-file and core options docs

loganfsmyth:
I think this is a pretty good way to split things up. Let me know what you think!


docs/config-files.md
```diff
@@ -0,0 +1,229 @@
...
+  mod1/
+    package.json
+    src/index.js
+  mod2/
+    package.json
+    src/index.js
+```
+the config will now be entirely ignored, because it is across a package boundary.
+
+One alternative would be to reate a `.babelrc` in each sub-package that uses ["extends"](options.md#extends) as
```
michael-brade: typo: I suppose this should have been "create" instead of "reate"

docs/config-files.md
```diff
@@ -0,0 +1,229 @@
...
+    src/index.js
+  mod2/
+    package.json
+    src/index.js
+```
+and the individual packages are responsible for running their builds (and their working
+directory for Babel is the individual packages), the `babel.config.js` file will not be automatically
+loaded, and users will be required to set the path to it manually.
+
+Project-wide configs can also be disable by setting ["configFile"](options.md#configFile) to `false`.
```
michael-brade: s/disable/disabled/

docs/config-files.md
```diff
@@ -0,0 +1,229 @@
...
+
+## File-relative configuration
+
+Babel loads `.babelrc` (and `.babelrc.js` / `package.json#babel`) files by searching up the
+directory structure starting from the ["filename"](options.md#filename) being compiled. This can
+be powerful because it allows you to create independent configurations for subsections of
+a repository. File-relative configurations are also [merge](options.md#merging) over top of
```
michael-brade: s/[merge]/[merged]/

docs/configuration.md
```diff
@@ -5,7 +5,7 @@ title: Configure Babel
 
 Babel can be configured! Many other tools have similar configs: ESLint (`.eslintrc`), Prettier (`.prettierrc`).
 
-All Babel API [options](core.md#options) are allowed. However if the option requires JavaScript, you may need to use a `.babelrc.js` file.
+All Babel API [options](options.md) are allowed. However if the option requires JavaScript, you may want to use a Javascript [configuration file](config-files.md).
```
michael-brade: IIRC there should be a comma after "However"

docs/config-files.md
```diff
@@ -0,0 +1,228 @@
...
+* `api.cache.invalidate(() => process.env.NODE_ENV)` - Cache based on the value of `NODE_ENV`.
+  Any time the `using` callback returns a value other than the one that was expected, the overall
+  config function will be called again and all entries in the cache will be replaced with the result.
+
+Since the actual callback result is used to check if the cache entry is valid, it is recommended
+that:
+
+* Callbacks should be small and side-effect free.
+* Callbacks should return values with the smallest range possible. For example, the
+  `.using(() => process.env.NODE_ENV)` usage above is not idea because it would create an unknown
```
danielo515: Here is a typo, I think you want to write ideal

docs/config-files.md
```diff
@@ -0,0 +1,228 @@
...
+there is that it makes caching harder. Babel wants to avoid re-executing the
+config function every time a file is compiled, because then it would also need to
+re-execute any plugin and preset functions referenced in that config.
+
+The avoid this, Babel expects users of config functions to tell it how to manage
+caching within a config file.
+
+* `api.cache.forever()` - Permacache the computed config and never call the function again.
+* `api.cache.never()` - Do not cache this config, and re-execute the function every time.
+* `api.cache.using(() => process.env.NODE_ENV)` - Cache based on the value of `NODE_ENV`.
```
danielo515: So any of the cache validation functions receive any argument? That lead them to work with global variables and... Nothing else because scope will not be retained between calls, or will it be? What if you want to validate the cache based on the files that have changed or something like that? Why not provide an argument to this functions?

loganfsmyth: What argument would it provide exactly? The primary usecase for these functions is to check for global flags that should affect config loading, so there isn't generally any state to pass in.

> What if you want to validate the cache based on the files that have changed or something like that?

You can potentially have the cache function do `api.cache.invalidate(() => fs.statSync("somefile").mtime)` to invalidate the config cache when that file changes.

docs/config-files.md
loganfsmyth: Thanks!

docs/config-files.md
danielo515: I thought that babel (watch for example) would keep some kind of report if what has changed and what it is intended to do.

loganfsmyth: Not currently.

PR #1623, opened 2018-04-09
external repl plugin search: use computedKeywords

Haroenv:
same as #1622, but for old-site

PR #1622, opened 2018-04-09
external repl plugin search: use computedKeywords

Haroenv:
The value is still working for both `keywords` and `computedKeywords`, but I want to put the handling of special keywords like babel plugins in there alone and remove the messing with `keywords` in the index.

cc @xtuc

PR #1795, opened 2018-08-29
fix(repl): search on computed plugins

Haroenv:
The plugins are not only those that have the keyword; but also those that start with it. 

see https://github.com/algolia/npm-search/commit/a8a52650cc857e4cfb558538fa5e86b743dc5d86 to the change that did this, 

I already did https://github.com/babel/website/pull/1623 and https://github.com/babel/website/pull/1622, but it seems like that was reverted?

cc @xtuc


Haroenv:
Is this the correct branch for the new site

existentialism:
Probably got inadvertently lost when I redesigned the modal

PR #1796, opened 2018-08-29
feat(repl): avoid browser autocompletion in plugin search

Haroenv:
My browser otherwise interprets this input as a contacts form (it would be better as type=search and in a form, but didn't want to mess with the styling now)

(this is how it used to look)
screen shot 2018-08-29 at 11 35 25 redacted


Haroenv:
👌

PR #1809, opened 2018-09-04
Script to sync published docs with "next" version

microbouji:
I noticed people often send a PR with something they want to change to the live site, but only change the files in the `docs/` folder. This script will remind them before committing to also change the files in `website/versioned_docs/version-7.0.0/`.

To test this, make a change or add a new markdown file to the `docs/` folder and stage it. The script run on precommit will copy the files you changed into the `version-7.0.0/` folder and print the message with info.

That `NEXT_DOCS` env variable is an added complexity/one more thing to remember, but I didn't want to encourage people to just commit with no-verify?

No need to review below that `docusaurusUtils` line as those are helper functions I took verbatim from docusaurus to copy the markdown metadata.

Let me know what you think. It's perfectly fine with me if you don't think this is needed. I don't know how often we'll publish new versions and I thought this might help with keeping them in sync.


JoelMarcey:
cc: @endiliey @yangshun -- just in case we might want to consider providing such a script now or for v2 of Docusaurus.

endiliey:
I would like to bring this up for v2

-------------------

https://docusaurus.io/docs/en/versioning#fallback-functionality

If you've been using Docusaurus for a while, you might notice that the Docusaurus creates versioned docs **only and if only** the docs content are **different.**

Let’s say we have a docs file

## **docs/hello.md**

```md
---
id: hello
title: hello
---
[Installation guide](getting-started-installation.md)
```


And then **we cut version 1.0.0,** Docusaurus will create **`website/versioned_docs/version-1.0.0/hello.md`**

## **website/versioned_docs/version-1.0.0/hello.md**

```md
---
id: version-1.0.0-hello
title: hello
original_id: hello
---
[Installation guide](getting-started-installation.md)
```

However, when we try to cut version 2.0.0 again, Docusaurus will not create `hello.md` again since there is no changes in the doc.

## **website/versioned_docs/version-2.0.0/hello.md** DOES NOT EXIST

# Problem

This is confusing for user, if they want to edit the 2.0.0 docs, they won’t be able to find any **`website/versioned_docs/version-2.0.0/hello.md`.** 

User have to edit the **`website/versioned_docs/version-1.0.0/hello.md`** or manually add **`website/versioned_docs/version-2.0.0/hello.md`.** (See https://github.com/facebook/jest/pull/6758#issuecomment-408274413)

P.S: this could potentially create another bug if we have other versions between `2.0.0` and `next`. 

In addition, this adds complexity on the codebase (for doing all the version fallback). And during the build time, Docusaurus had to replace the linking to the correct version.
This is also the cause of https://github.com/facebook/Docusaurus/issues/845

For v2, I propose that every time we cut docs version, we take a snapshot of all the docs

This is a space complexity trade-offs for better developer & user experience.

endiliey:
Notice that with this PR, if version-7.0.0 and `next` docs are in sync, that means when babel cut new version (e.g: 7.1.0) , there will be no **website/versioned_docs/version-7.1.0/****

This is exactly why I propose above for v2 😄

existentialism:
> This is a space complexity trade-offs for better developer & user experience.

Agreed!

microbouji:
Wow thanks guys! FWIW I like that snapshot idea as well.

As for the use case, I guess the idea is not to keep next and published versions totally in sync, but to allow real version-related changes to be applied only to the next version if the user chooses so, and quick fixes to be retroactively applied to at least the published version. That will require some diffing magic though, not the copy-paste logic that I'm using here.

PR #1812, opened 2018-09-06
Fixing bad links

atomicpages:


hzoo:
I think this should be ok now, thanks for the PR!

PR #1813, opened 2018-09-07
Add BitMidi to users

feross:
This PR adds the website [BitMidi](https://bitmidi.com) to the readme.

BitMidi uses Babel for class property, object rest spread, optional catch binding, ES modules, and JSX support! We don't transpile most features since we only support the latest version of each browser. So our bundle is extremely light – under 50KB!

Website screenshot: 

bitmidi-screenshot

Logo:

bitmidi


existentialism:
@feross thanks! (btw, BitMidi is so cool, brings back a bunch of nostalgia for me!)

PR #1818, opened 2018-09-10
Update ignore files section

sto3psl:
With Babel 7 the `--ignore` flag needs proper patterns to match files. This updates the documentation to reflect that change.

There is some more information in this issue https://github.com/babel/babel/issues/8631.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -102,7 +102,7 @@ npx babel src --out-file script-compiled.js
 Ignore spec and test files
 
 ```sh
-npx babel src --out-dir lib --ignore spec.js,test.js
+npx babel src --out-dir lib --ignore src/**/*.spec.js,src/**/*.test.js
 ```
 
 ### Copy files
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -102,7 +102,7 @@ npx babel src --out-file script-compiled.js
 Ignore spec and test files
 
 ```sh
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -102,7 +102,7 @@ npx babel src --out-file script-compiled.js
 Ignore spec and test files
 
 ```sh
-npx babel src --out-dir lib --ignore spec.js,test.js
+npx babel src --out-dir lib --ignore src/**/*.spec.js,src/**/*.test.js
```
xtuc: Could you please add quotes around the double stars?

Otherwise, zsh will expand it before (I guess other Bashes too):
```sh
$ echo **/*.js
zsh: no matches found: **/*.js

# vs

$ echo "**/*.js"
**/*.js
```

docs/cli.md
```diff
@@ -102,7 +102,7 @@ npx babel src --out-file script-compiled.js
 Ignore spec and test files
 
 ```sh
-npx babel src --out-dir lib --ignore spec.js,test.js
+npx babel src --out-dir lib --ignore "src/**/*.spec.js","src/**/*.test.js"
```
xtuc: `"src/**/*.spec.js,src/**/*.test.js"` works too but that's about the same.

sto3psl: Good to know. Thanks!

PR #1820, opened 2018-09-10
preset-env: update `modules` default value

Hypnosphi:
See https://github.com/babel/babel/pull/8485

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -173,7 +173,7 @@ Enable ["loose" transformations](http://2ality.com/2015/12/babel6-loose-mode.htm
 
 ### `modules`
 
-`"amd" | "umd" | "systemjs" | "commonjs" | "cjs" | false`, defaults to `"commonjs"`.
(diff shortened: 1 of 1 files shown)


existentialism:
What are your thoughts on leaving the default value as-is, and instead, add a section atop it that describes how babel-loader@8 uses the [caller option](http://babeljs.io/docs/en/options#caller) to set it to `false` automatically?

Hypnosphi:
It's just not true that default value is `commonjs`. `modules: 'commonjs'` and absense of `modules` option have different effects (the former won't be overridden by babel-loader)

loganfsmyth:
Maybe we can split the 
```
defaults to `"commonjs"`.
```
into a new section that explains it explicitly?

existentialism:
Yeah, my initial thought was to just call out the fact that it can be different based on caller, and use babel-loader as an example, but I'm fine with replacing the `defaults to` bit completely.

Hypnosphi:
“Allow edits from maintainers” is on, so feel free to reword it

PR #1821, opened 2018-09-11
Fix "WordPress" spelling in the 7.0.0 release post

jsnajdr:
Makes the "P" in "WordPress" capital and changes "WordPress" to "WordPress.com", which is a more precise description of the Calypso frontend: it's part of the commercial service rather than the open source WordPress software itself.


existentialism:
@jsnajdr thanks!

PR #1833, opened 2018-09-18
Adds note to options.md about passing cli options

lukebatchelor:
Super small docs update, but I think it's useful (I've been burned a few times).

Happy to make any changes.


suchipi:
Thank you!

PR #1835, opened 2018-09-20
Update plugins.md

Ore4444:
added link to typescript transform

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -143,6 +143,7 @@ These plugins are in the minify repo.
 - [regenerator](plugin-transform-regenerator.md)
 - [runtime](plugin-transform-runtime.md)
 - [strict-mode](plugin-transform-strict-mode.md)
+- [typescript](plugin-transform-typescript.md)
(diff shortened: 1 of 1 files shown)


Ore4444:
* `preset-typescript` is not listed on this page
* presets in general are not listed on this page
* this page lists transform plugins and `plugin-transform-typescript` is one.

hzoo:
We do have a separate presets page which is linked to at the top and it does link to the typescript preset if that helps

xtuc:
Sorry I didn't realize that this page only list plugins, I don't really mind adding this but i'm just worried that it doesn't really help the user and might confuse them.

hzoo:
Yeah I think we don't really want the user to use the plugin directly anyway that's all

Ore4444:
Yes, but the preset-typescript page doesn't include (or link to) important info like which TS features are not supported by Babel.

hzoo:
ok sure, i noticed we put the flow one there anyway

PR #1849, opened 2018-10-02
Add more Babel talks

thiagoarrais:
Closes #1767

I've tried to dig all the info from the video links. Please do revise.

PR #1854, opened 2018-10-03
Allow multiple authors in videos page

thiagoarrais:
Re: #1849 

This allows displaying two authors for the same talk.

It also changes the data format for videos from yml to js.

This may not be idiomatic React, since I don't have a lot of experience in it. Please point out where I can be more idiomatic.


thiagoarrais:
I think this is ready to be merged. Please let me know if you need squashing or rebasing.

hzoo:
Didn't need to do anything, just time. thanks

PR #1862, opened 2018-10-08
Correction in plugin-proposal-decorators.md

mbrowne:
The `decoratorsBeforeExport` option is actually required, so I think the mention of a default is out-of-date - I removed it. I also moved the `false` example before the `true` version so it's easier to see when reading this for the first time that this option only applies to exports and doesn't make a difference otherwise.


hzoo:
thanks!

PR #1865, opened 2018-10-09
Fix error and info panels to be monospaced in the REPL

Retsam:
Due to a CSS reset (I can't quite track down where it's coming from), `pre` tags aren't automatically styled as monospaced.  This causes the error and info panel to not be monospaced, which is especially annoying since the caret doesn't line up properly for parse errors:

Before:

screen shot 2018-10-09 at 4 33 38 pm

After:

screen shot 2018-10-09 at 4 33 52 pm

### DIFF
--- a/js/repl/CodeMirrorPanel.js
+++ b/js/repl/CodeMirrorPanel.js
@@ -70,6 +70,7 @@ const styles = {
     backgroundColor: colors.errorBackground,
     borderTop: `1px solid ${colors.errorBorder}`,
     color: colors.errorForeground,
+    fontFamily: "monospace",
     ...sharedBoxStyles,
   }),
   info: css({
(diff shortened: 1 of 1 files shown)


js/repl/CodeMirrorPanel.js
```diff
@@ -70,13 +70,15 @@ const styles = {
     backgroundColor: colors.errorBackground,
     borderTop: `1px solid ${colors.errorBorder}`,
     color: colors.errorForeground,
+    fontFamily: "monospace",
```
existentialism: We can add this to `sharedBoxStyles` to avoid needing to add it twice.

Retsam: 👍 I didn't notice that `sharedBoxStyles` is only used for these two boxes.

hzoo:
awesome thanks!

PR #1866, opened 2018-10-10
Fix linting for page files

thiagoarrais:
Hey, folks, I've noticed that the js files under `pages` weren't being linted. I've made this change and now it works for me. Please check if this is useful for anyone else.


hzoo:
thanks

PR #1867, opened 2018-10-10
Updates babel/node docs to include missing flags

lukebatchelor:
The `--config-file` and `--env-name` flags were missing from the `@babel/node` docs pages so I've added them here.

Descriptions taken from the [PR that added them](https://github.com/babel/babel/pull/8010/files).

Doesn't look like any other flags have been added since then either. 👌 

I also updated the 7.1 docs for `options.md`. I recently made a PR that called this out but didn't realise there were versioned docs as well, so I had only updated next. (#1833). 

Give me a yell if that's not right.

Thanks!


suchipi:
Thanks!

PR #1869, opened 2018-10-11
Fix name conflict

SindreSvendby:
I copy notices while i was copy pasting the code that it was declared two times in the scope. 

Now the code is `copy-past` ready :smile:


existentialism:
@SindreSvendby thanks!

PR #1871, opened 2018-10-14
Fix links in preset-env

vikr01:
- Fixed the self link for the `ignoreBrowserslistConfig` option.
- Add a link to `babel-preset-latest` for where it's mentioned


hzoo:
Thanks!

PR #1873, opened 2018-10-16
docs(caveats): update IE transform classes plugin

lucaperret:
As mentioned on the upgrade guide:
> Some of the plugins had `-es3-` or `-es2015-` in the names, but these were unnecessary
https://babeljs.io/docs/en/v7-migration#remove-the-year-from-package-names-blog-2017-12-27-nearing-the-70-releasehtml-renames-drop-the-year-from-the-plugin-name


existentialism:
@lucaperret thanks for the PR!

There's a similar reference [here](https://github.com/babel/website/blame/master/docs/caveats.md#L24), can you fix that one too?

PR #1880, opened 2018-10-30
should update the cli doc to 7.1.0 version

itrufeng:
I have updated `@babel/cli` version to 7.1.0 on setup document, I think it should be kept the same version with 7.1.0, not 6.0.0.


microbouji:
This was fixed in https://github.com/babel/website/pull/1930

PR #1893, opened 2018-11-14
Update new default value

hswolff:
This bit me while upgrading to babel 7 w/ present-env.

It says the default is commonjs, however that's not true. When I turn debug on and I don't specify a modules value it says it defaults to auto.

This adds auto as an option and denotes it being the default value.


existentialism:
@hswolff thanks!

PR #1912, opened 2018-12-08
fix: noted absoluteRuntime option

Aghassi:
Was not specified for plugin-transform-runtime.md. Needed if running babel-loader from a different directory than the current working directory.

See 
https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-runtime/src/index.js#L41
https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-runtime/src/index.js#L130
https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-runtime/src/index.js#L9

### DIFF
--- a/docs/plugin-transform-runtime.md
+++ b/docs/plugin-transform-runtime.md
@@ -62,7 +62,7 @@ With options (and their defaults):
         "corejs": false,
         "helpers": true,
         "regenerator": true,
-        "useESModules": false,
+        "useESModules": false
       }
     ]
   ]
--- a/docs/plugin-transform-runtime.md
+++ b/docs/plugin-transform-runtime.md
@@ -61,7 +61,8 @@ With options (and their defaults):
         "corejs": false,
         "helpers": true,
         "regenerator": true,
-        "useESModules": false
+        "useESModules": false,
+        "absoluteRuntime": false,
       }
     ]
   ]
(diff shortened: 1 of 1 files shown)


docs/plugin-transform-runtime.md
```diff
@@ -61,7 +61,8 @@ With options (and their defaults):
         "corejs": false,
         "helpers": true,
         "regenerator": true,
-        "useESModules": false
+        "useESModules": false,
+        "absoluteRuntime": false,
```
existentialism: Hit: Can we alpha sort these?

docs/plugin-transform-runtime.md
```diff
@@ -154,6 +155,12 @@ export default function(instance, Constructor) {
 }
 ```
 
+### `absoluteRuntime`
+
+`boolean`, defaults to `false`.
+
+When enabled, the loader will get the path of modules being passed into this module using the `cwd` set by the `babel-loader`. For example, this allows you to run this preset via a CLI tool and specify the `cwd` as the root directory of the CLI tool instead of the `process.cwd`. This is useful when your tool may have references to things like `@babel/runtime` and the `process.cwd` doesn't.
```
existentialism: > For example, this allows you to run this preset via a CLI tool

Should this be "a preset"?

docs/plugin-transform-runtime.md
```diff
@@ -154,6 +155,12 @@ export default function(instance, Constructor) {
 }
 ```
 
+### `absoluteRuntime`
+
+`boolean`, defaults to `false`.
```
existentialism: Note this can also be a string

docs/plugin-transform-runtime.md
existentialism: I feel like @loganfsmyth did a fantastic job at describing the behavior of this option in https://github.com/babel/babel/pull/8435, wonder if we can just adapt it here?

existentialism:
Really appreciate the PR, we definitely missed documenting this! :)

docs/plugin-transform-runtime.md
Aghassi: Can you elaborate more? You mean, having a preset called `cli` that just turns this on? If so, I'd be very happy with that.

existentialism: I guess I'm not sure what "this preset" means in this context... this isn't a preset?

existentialism:
@Aghassi thanks!

PR #1914, opened 2018-12-11
Update plugin-proposal-decorators.md

mountainmoon:
Examples writen in old proposal syntax should be updated since Babel 7.1.0 finally supports the new decorators proposal.

### DIFF
--- a/docs/plugin-proposal-decorators.md
+++ b/docs/plugin-proposal-decorators.md
@@ -6,45 +6,118 @@ sidebar_label: proposal-decorators
 
 ## Example
 
-(examples are from proposal)
-
-### Simple class decorator
+(examples are from [proposal][proposal])
 
 ```js
-@annotation
-class MyClass { }
+@defineElement('num-counter')
+class Counter extends HTMLElement {
+  @observed #x = 0;
 
-function annotation(target) {
-   target.annotated = true;
+  @bound
+  #clicked() {
+    this.#x++;
+  }
+
+  constructor() {
+    super();
+    this.onclick = this.#clicked;
+  }
+
+  connectedCallback() { this.render(); }
+
+  @bound
+  render() {
+    this.textContent = this.#x.toString();
+  }
 }
 ```
 
 ### Class decorator
 
 ```js
-@isTestable(true)
-class MyClass { }
+// Define the class as a custom element with the given tag name
+function defineElement(tagName) {
+  // In order for a decorator to take an argument, it takes that argument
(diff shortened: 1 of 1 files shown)


docs/plugin-proposal-decorators.md
```diff
@@ -6,45 +6,118 @@ sidebar_label: proposal-decorators
...
-
-### Simple class decorator
+(examples are from [proposal][proposal])
 
 ```js
-@annotation
-class MyClass { }
+@defineElement('num-counter')
+class Counter extends HTMLElement {
+  @observed #x = 0;
```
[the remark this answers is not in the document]
mountainmoon: Updated.

docs/plugin-proposal-decorators.md
```diff
@@ -41,13 +63,20 @@ class C {
...
-     descriptor.enumerable = value;
-     return descriptor;
-  }
+  return function(elementDescriptor) {
+    return {
+      ...elementDescriptor,
+      descriptor: {
+        ...elementDescriptor.descriptor,
+        enumerable: value
+      }
```
nicolo-ribaudo: All this function can be simplified to
```js
return function(elementDescriptor) {
  elementDescriptor.descriptor.enumerable = value;
};
```

mountainmoon: I just noticed the examples of the latest proposal were wroten in immutable style. But it looks like not necessary.
So the others can also be simplified to
```js
function annotation(classDescriptor) {
  classDescriptor.elements.push({
    kind: 'field',
    key: 'annotated',
    placement: 'static',
    descriptor: {},
    initializer: () => true,
  });
}

function isTestable(value) {
  return function(classDescriptor) {
    classDescriptor.elements.push({
      kind: 'field',
      key: 'isTestable',
      placement: 'static',
      descriptor: {},
      initializer: () => value,
    });
  };
}
```
It works in Babel 7.1.0. However, the [proposal](https://tc39.github.io/proposal-decorators/#sec-decorator-functions) said a decorator function returns a decorator. So should I add a return ?
```js
return function(elementDescriptor) {
  elementDescriptor.descriptor.enumerable = value;
  return elementDescriptor
};
```

nicolo-ribaudo: If a decorator returns nothing, it returns the original descriptor (probably that introduction needs to be updated, in you want to open a PR).

I'd like to have examples in both styles, to show different ways of using decorators.

PR #1915, opened 2018-12-11
Update repl to Babel 7

### LINKED ISSUE #1901: REPL still on 6.26.0
The 'Try it out' link takes you to https://babeljs.io/repl which is still at v6.26; most recent plugins require v7 (ex optional chaining).

https://babeljs.io/repl/build/master/ is already at 7.1.6.

nicolo-ribaudo:
- Remove separate handling of Babel and preset-env version, since
  they are both 7 now
- Remove code for the "debug" option, which was already disabled
  for v7

(Hide whitespace changes when reviewing)

Fixes #1901


js/minirepl.js
```diff
@@ -109,7 +109,7 @@ function compileCode(sourceEditor, targetEditor) {
 
   try {
     transformed = Babel.transform(sourceEditor.getValue(), {
-      presets: ["latest"],
+      presets: ["es2015", "es2016", "es2017"],
```
hzoo: we should would want to get preset-env to be default eventually right?

nicolo-ribaudo: I didn't want to load two different bundles just for the minirepl.
I think that, since it is just a demo, it is ok like this?

hzoo: totally didn't see it was the minirepl, makes sense!

PR #1916, opened 2018-12-11
Add link to `babeljs.io/env` reference

gcrsaldanha:
Docs tells the user to go to `babeljs.io/env` but does not provide a clickable link.

Note: `https://babeljs.io/env` actually redirects to `https://babeljs.io/docs/en/env/`, but I believe it's the expected behavior.


nicolo-ribaudo:
Thank you!

PR #1920, opened 2018-12-18
update docs: flow-comment requireDirective

tanhauhau:


existentialism:
Just so I (and other team members) don't forget, this is applicable if/when  https://github.com/babel/babel/pull/9202 lands.

PR #1933, opened 2019-01-01
AST Explorer

sujinleeme:
Happy new year! 🎊
It is a very basic concept of AST panel as following as [AST explorer](http://forivall.com/astexplorer/).

Except for an auto focusing option in AST explorer, on/off methods, empty keys and location data features are completed.

ast

I appreciate your review and help in advance.

Related Issue : [Babel Roadmap - AST Explorer](https://babeljs.io/docs/en/roadmap#astexplorer)

### DIFF
--- a/js/repl/ASTPanel.js
+++ b/js/repl/ASTPanel.js
@@ -51,12 +51,7 @@ export default class ASTPanel extends React.Component<Props, State> {
   static getDerivedStateFromProps(nextProps: Props, prevState: State) {
     if (nextProps.src && nextProps.src !== prevState.src) {
       let flattenSrc = flatten(nextProps.src);
-      const deleteKeys = Object.keys({
-        ...filterFlatten(flattenSrc, "_fromTemplate"),
-        ...filterFlatten(flattenSrc, "_letDone"),
-      });
-
-      flattenSrc = reject(flattenSrc, deleteKeys);
+      flattenSrc = reject(flattenSrc, "_");
       const src = unflatten(flattenSrc);
 
       return {
@@ -172,11 +167,10 @@ const styles = {
   settingsLabel: css({
     alignItems: "center",
     display: "flex",
-    flex: "0 0 1.5rem",
     flexDirection: "colum",
     fontSize: "0.875rem",
     fontWeight: "normal",
-    padding: "0 1rem",
+    padding: "0.19rem 1rem",
     transition: "background-color 250ms ease-in-out, color 250ms ease-in-out",
     "&:hover": {
       backgroundColor: colors.inverseBackgroundDark,
--- a/js/repl/ASTUtils.js
+++ b/js/repl/ASTUtils.js
@@ -135,13 +135,24 @@ function mergeFlatten(currentSrc, nextSrc) {
   };
 }
 
-function reject(obj, keys) {
-  const result = Object.keys(obj)
-    .filter(k => !keys.includes(k))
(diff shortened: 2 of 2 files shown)


xtuc:
[Preview](https://deploy-preview-1933--babel.netlify.com/repl#?ast=true&babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDYAbOEIwIIwN4C-QA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0&envVersion=)

That looks good. Would be cool to be able to click on the code and that it shows the location in the AST, as ASTExplorer.

Also, do you have an idea how we could remove our internal properties in the AST? (`_fromTemplate`, `_letDone` etc).

namuwikilover:
우연히 웹서핑하다가 보게됐는데 바벨 컨트리뷰터라니 존경합니다ㄷㄷ

nicolo-ribaudo:
We should remove every property prefixed with _

danez:
replaces #1421

nicolo-ribaudo:
The output looks good (I haven't read the code yet).
Another possible improvement is to show `type` always as the first property since it is the most important one, or to hide it and print something like
```js
BinaryExpression {
	"left": NumericLiteral {
		"value":1
	},
	"operator": "+",
	"right": NumericLiteral {
		"value":1
	}
}
```

js/repl/compile.js
```diff
@@ -137,6 +138,8 @@ export default function compile(code: string, config: CompileConfig): Return {
     };
 
     const transformed = Babel.transform(code, babelConfig);
+
+    astContext = JSON.stringify(transformed.ast);
```
danez: The problem with this is the same as also with #1421. The ast is for the transformed output which is confusing imho. So for example if the input is `() => {}` and preset-env is enabled there is no ArrowFunction in the ast instead there is a FunctionExpression.

I wonder if there is a usecase were people want to see the transformed output, but atm I think it is more confusing. So I would prefer doing `Babel.parse` here when the ast panel is active (and also disabling all transform options in the interface maybe?)

But not sure what other people think?

js/repl/ASTPanel.js
```diff
@@ -0,0 +1,192 @@
...
+  };
+
+  static getDerivedStateFromProps(nextProps: Props, prevState: State) {
+    if (nextProps.src && nextProps.src !== prevState.src) {
+      let flattenSrc = flatten(nextProps.src);
+      flattenSrc = reject(flattenSrc, "_");
+      const src = unflatten(flattenSrc);
+
+      return {
+        src: src,
```
danez: When the input is changed (`props.src`) the options are ignored. So I disabled `location` but as soon as I change the input the locations are back. I have to enable and then disable `locations` again to remove locations.

danez:
There also seems to be a lot of unrelated changes to `docs` and `versioned_docs`, they should be reverted.

js/repl/compile.js
nicolo-ribaudo: I agree :+1:
Maybe we could move the "AST" option to the top and make it a dropdown?

schermata da 2019-02-10 00-05-33

PR #1936, opened 2019-01-08
Update preset-env.md

loveky:
since we no longer ship yearly presets since 2018, maybe we should update the related doc to prevent confusions in the future.


hzoo:
fine with that! thanks!

PR #1937, opened 2019-01-08
update sponsors, modify view

hzoo:
Could be better but the current site takes too much space anyway? Should do a separate donator page as well (feel free to modify though)

image

### DIFF
--- a/website/pages/en/index.js
+++ b/website/pages/en/index.js
@@ -44,13 +44,6 @@ const MiniRepl = ({ language }) => {
           <div className="hero-repl__error" />
         </div>
       </div>
-      <div className="hero-repl__footer">
-        <a href={siteConfig.getPageUrl("repl.html", language)}>
-          <translate>
-            Check out our REPL to experiment more with Babel!
-          </translate>
-        </a>
-      </div>
 
       <script
         src="https://unpkg.com/babel-standalone@6/babel.min.js"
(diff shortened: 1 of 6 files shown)


website/pages/en/index.js
```diff
@@ -297,6 +269,20 @@ const Hero = ({ language }) => (
       </div>
 
       <MiniRepl language={language} />
+
+      <h3>Special Sponsor</h3>
+      <div class="sponsors-tier">
+        <a href="https://www.handshake.org" title="Handshake" target="_blank">
+          <img
+            src="https://handshake.org/images/landing/logo-dark.svg"
```
nicolo-ribaudo: I slightly prefer `https://handshake.org/images/landing/logo-light.svg`

website/pages/en/index.js
```diff
@@ -230,29 +203,28 @@ const OpenCollectiveSponsors = ({ language }) => {
...
             title="Gold Sponsors (Open Collective)"
             tier="gold-sponsors"
-            button={ocButton}
           />
           <SponsorTier
-            type="other"
-            title="Misc Sponsors"
-            tier="other-sponsors"
+            type="opencollective"
+            title="Silver Sponsors (Open Collective)"
```
nicolo-ribaudo: Is this title shown anywhere? If not, we could maybe do something like this: (it's in the "title" of the link, that is shown on hover)

schermata da 2019-01-08 23-40-56

website/pages/en/index.js
hzoo: didn't think of that!

hzoo:
Merging for now, incremental steps!

Issue #1934, opened 2019-01-08
@babel/traverse documentation needs improvement


oyilmaztekin:
We can use the `traverse` method with the `enter` method and also the object property name like `FunctionDeclaration` regarding **@babel/types**. There is no information about this usage in this repo besides babel-handbook that you refer. 

```javascript
traverse(ast, {
    FunctionDeclaration: function(path) {
             path.node.id.name = "x";
    }
})
```
I want to add this usage and its explanation if you allow.

https://github.com/babel/website/blob/master/docs/traverse.md

suchipi:
Yes please!

oyilmaztekin:
I created the [**PR**](https://github.com/babel/website/pull/1938) for this issue but I cannot see changes at the [deploy preview](https://deploy-preview-1938--babel.netlify.com). Have I done something wrong?

suchipi:
Hmm, I don't know why it's not working :( someone else might

oyilmaztekin:
the traverse.MD file in versioned_docs/version-7.0.0 has been changed too. Now problem solved. 🙂

PR #1938, opened 2019-01-09
improved @babel/traverse docs

oyilmaztekin:
This is an improvement about @babel/traverse method. An alternative usage has been added to the doc.
[babel/website#1934](https://github.com/babel/website/issues/1934)


website/versioned_docs/version-7.0.0/traverse.md
```diff
@@ -33,5 +33,17 @@ traverse(ast, {
   }
 });
 ```
+
+Also, we can target particular [**node types**](https://babeljs.io/docs/en/babel-types#api) in the Syntax Tree
+
+```js
+traverse(ast, {
+    FunctionDeclaration: function(path) {
+             path.node.name = "x";
```
suchipi: FunctionDeclaration nodes don't have a `name` property; could you change this to Identifier or change the code?

oyilmaztekin: Sorry. Id property of the FunctionDecleration node has a name.  I missed. I'm on it.

oyilmaztekin: I fixed it. Could you please check...

website/versioned_docs/version-7.1.0/traverse.md
```diff
@@ -0,0 +1,48 @@
+---
```
suchipi: You don't need to add this; all versions of 7.x fall back to the 7.0 docs if there's no file in this directory

PR #1941, opened 2019-01-09
Add `api.cache( true ) to example configuration

soulofmischief:
Babel requires a cache setting, so the example config is updated to reflect that.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -27,7 +27,10 @@ All Babel API [options](options.md) are allowed. However, if the option requires
 Create a file called `babel.config.js` with the following content at the root of your project (where the `package.json` is).
 
 ```js
-module.exports = function () {
+module.exports = function ( api ) {
+  // Cache returned configuration.
+  api.cache( true )
+
   const presets = [ ... ];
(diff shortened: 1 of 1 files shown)


suchipi:
Babel doesn't *require* a cache setting- could you explain this a bit further?

I don't think everyone needs this in their config, so it's kinda noise. Maybe this should be documented somewhere else, or in a different way?

soulofmischief:
@suchipi I was unable to use the `babel.config.js` example provided without a call to `api.cache()`. babel-cli explained I needed to set the cache settings and provided a few examples on how to do it.

However I think @loganfsmyth is correct, a default of `api.cache( true )` makes sense.

halfninja:
This doc got changed in #1950 to reflect the current reality - I'd also be fine if Babel changed to have a sensible default, at which point the docs could be changed back.

halfninja:
Just a timing issue I imagine, who happened to be paying attention to incoming PRs at the time. Your change looks just as good, I wouldn't take it personally 🙂

soulofmischief:
Nothing personal, just taking note for future commits. I'll close this now.

PR #1944, opened 2019-01-10
Added Taskade to users list

johnxie:
Added Taskade (https://taskade.com) to users list.


johnxie:
We use Babel at [Taskade](https://taskade.com) to transpile both backend and frontend codes so that we can use the latest ES.

PR #1947, opened 2019-01-15
Fix babel-plugin-module-resolver info in v7 migration api notes

paweljedrzejczyk:
Linked PR states that `babel-plugin-module-resolver@3` should be used, not `@babel/plugin-module-resolver` which haven't been released yet.


existentialism:
@paweljedrzejczyk thanks!

PR #1950, opened 2019-01-16
Update configuration.md

halfninja:
Babel fails if you don't at least configure `api.cache`, so updated the example.


hzoo:
Thanks!

suchipi:
@hzoo what does `api.cache(true)` do? I didn't think it was needed?

hzoo:
Ah I see it's the same as https://github.com/babel/website/pull/1941, I thought it would fail if you didn't set it in a js file https://github.com/babel/babel/blob/bd0c62dc0c30cf16a4d4ef0ddf21d386f673815c/packages/babel-core/src/config/files/configuration.js#L292-L299 for some reason

PR #1952, opened 2019-01-17
Fix logic error in plugin-transform-strict-mode.md

ismailarilik:
- `strictMode` is not an option of `@babel/plugin-transform-modules-commonjs` (https://bit.ly/2W0gv15).
`strict` defaults to `false` so to disable, no action is required (https://bit.ly/2VYMLSs).


ismailarilik:
Sorry, I made a mistake.

PR #1960, opened 2019-02-07
docs(core): Add DEFAULT_EXTENSIONS export

nicojs:
Add DEFAULT_EXTENSIONS  export to the documented list of features.


nicojs:
If this list is documented we will be able to use it in Stryker, the JavaScript mutation testing framework.

nicojs:
I don't see my changes here: https://deploy-preview-1960--babel.netlify.com/docs/en/babel-core

Dit I do something wrong?

danez:
Sorry for the late reply. The docs also need to be addedd to the 7.0.0 docs folder, the root docs folder is in theory the next version, but we do not make much use of it.

I added it

PR #1965, opened 2019-02-09
docs(core): Fix typo

ope-oguntoye:
~then~ _they_


danez:
Thank you very much for contributing!

PR #1967, opened 2019-02-11
Add missing docs for proposal-private-methods

existentialism:
Fixes https://github.com/babel/website/issues/1966.


nicolo-ribaudo:
It has a loose option, whose value must be the same as the one of the class-properties plugin.

existentialism:
@nicolo-ribaudo @hzoo updated this to include its addition in 7.2.0 and adding accessors in 7.3.0

PR #1971, opened 2019-02-14
Each stage preset enables the more advanced stages

nicolo-ribaudo:
Show all the enabled stages in the UI, and don't enable them twice since it causes problems*

Demo: https://deploy-preview-1971--babel.netlify.com/repl#?presets=

\* Try enabling both the stage-2 and stage-3 presets in the current repl, using legacy decorators: it will throw an error about plugin ordering.


hzoo:
seems like a good compromise until we just remove that in the repl?

PR #1978, opened 2019-03-03
Add params for preset-typescript

tsangint:
https://github.com/babel/babel/blob/master/packages/babel-preset-typescript/src/index.js#L30


nicolo-ribaudo:
I'd prefer to pass something like `filename: "example.ts"` and document that it needs a filename or the allExtension a option.

tsangint:
@nicolo-ribaudo Thanks for your suggestion, I have modified and recommited.

PR #1982, opened 2019-03-07
Minor update to plugin-transform-typescript.md

orta:
Moves the workarounds to their own line (but still inside the `<li>` ) - it makes the section read better.


existentialism:
@orta thanks!

PR #1987, opened 2019-03-14
Update docs for core-js 3

nicolo-ribaudo:
Babel PR: babel/babel#7646

### DIFF
--- a/docs/plugin-transform-runtime.md
+++ b/docs/plugin-transform-runtime.md
@@ -88,27 +88,22 @@ require("@babel/core").transform("code", {
 
 ### `corejs`
 
-`false`, `CorejsVersion` or `{ version: CorejsVersion, proposals: boolean }`, defaults to `false`.
-`CorejsVersion` can be either `2` or `3`.
+`false`, `2`, `3` or `{ version: 2 | 3, proposals: boolean }`, defaults to `false`.
 
 e.g. `['@babel/plugin-transform-runtime', { corejs: 3 }],`
 
-When this option is not false, all polyfillable API usages will be rewritten to reference helpers from `core-js` instead.
-`corejs: 2` only supports global variables (e.g. `Promise`) and static properties (e.g. `Array.from`), while `corejs: 3` also supports instance properties (e.g. `[].includes`).
+Specifying a number will rewrite the helpers that need polyfillable APIs to reference helpers from that (major) version of `core-js` instead
+Please note that `corejs: 2` only supports global variables (e.g. `Promise`) and static properties (e.g. `Array.from`), while `corejs: 3` also supports instance properties (e.g. `[].includes`).
 
-By default, `@babel/plugin-transform-runtime` doesn't polyfill proposals. If you are using `corejs: 3`, you can opt-in using the `proposals: true` boolean flag.
+By default, `@babel/plugin-transform-runtime` doesn't polyfill proposals. If you are using `corejs: 3`, you can opt into this by enabling using the `proposals: true` option.
 
 This option requires changing the dependency used to provide the necessary runtime helpers:
 
-| `corejs` option | Runtime package          |
-|-----------------|--------------------------|
-| `false`         | `@babel/runtime`         |
-| `2`             | `@babel/runtime-corejs2` |
-| `3`             | `@babel/runtime-corejs3` |
-
-```sh
-npm install --save [Runtime package]
-```
+| `corejs` option | Install command                             |
+|-----------------|---------------------------------------------|
+| `false`         | `npm install --save @babel/runtime`         |
+| `2`             | `npm install --save @babel/runtime-corejs2` |
+| `3`             | `npm install --save @babel/runtime-corejs3` |
 
 ### `helpers`
 
--- a/docs/plugin-transform-runtime.md
+++ b/docs/plugin-transform-runtime.md
@@ -88,16 +88,26 @@ require("@babel/core").transform("code", {
 
 ### `corejs`
 
-`boolean` or `number` , defaults to `false`.
+`false`, `CorejsVersion` or `{ version: CorejsVersion, proposals: boolean }`, defaults to `false`.
+`CorejsVersion` can be either `2` or `3`.
 
-e.g. `['@babel/plugin-transform-runtime', { corejs: 2 }],`
+e.g. `['@babel/plugin-transform-runtime', { corejs: 3 }],`
 
-Specifying a number will rewrite the helpers that need polyfillable APIs to reference `core-js` instead.
+When this option is not false, all polyfillable API usages will be rewritten to reference helpers from `core-js` instead.
+`corejs: 2` only supports global variables (e.g. `Promise`) and static properties (e.g. `Array.from`), while `corejs: 3` also supports instance properties (e.g. `[].includes`).
 
-This requires changing the dependency used to be [`@babel/runtime-corejs2`](runtime-corejs2.md) instead of `@babel/runtime`.
+By default, `@babel/plugin-transform-runtime` doesn't polyfill proposals. If you are using `corejs: 3`, you can opt-in using the `proposals: true` boolean flag.
+
+This option requires changing the dependency used to provide the necessary runtime helpers:
+
+| `corejs` option | Runtime package          |
+|-----------------|--------------------------|
+| `false`         | `@babel/runtime`         |
+| `2`             | `@babel/runtime-corejs2` |
+| `3`             | `@babel/runtime-corejs3` |
 
 ```sh
-npm install --save @babel/runtime-corejs2
+npm install --save [Runtime package]
 ```
 
 ### `helpers`
@@ -247,7 +257,9 @@ The plugin transforms the following:
 ```javascript
 var sym = Symbol();
 
-var promise = new Promise();
+var promise = Promise.resolve();
+
+var check = arr.includes("yeah!");
 
 console.log(arr[Symbol.iterator]());
 ```
@@ -255,35 +267,24 @@ console.log(arr[Symbol.iterator]());
 into the following:
 
 ```javascript
-"use strict";
+import _getIterator from "@babel/runtime-corejs3/core-js/get-iterator";
+import _includesInstanceProperty from "@babel/runtime-corejs3/core-js-stable/instance/includes";
+import _Promise from "@babel/runtime-corejs3/core-js-stable/promise";
+import _Symbol from "@babel/runtime-corejs3/core-js-stable/symbol";
 
-var _getIterator2 = require("@babel/runtime-corejs2/core-js/get-iterator");
+var sym = _Symbol();
 
-var _getIterator3 = _interopRequireDefault(_getIterator2);
+var promise = _Promise.resolve();
 
-var _promise = require("@babel/runtime-corejs2/core-js/promise");
+var check = _includesInstanceProperty(arr).call(arr, "yeah!");
 
-var _promise2 = _interopRequireDefault(_promise);
-
-var _symbol = require("@babel/runtime-corejs2/core-js/symbol");
-
-var _symbol2 = _interopRequireDefault(_symbol);
-
-function _interopRequireDefault(obj) {
-  return obj && obj.__esModule ? obj : { default: obj };
-}
-
-var sym = (0, _symbol2.default)();
-
-var promise = new _promise2.default();
-
-console.log((0, _getIterator3.default)(arr));
+console.log(_getIterator(arr));
 ```
 
-This means is that you can seamlessly use these native built-ins and static methods
+This means is that you can seamlessly use these native built-ins and methods
 without worrying about where they come from.
 
-**NOTE:** Instance methods such as `"foobar".includes("foo")` will **not** work.
+**NOTE:** Instance methods such as `"foobar".includes("foo")` will only work when using `corejs: 3`.
 
 ### Helper aliasing
 
--- a/docs/plugin-transform-runtime.md
+++ b/docs/plugin-transform-runtime.md
@@ -6,7 +6,7 @@ sidebar_label: transform-runtime
 
 A plugin that enables the re-use of Babel's injected helper code to save on codesize.
 
-> NOTE: Instance methods such as `"foobar".includes("foo")` will not work since that would require modification of existing built-ins (you can use [`@babel/polyfill`](polyfill.md) for that).
+> NOTE: Instance methods such as `"foobar".includes("foo")` will only work with `core-js@3`. If you need to polyfill them, you can directly import `"core-js"` or use `@babel/preset-env`'s `useBuiltIns` option.
 
 ## Installation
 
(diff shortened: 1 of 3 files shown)


docs/preset-env.md
```diff
@@ -229,38 +229,64 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
 
 `"usage"` | `"entry"` | `false`, defaults to `false`.
 
-> This option adds direct references to the `core-js` module as bare imports. Thus `core-js` will be resolved relative to the file itself and needs to be accessible. You may need to specify `core-js@2` as a top level dependency in your application if there isn't a `core-js` dependency or there are multiple versions.
+> When using this options, you also need to configure the [`corejs`](#corejs) option.
 
 This option configures how `@babel/preset-env` handles polyfills.
 
+This option adds direct references to the `core-js` module as bare imports. For this reason, you need to install `core-js@3` or `core-js@2` (depending on the value of the [`corejs`](#corejs) option) as a dependency of your application:
```
existentialism:
```suggestion
When either the `usage` or `entry` options are used, `@babel-preset-env` will add direct references to `core-js` modules as bare imports (or requires). This means `core-js` will be resolved relative to the file itself and needs to be accessible.

Since `@babel/polyfill` was deprecated in 7.4.0, we recommend directly adding `core-js` and (if necessary), setting the version via the [`corejs`](#corejs) option.
```

docs/preset-env.md
```diff
@@ -229,38 +229,64 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
+```sh
+npm install core-js@3 --save
+```
+
 #### `useBuiltIns: 'entry'`
 
-> NOTE: Only use `require("@babel/polyfill");` once in your whole app.
-> Multiple imports or requires of `@babel/polyfill` will throw an error since it can cause global collisions and other issues that are hard to trace.
-> We recommend creating a single entry file that only contains the `require` statement.
+> NOTE: Only use `import "core-js";` and `import "regenerator-runtime/runtime";` once in your whole app.
```
existentialism: I think we need to rework this section to note that while `@babel/polyfill` is deprecated, it still works?

docs/preset-env.md
```diff
@@ -229,38 +229,64 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
+import "core-js/modules/esnext.math.radians";
+import "core-js/modules/esnext.math.scale";
 ```
 
-This will also work for `core-js` directly (`import "core-js";` or `require('core-js');`)
+> NOTE: When using `core-js@2` (either explicitly using the [`corejs: 2`](#corejs) option or implicitly), `@babel/preset-env` will also imports and requires of `@babel/polyfill`.
+> This behavior is deprecated because it isn't possible to use `@babel/polyfill` with different `core-js` versions.
 
-#### `useBuiltIns: 'usage'` (experimental)
+#### `useBuiltIns: 'usage'`
```
existentialism: As a sidenote, if we're going to remove the experimental tag here, we should enumerate all the caveats of using `usage` (e.g. `includes` doesn't work, etc) as there's lots of issues/confusion here.

docs/preset-env.md
```diff
@@ -302,7 +328,38 @@ var b = new Map();
 
 #### `useBuiltIns: false`
 
-Don't add polyfills automatically per file, or transform `import "@babel/polyfill"` to individual polyfills.
+Don't add polyfills automatically per file, or transform `import "core-js"` to individual polyfills.
```
existentialism: Again, I think we should be careful about wholesale removing `@babel/polyfill` references as we don't want to give the impression that support is completely gone (breaking change).

docs/preset-env.md
```diff
@@ -302,7 +328,38 @@ var b = new Map();
...
+
+### `corejs`
+
+`CorejsVersion` or `{ version: CorejsVersion, proposals: boolean }`, defaults to `2`.
```
existentialism: Thoughts on making this match an earlier edit?


```suggestion
`2|3` or `{ version 2|3, proposals: boolean }`, defaults to `2`.
```

docs/preset-env.md
```diff
@@ -302,7 +328,38 @@ var b = new Map();
...
+`CorejsVersion` can be either `2` or `3`.
+
+This option, which can only be used when `useBuiltIns` is not `false`, configures `@babel/preset-env` to inject imports to the correct `core-js` version.
+By default, only polyfills for stable ECMAScript features are injected: if you want to polyfill them, you need to set `proposals: true`.
+
+`corejs: 2` only supports global variables (e.g. `Promise`) and static properties (e.g. `Array.from`), while `corejs: 3` also supports instance properties (e.g. `[].includes`).
+
+By default, `@babel/plugin-transform-runtime` doesn't polyfill proposals. If you are using `corejs: 3`, you can opt-in using the `proposals: true` boolean flag:
```
existentialism: Seems unnecessary/redundant, we mention this above?

> By default, only polyfills for stable ECMAScript features are injected: if you want to polyfill them, you need to set `proposals: true`.

docs/preset-env.md
```diff
@@ -302,7 +328,38 @@ var b = new Map();
...
 
-Don't add polyfills automatically per file, or transform `import "@babel/polyfill"` to individual polyfills.
+Don't add polyfills automatically per file, or transform `import "core-js"` to individual polyfills.
+
+### `corejs`
+
+`CorejsVersion` or `{ version: CorejsVersion, proposals: boolean }`, defaults to `2`.
+`CorejsVersion` can be either `2` or `3`.
+
+This option, which can only be used when `useBuiltIns` is not `false`, configures `@babel/preset-env` to inject imports to the correct `core-js` version.
```
existentialism:
```suggestion
This option only has an effect when used alongside `useBuiltIns: usage` or `useBuiltIns: entry`, and ensures `@babel/preset-env` injects the correct imports for your `core-js` version.
```

docs/preset-env.md
```diff
@@ -302,7 +328,38 @@ var b = new Map();
...
+By default, only polyfills for stable ECMAScript features are injected: if you want to polyfill them, you need to set `proposals: true`.
```
existentialism: We should explain why `proposals: true` is necessary and is different than `useShippedProposals` (which I'm still fuzzy on, personally).

docs/preset-env.md
nicolo-ribaudo: Well, `includes` will work after the core-js 3 pr

existentialism: Ah, I must have missed that.

Does it fix it for core-js 2 (didn't see that mentioned)? If not, it's even more reason for us to enumerate the caveats depending on the version of core-js being used?

docs/preset-env.md
```diff
@@ -229,38 +229,64 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
 
 `"usage"` | `"entry"` | `false`, defaults to `false`.
 
-> This option adds direct references to the `core-js` module as bare imports. Thus `core-js` will be resolved relative to the file itself and needs to be accessible. You may need to specify `core-js@2` as a top level dependency in your application if there isn't a `core-js` dependency or there are multiple versions.
+> When using this options, you also need to configure the [`corejs`](#corejs) option.
```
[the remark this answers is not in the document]
zloirock: Without this option, the user will see a warning that he should set `corejs` option directly and `corejs: 2` will be used by default, so the current version is correct.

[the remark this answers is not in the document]
nicolo-ribaudo: This is done to match transform-runtime, which requires the corejs option to load polyfills: since the default should always be the last version but it isn't possible without breaking changes, it is better not to have a default (or warn when it is not set).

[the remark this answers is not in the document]
existentialism: >  so the current version is correct

I'm perfectly fine with letting the user know that stuff is changing and they _should_ set the core-js version... I'm just advocating we use slightly softer language to not give the impression that they _must_ set the version immediately, is all.

(FWIW, I personally disagree with _forcing_ the option)

> This is done to match transform-runtime, which requires the corejs option to load polyfills

Not sure I agree we need to exactly match behavior from `transform-runtime`.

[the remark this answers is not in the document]
nicolo-ribaudo: > (FWIW, I personally disagree with requiring the option)

Let's discuss this in the other PR. I'm fine both with requiring or not requiring it.

I can change the wording

docs/plugin-transform-runtime.md
```diff
@@ -88,16 +88,26 @@ require("@babel/core").transform("code", {
...
+e.g. `['@babel/plugin-transform-runtime', { corejs: 3 }],`
 
-Specifying a number will rewrite the helpers that need polyfillable APIs to reference `core-js` instead.
+When this option is not false, all polyfillable API usages will be rewritten to reference helpers from `core-js` instead.
+`corejs: 2` only supports global variables (e.g. `Promise`) and static properties (e.g. `Array.from`), while `corejs: 3` also supports instance properties (e.g. `[].includes`).
 
-This requires changing the dependency used to be [`@babel/runtime-corejs2`](runtime-corejs2.md) instead of `@babel/runtime`.
+By default, `@babel/plugin-transform-runtime` doesn't polyfill proposals. If you are using `corejs: 3`, you can opt-in using the `proposals: true` boolean flag.
+
+This option requires changing the dependency used to provide the necessary runtime helpers:
```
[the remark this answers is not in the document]
nicolo-ribaudo: Yeah it's better: people will only need to copy-paste

docs/preset-env.md
nicolo-ribaudo: I think that we both confused `@babel/preset-env` with `@babel/transform-runtime` :stuck_out_tongue: . Instance methods were already polyfilled when using `core-js@2` [repl](https://babeljs.io/repl/build/master#?babili=false&browsers=&build=&builtIns=usage&spec=false&loose=false&code_lz=GYexDoEsDsGMBsCuATApgZwNxA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env&prettier=false&targets=&version=7.3.4)

Input:
```js
foo.includes;
```

Output:
```js
require("core-js/modules/es7.array.includes");
require("core-js/modules/es6.string.includes");

foo.includes;
```

zloirock: @nicolo-ribaudo with `core-js@2`, `useBuiltIns: usage` had really many limitations and bugs. With `core-js@3` target a serious part of them was fixed and I think that we can remove `experimental` tag.

zloirock: Just some examples:
```js
new Map(); // => '[object Object] - `Object#toString` definition was broken
Object.getOwnPropertySymbols; // => undefined - this definition also was broken

import(something).then(...); // => Error  - `Promise` is not polyfilled in Webpack helpers

global.Set; // => undefined

something.finally;

Promise.resolve().finally; // => most likely undefined because of incorrect polyfilling order

'values' in Object; // => false
```
and many other.

docs/preset-env.md
nicolo-ribaudo: Yeah, I think I accidentally copied these last two sentences from the transform-runtime docs

docs/preset-env.md
existentialism: Sure, it's awesome that `usage` with `core-js@3` is much improved over the previous version, but we need to communicate to users caveats and issues with with both versions as not everyone may have the freedom or luxury to upgrade immediately.

@nicolo-ribaudo I was referring to the the `@babel/preset-env` `includes` option not working with `usage`

docs/plugin-transform-runtime.md
```diff
@@ -88,17 +88,22 @@ require("@babel/core").transform("code", {
 
 ### `corejs`
 
-`boolean` or `number` , defaults to `false`.
+`false`, `2`, `3` or `{ version: 2 | 3, proposals: boolean }`, defaults to `false`.
```
zloirock: At the top of this page a note that `runtime` does not support instance methods, now it's not correct.

Too many links to deprecated `@babel/polyfill` as a global alternative.

This page does not contain a link to `core-js` documentation. However, users should know what they are using. I think that we need to add it at least on the first `core-js` entry.

docs/preset-env.md
```diff
@@ -302,7 +333,18 @@ var b = new Map();
...
+### `corejs`
+
+`2`, `3` or `{ version: 2 | 3, proposals: boolean }`, defaults to `2`.
+
+This option only has an effect when used alongside `useBuiltIns: usage` or `useBuiltIns: entry`, and ensures `@babel/preset-env` injects the correct imports for your `core-js` version.
+
+By default, only polyfills for stable ECMAScript features are injected: if you want to polyfill them, you have three different options:
+- set the [`shippedProposals`](#shippedproposals) option to `true`. This will enable polyfills and transforms for proposal which have already been shipped in browsers for a while.
+- use `corejs: { version: 3, proposals: true }`. This will enable polyfilling of every proposal supported by `core-js`.
+- when using `useBuiltIns: "entry"`, you can directly import a [proposal polyfill](https://github.com/zloirock/core-js/tree/master/packages/core-js/proposals): `import "core-js/proposals/string-replace-all"`.
```
zloirock: I am not sure that this is a clear enough wording. For polyfilling proposals, `shippedProposals` and `proposals` flags make sense only for `usage`. For `entry`, makes sense the third option (and not only from `proposals` path - for example, `import "core-js"` includes import of all polyfills) without any flags.

docs/preset-env.md
```diff
@@ -229,38 +229,69 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
 
 `"usage"` | `"entry"` | `false`, defaults to `false`.
 
-> This option adds direct references to the `core-js` module as bare imports. Thus `core-js` will be resolved relative to the file itself and needs to be accessible. You may need to specify `core-js@2` as a top level dependency in your application if there isn't a `core-js` dependency or there are multiple versions.
-
 This option configures how `@babel/preset-env` handles polyfills.
 
+When either the `usage` or `entry` options are used, `@babel-preset-env` will add direct references to `core-js` modules as bare imports (or requires). This means `core-js` will be resolved relative to the file itself and needs to be accessible.
+
```
zloirock: The same as in `runtime` docs - I don't see any links to `core-js` documentation in this file.

zloirock: `Built-ins` link in `include` should be updated.

docs/preset-env.md
```diff
@@ -372,7 +414,8 @@ The following are currently supported:
 
 **Builtins**
 
-- [es7.array.flat-map](https://github.com/tc39/proposal-flatMap)
+- [esnext.global-this](https://github.com/tc39/proposal-global) (only supported by `core-js@3`)
+- [esnext.string.match-all](https://github.com/tc39/proposal-string-matchall) (only supported by `core-js@3`)
```
zloirock: Makes sense that will be injected only with `useBuiltIns: usage`.

docs/preset-env.md
```diff
@@ -302,7 +335,19 @@ var b = new Map();
 
 #### `useBuiltIns: false`
 
-Don't add polyfills automatically per file, or transform `import "@babel/polyfill"` to individual polyfills.
+Don't add polyfills automatically per file, and don't transform `import "core-js"` or `import "@babel/polufill"` to individual polyfills.
```
danez:
```suggestion
Don't add polyfills automatically per file, and don't transform `import "core-js"` or `import "@babel/polyfill"` to individual polyfills.
```

zloirock:
Seems also need to update `preset-env` options in the repl.

docs/polyfill.md
```diff
@@ -4,6 +4,12 @@ title: @babel/polyfill
 sidebar_label: polyfill
 ---
 
+> 🚨 As of Babel 7.4.0, this package has been deprecated in favor of directly including `core-js/stable` (to polyfill ECMAScript features) and `regenerator-runtime/runtime` (needed to use transpiled generator functions):
+> ```js
+> import "core-js/stable";
+> import "renegerator-runtime/runtime";
```
MayhemYDG: `regenerator-runtime` instead of `renegerator-runtime` 🙂

MayhemYDG:
```suggestion
> import "regenerator-runtime/runtime";
```

docs/plugin-transform-runtime.md
radum: As part of this doc part I would also add a bit more clarity around how `corejs` works in conjunction with the `corejs` option from `@babel/preset-env`.

If both options are set to something other than false then the corejs imports will be duplicated. One fore preset-env and one for transform-runtime.

A bit more clarity that if you use preset-env with corejs option you should not use the same for transform runtime will help some developer understand what is going on.

This relates to some confusion from this issue [#9728](https://github.com/babel/babel/issues/9728).

The same info could be added to the preset-env doc also for reference.

PR #1988, opened 2019-03-14
Add docs for template placeholders

nicolo-ribaudo:
Babel PR: https://github.com/babel/babel/pull/9648

### DIFF
--- a/docs/template.md
+++ b/docs/template.md
@@ -20,6 +20,27 @@ import generate from "@babel/generator";
 import * as t from "@babel/types";
 
 const buildRequire = template(`
+  var %%importName%% = require(%%source%%);
+`);
+
+const ast = buildRequire({
+  importName: t.identifier("myModule"),
+  source: t.stringLiteral("my-module"),
+});
+
+console.log(generate(ast).code);
+```
+
+Output:
+
+```js
+const myModule = require("my-module");
+```
+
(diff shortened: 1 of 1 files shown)


docs/template.md
```diff
@@ -20,22 +20,37 @@ import generate from "@babel/generator";
 import * as t from "@babel/types";
 
 const buildRequire = template(`
-  var IMPORT_NAME = require(SOURCE);
+  var %%importName%% = require(%%source%%);
```
existentialism: We should probably note that using this syntax requires `syntacticPlaceholders`?

I think we should keep the original docs and add a "check out this new hotness via this option" section right after it?

nicolo-ribaudo: No, it doesn't. If you are using `%%foo%%`, that option is automatically enabled.

existentialism: Sounds good.

> You can use either new or old placeholders by default, you just can't mix them.

Seems necessary to mention though?

docs/template.md
```diff
@@ -27,15 +47,15 @@ const ast = buildRequire({
...
 ```
 
+Output:
+
 ```js
 const myModule = require("my-module");
 ```
 
-### `.ast`
+### `.ast`!
```
nicolo-ribaudo:
```suggestion
### `.ast`
```

PR #2001, opened 2019-03-26
fix wrong output in babel-plugin-transform-member-expression-literals page

ssshooter:


ssshooter:
image

but...

nicolo-ribaudo:
In that example, it is only transforming `.const`.

nicolo-ribaudo:
Oh ok I see it. Could you update the _input_, so that it starts from obj.foo?

PR #2004, opened 2019-03-26
Document alternatives to tsconfig.json

nicolo-ribaudo:
Closes https://github.com/babel/babel/issues/9681

### DIFF
--- a/docs/plugin-transform-typescript.md
+++ b/docs/plugin-transform-typescript.md
@@ -20,6 +20,10 @@ Does not type-check its input. For that, you will need to install and set up Typ
 
   **Workaround**: Convert to using `export default` and `export const`, and `import x, {y} from "z"`.
 
+* Behaves as if the `--isolatedModules` option was passed to the TypeScript Compiler. This can't be worked around because Babel doesn't support cross-file analysis.
+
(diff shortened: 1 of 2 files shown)


docs/plugin-transform-typescript.md
```diff
@@ -79,8 +83,75 @@ Replace the function used when compiling JSX expressions.
...
+  ```js
+  module.exports = {
+    parserOpts: { strictMode: true },
+  };
+  ```
+- `--downlevelIteration`
+  You can use the `@babel/plugin-transform-for-of` plugin. If you are using `@babel/preset-env`, `for...of` is already transpiled using iterators when it isn't supported by your compilation target.
+- `--emitDecoratorMetadata`
+  This option isn't supported by an official Babel package since it is not part of the decorators proposal, but it is a TypeScript-specific addition.
+  If you rely on this feature, you can use the community plugin [babel-plugin-transform-typescript-metadata](https://github.com/leonardfactory/babel-plugin-transform-typescript-metadata#readme).
```
nicolo-ribaudo: @leonardfactory Can we mention your plugin?

leonardfactory: @nicolo-ribaudo Absolutely! Thank you for asking 👍

ChristopherHButler: Does this plugin work the same way with babel standalone?

PR #2016, opened 2019-05-04
Fix references to babel-node

soulofmischief:
Got an error running `npx babel-node` and realized the docs were pointing to the old, non-scoped package.

I wasn't sure if it was semantically appropriate to change the references to `babel-node` made in the description as well. Is this project referred to as `babel-node` or `@babel/node`?

### DIFF
--- a/docs/node.md
+++ b/docs/node.md
@@ -28,31 +28,31 @@ it will compile ES6 code before running it.
 Launch a REPL (Read-Eval-Print-Loop).
 
 ```sh
-npx babel-node
+npx @babel/node
 ```
 
 Evaluate code.
 
 ```sh
-npx babel-node -e "class Test { }"
+npx @babel/node -e "class Test { }"
 ```
 
 Compile and run `test.js`.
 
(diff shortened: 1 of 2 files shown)


microbouji:
I believe the doc is referring to the use case when you've installed it locally first. In that case, the binary is still named `babel-node` so `npx babel-node` should be ok.

`npx @babel/node` will do that "one-off" execution from the registry without a local installation.

soulofmischief:
So `npx @bable/node` is correct then. The point of `npx` is to run a package which may not be installed locally. `npx babel-node` depends on the package to be installed locally. `npx @babel/node` still picks up the local package. So we should be using `npx @babel/node`.

danez:
Can you please also do the changes in this file? https://github.com/babel/website/blob/master/website/versioned_docs/version-7.0.0/node.md

soulofmischief:
@danez Done.

soulofmischief:
Anybody wanna merge and close this?

PR #2019, opened 2019-05-06
Add filename option in REPL config

### LINKED ISSUE #1666: Include file extension in REPL config options
With https://github.com/babel/babel/pull/7955 Babel now allows both the flow and typescript presets to be used in parallel, and allows JSX only in `.tsx`, not `.ts` files. Given that, we'll want to stop hard-coding the filename "repl" with no extension in https://github.com/babel/website/blob/71a346f6ab95047b9c6a1df373aca5b9dc609501/js/repl/compile.js#L84

Including the extension is important now, so the UI should expose that as well.

hackape:
fix #1666, fix https://github.com/babel/babel/issues/8117
the long pending issue that breaks typescript compilation

### DIFF
--- a/js/repl/PluginConfig.js
+++ b/js/repl/PluginConfig.js
@@ -101,6 +101,7 @@ const replDefaults: ReplState = {
   fileSize: false,
   timeTravel: false,
   sourceType: "module",
+  filename: "repl.js",
   forceAllTransforms: false,
   isEnvPresetTabExpanded: false,
   isPluginsExpanded: false,
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -67,6 +67,7 @@ type State = {
   fileSize: boolean,
   timeTravel: boolean,
   sourceType: SourceType,
+  filename: string,
   isSidebarExpanded: boolean,
   lineWrap: boolean,
   meta: Object,
@@ -151,6 +152,7 @@ class Repl extends React.Component<Props, State> {
       fileSize: persistedState.fileSize,
       timeTravel: persistedState.timeTravel,
       sourceType: persistedState.sourceType,
+      filename: persistedState.filename,
       isSidebarExpanded: persistedState.showSidebar,
       lineWrap: persistedState.lineWrap,
       meta: {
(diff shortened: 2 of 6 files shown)


hackape:
@loganfsmyth can you help review this PR?

danez:
besides the extension I guess there is not much use for the filename, right?

Why don't we just change the filename internally when the  typescript preset is on?

hackape:
@danez According to @nicolo-ribaudo changing the filename is the preferred solution, though I cannot tell the what's the preferred way to change it. My thought is the filename might play a role in scenarios other than typescript, maybe in the future. So an option to allow user to manually change filename might be better than hard-coded to fit typescript in a sense it's future proof.

hackape:
With all due respect, this issue has been bikeshedded for a whole year. People just want typescript compilation to work in repl, that's all. Whatever conclusion you guys come to, I'll respect that and change my implementation accordingly. But please, don't leave this pending like before.

existentialism:
It depends how much we want the REPL to illustrate/educate the differences between the functionality of presets and `.js` vs `.ts` vs. `.tsx`. 

If we want TS to _just work_, then my initial reaction is to just change the filename internally.

(FWIW, just because it's been an open issue hardly means it's been bikeshedded, and as far as I know we actually _haven't_ discussed this much.)

hackape:
https://github.com/babel/website/blob/4459be55adc517a5d5d1ae846f2b0ed3b6e1e542/js/repl/compile.js#L90-L93

If I understand it right, for the current state of repl, changing ln:92 to `filename: "repl.tsx"` suffices to make everything work. Is that acceptable?

nicolo-ribaudo:
No, that would prevent `<Type> foo` type cast from working.
The extension (at least) needs to be configurable; but I don't see anything bad in allowing configuring the whole filename.

hzoo:
This is cool, but my hesitation is that it is extremely unlikely a user is going to know to change the filename (especially if we don't have an error message for it).

Screen Shot 2019-05-20 at 10 53 09 AM

It just looks like it doesn't work. So I would suggest the "Why don't we just change the filename internally when the typescript preset is on?" suggestion (exposing the filename is fine but seems unnecessary in a repl, especially if it doesn't do anything else). There's limited space for options and it's just for testing, in this environment/scenario everything should probably just work with the least changing/config possible.

PR #2021, opened 2019-05-08
feat: add link to preset/plugins options docs

tanhauhau:
Docs improvements based on https://github.com/babel/babel/issues/9428


TheJaredWilcurt:
This is good. But it isn't really a solution to the problem I had. My problem was that I could see there were options that could be set, but there was no easy example of understanding **where** they go.

Here is an example of what would have made it super easy:

* https://github.com/babel/website/pull/2029

It just shows exactly where this options object goes, with complete context.

PR #2034, opened 2019-05-22
Expression was false

Pierre-Thibault:
Breaking it in two lines makes it true.


xtuc:
I don't think we are interested in correctness here but just showing an example.

Pierre-Thibault:
> I don't think we are interested in correctness here but just showing an example.

I think it is better to write expressions that are true for better comprehension.

PR #2037, opened 2019-05-24
Updates @babel/plugin-syntax-dynamic-import core-js module paths (core-js@3)

jbottigliero:
- Adds core-js@3 plugin path varients to documentation.

### DIFF
--- a/docs/plugin-syntax-dynamic-import.md
+++ b/docs/plugin-syntax-dynamic-import.md
@@ -44,8 +44,12 @@ Currently, `@babel/preset-env` is unaware that using `import()` with [Webpack re
 // webpack config
 const config = {
   entry: [
+    // using core-js@2
     "core-js/modules/es.promise",
     "core-js/modules/es.array.iterator",
+    // or using core-js@3
+    "core-js/stable/promise",
+    "core-js/stable/array/iterator",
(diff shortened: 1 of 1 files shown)


docs/plugin-syntax-dynamic-import.md
```diff
@@ -44,8 +44,12 @@ Currently, `@babel/preset-env` is unaware that using `import()` with [Webpack re
 // webpack config
 const config = {
   entry: [
+    // using core-js@2
     "core-js/modules/es.promise",
```
nicolo-ribaudo: This is already for `core-js@3`: `core-js@2` uses `es6.` instead of `es.`.

Also, those files exist in `core-js@3`: https://unpkg.com/core-js@3.1.2/modules/

zloirock:
`preset-env` with `core-js@3` should automatically inject `Promise` with dependencies on dynamic import.

docs/plugin-syntax-dynamic-import.md
jbottigliero: I was seeing module resolution errors with the (internal) `modules/` based paths after swapping the configuration over to `core-js@3` and found swapping to the `/stable/` variants to be the solution. 

Something else could certainly be the cause of the resolution error if the `es.` version is supported.

PR #1724, opened 2018-07-13
Fix v6 links for renamed plugins

microbouji:
.md links for renamed plugins in https://babeljs.io/docs/en/plugins#transform-plugins are not working because of https://github.com/facebook/Docusaurus/issues/845. Until that is fixed we can switch to relative links.

PR #2044, opened 2019-06-13
babylon docs link broken.

konojunya:
Hi, babel team!
babel is a great tool, thank you :)

## What does this change?

Since link is broken, I will fix it so that I can redirect it.
The cause of the problem is a missing document accompanying babylon rename to parser.
I think that placing `docs/babylon.md` and redirecting is a good way to reduce the scope of the impact, rather than touching the tool that produces the documentation.

## Problematic URL

https://babeljs.io/docs/en/6.26.3/babylon

The link to `docs/babylon.md` on the edit button on this page does not work.

## Points to be careful

Is the content of the document the quality of the document of babel?

If you need to change, please tell me where to change.

### DIFF
--- a/docs/babylon.md
+++ b/docs/babylon.md
@@ -1,0 +1,5 @@
+# Babylon is now [moved into the main Babel mono-repo](https://github.com/babel/website/tree/master/docs/parser.md) `babel-parser`.
+
+The move makes it much easier to release and develop in sync with the rest of Babel!
+
+## Check out the 6.x README [here](https://github.com/babel/babylon/blob/6.x/README.md)


konojunya:
This link if you want to see the look immediately:

https://github.com/konojunya/website/blob/fix/6.26.3-babylon-docs-link/docs/babylon.md

microbouji:
Thank you for looking into this. Unfortunately we have quite a few more of these renamed pages ([example](https://github.com/babel/website/pull/1724/files)) that would have to be fixed.

For now we're kinda waiting for docusaurus v2 which seems will fix this.

harish-sethuraman:
@JLHwung I think this PR can be closed? should have probably been closed with docusaurus migration?

JLHwung:
Closing this PR as the V6 docs has long been archived, the Babel v6 babylon docs already featured a redirection link as proposed in this PR: https://v6.babeljs.io/docs/core-packages/babylon/

PR #2047, opened 2019-06-18
docs(@babel/parser): Document allowUndeclaredExports option.

benjamn:
Documents the changes from https://github.com/babel/babel/pull/9864 implemented by @nicolo-ribaudo.

### DIFF
--- a/docs/parser.md
+++ b/docs/parser.md
@@ -46,6 +46,12 @@ mind. When in doubt, use `.parse()`.
   outside of class and object methods. Set this to `true` to accept such
   code.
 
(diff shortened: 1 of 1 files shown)


docs/parser.md
```diff
@@ -46,6 +46,12 @@ mind. When in doubt, use `.parse()`.
   outside of class and object methods. Set this to `true` to accept such
   code.
 
+- **allowUndeclaredExports**: By default, exporting an identifier that was
+  not declared in the current module scope will raise an error. Set this
```
nicolo-ribaudo: Maybe add a note that this is the spec behavior?

PR #2063, opened 2019-07-13
Remove Coming Soon! from fsharp pipelines

mAAdhaTTah:
Since they've now landed, we can remove this from the website.


danez:
Can you copy this file to `/website/versioned_docs/version-7.5.0/` and commit it? then it will also show up in the current version 7.5.0 on the website. The `/docs/` folder is only the next/master version of babel, https://babeljs.io/versions.

mAAdhaTTah:
Ah ok. Done!

PR #2065, opened 2019-07-16
Modernize examples for plugin-transform-react-jsx

loklaan:
👋

I think the code examples for `@babel/plugin-transform-react-jsx` could benefit from a small update, to keep them relevant!

* `deku` → `@emotion/core`
* es5 + cjs → esnext + esm

Let me know if this change would be undesirable, cheers!

### DIFF
--- a/docs/plugin-transform-react-jsx.md
+++ b/docs/plugin-transform-react-jsx.md
@@ -33,9 +33,9 @@ const profile = React.createElement("div", null,
 **In**
 
 ```javascript
-/** @jsx jsx */
+/** @jsx Preact.h */
 
-import { jsx } from '@emotion/core';
+import Preact from 'preact';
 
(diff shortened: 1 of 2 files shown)


suchipi:
I actually like that the `/* @jsx dom */` makes it clear that the thing after the pragma is a variable name. I don't think that is as clear with `/* @jsx jsx */`. Could we keep it using deku? Or maybe hyperscript or something else whose element constructor is not `jsx`?

loklaan:
Oh yeah, good point!

I'll swap out the `@emotion/core` → `preact` (which also has a custom `Fragment` jsxFrag to use 👌)

suchipi:
Perfect, thanks!

suchipi:
Nice to get a real fragment thing in there, too :)

PR #2067, opened 2019-07-19
@babel/preset-env: Clarify `useBuiltIns` usage and purpose

babramczyk:
This PR clarifies what `useBuiltIns` is for, and how to use it, better (IMO 🙂).

Specifically, I really struggled using this option in one of my projects. It took me a long time to understand:

- That this option only told Babel _how_ to import polyfills, and that it didn't come with them
  - Specifically, that you had to `npm install` them yourself
- `core-js` and `regenerator`/`regenerator-runtime` were exactly the two polyfills it supported
  - The old documentation focuses on `core-js`, and brushes over `regenerator`
- `regenerator-runtime/runtime` was installed via the `regenerator` npm package
  - Just took me a while to figure which exact package to install (`regenerator-runtime`? Just `regenerator`?)

Feedback welcome 😃

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -266,7 +266,7 @@ import "regenerator-runtime/runtime";
 // If the target envs don't support these methods
 import "core-js/modules/es.string.pad-start";
 import "core-js/modules/es.string.pad-end";
-// If the target envs don't support generators or `async`/`await`
+// If the target envs don't support generators
 import "regenerator-runtime/runtime`;
 ```
 
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -229,14 +229,14 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
 
 `"usage"` | `"entry"` | `false`, defaults to `false`.
 
-This option configures how `@babel/preset-env` imports various polyfills in your generated code.
+This option configures how `@babel/preset-env` imports polyfills and runtime support into your generated code.
 
-Specifically, this option supports the following external polyfill packages:
+Specifically, this option supports the following external packages:
 
 - `core-js`: For standard JS built-ins
-- `regenerator`: For generator and `async`/`await` support
+- `regenerator-runtime`: For generator and `async`/`await` support
 
-Since these polyfills do not come with Babel or `@babel/preset-env` itself, they'll need to be installed as production-level dependencies. This way, when the Babel-generated code imports them, they'll be accessible (in `node_modules`).
+Since these packages do not come with Babel or `@babel/preset-env` itself, they'll need to be installed as production-level dependencies. This way, when the Babel-generated code imports them, they'll be accessible (in `node_modules`).
 
 ```sh
 npm install core-js regenerator --save
@@ -251,7 +251,7 @@ npm install core-js regenerator --save
 > Multiple imports or requires of those packages might cause global collisions and other issues that are hard to trace.
 > We recommend creating a single entry file that only contains the `import` statements.
 
-This option enables a new plugin that replaces the `import "core-js";` and `import "regenerator-runtime/runtime"` statements with only what's needed from each of them, depending on the target environments.
+This option enables a new plugin that replaces the `import "core-js"` and `import "regenerator-runtime/runtime"` statements with only what's needed from each of them, depending on the target environments.
 
 **In**
 
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -234,7 +234,7 @@ This option configures how `@babel/preset-env` imports polyfills and runtime sup
 Specifically, this option supports the following external packages:
 
 - `core-js`: For standard JS built-ins
-- `regenerator-runtime`: For generator and `async`/`await` support (for browsers that lack support)
+- `regenerator-runtime`: For generator and `async`/`await` support (for browsers that lack support for generators)
 
 Since these packages do not come with Babel or `@babel/preset-env` itself, they'll need to be installed as production-level dependencies. This way, when the Babel-generated code imports them, they'll be accessible (in `node_modules`).
 
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -239,7 +239,7 @@ Specifically, this option supports the following external packages:
 Since these packages do not come with Babel or `@babel/preset-env` itself, they'll need to be installed as production-level dependencies. This way, when the Babel-generated code imports them, they'll be accessible (in `node_modules`).
 
 ```sh
-npm install core-js regenerator --save
+npm install core-js regenerator-runtime --save
 ```
 
 > NOTE: Previously, these would be installed via `@babel/polyfill`, which is now deprecated. The above method of installing these polyfill packages is the recommended way.
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -234,7 +234,7 @@ This option configures how `@babel/preset-env` imports polyfills and runtime sup
 Specifically, this option supports the following external packages:
 
 - `core-js`: For standard JS built-ins
-- `regenerator-runtime`: For generator and `async`/`await` support
+- `regenerator-runtime`: For generator and `async`/`await` support (for browsers that lack support)
 
 Since these packages do not come with Babel or `@babel/preset-env` itself, they'll need to be installed as production-level dependencies. This way, when the Babel-generated code imports them, they'll be accessible (in `node_modules`).
 
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -229,40 +229,45 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
 
 `"usage"` | `"entry"` | `false`, defaults to `false`.
 
-This option configures how `@babel/preset-env` handles polyfills.
+This option configures how `@babel/preset-env` imports various polyfills in your generated code.
 
-When either the `usage` or `entry` options are used, `@babel-preset-env` will add direct references to `core-js` modules as bare imports (or requires). This means `core-js` will be resolved relative to the file itself and needs to be accessible.
+Specifically, this option supports the following external polyfill packages:
 
-Since `@babel/polyfill` was deprecated in 7.4.0, we recommend directly adding `core-js` and setting the version via the [`corejs`](#corejs) option.
+- `core-js`: For standard JS built-ins
+- `regenerator`: For generator and `async`/`await` support
+
+Since these polyfills do not come with Babel or `@babel/preset-env` itself, they'll need to be installed as production-level dependencies. This way, when the Babel-generated code imports them, they'll be accessible (in `node_modules`).
 
 ```sh
-npm install core-js@3 --save
-
-# or
-
-npm install core-js@2 --save
+npm install core-js regenerator --save
 ```
+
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Could we also add a note that it is incompatible with `transform-runtime`? (ref: https://github.com/babel/babel/issues/10271)

docs/preset-env.md
```diff
@@ -229,40 +229,45 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
 
-This option configures how `@babel/preset-env` handles polyfills.
+This option configures how `@babel/preset-env` imports various polyfills in your generated code.
 
-When either the `usage` or `entry` options are used, `@babel-preset-env` will add direct references to `core-js` modules as bare imports (or requires). This means `core-js` will be resolved relative to the file itself and needs to be accessible.
+Specifically, this option supports the following external polyfill packages:
 
-Since `@babel/polyfill` was deprecated in 7.4.0, we recommend directly adding `core-js` and setting the version via the [`corejs`](#corejs) option.
+- `core-js`: For standard JS built-ins
+- `regenerator`: For generator and `async`/`await` support
```
zloirock: `regenerator` is not a polyfill.

docs/preset-env.md
```diff
@@ -229,40 +229,45 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
 
 `"usage"` | `"entry"` | `false`, defaults to `false`.
 
-This option configures how `@babel/preset-env` handles polyfills.
+This option configures how `@babel/preset-env` imports various polyfills in your generated code.
 
-When either the `usage` or `entry` options are used, `@babel-preset-env` will add direct references to `core-js` modules as bare imports (or requires). This means `core-js` will be resolved relative to the file itself and needs to be accessible.
+Specifically, this option supports the following external polyfill packages:
 
-Since `@babel/polyfill` was deprecated in 7.4.0, we recommend directly adding `core-js` and setting the version via the [`corejs`](#corejs) option.
```
zloirock: This note is useful for many users, so I don't think that remove it and add a stripped version is a good idea.

docs/preset-env.md
```diff
@@ -229,40 +229,45 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
+- `core-js`: For standard JS built-ins
+- `regenerator`: For generator and `async`/`await` support
 
-```sh
-npm install core-js@3 --save
-
-# or
+Since these polyfills do not come with Babel or `@babel/preset-env` itself, they'll need to be installed as production-level dependencies. This way, when the Babel-generated code imports them, they'll be accessible (in `node_modules`).
 
-npm install core-js@2 --save
```
zloirock: Usage of different `core-js` versions causes many questions and problems for users, so remove the note of how to install different `core-js` versions is not a good idea.

patrick-mcdougle:
> Could we also add a note that it is incompatible with `transform-runtime`? (ref: [babel/babel#10271](https://github.com/babel/babel/issues/10271))

@nicolo-ribaudo But if preset-env and transform-runtime don't work together, won't people lose the benefit of sharing the runtime/helper code across the entire project? Not sure people are going to like that.

nicolo-ribaudo:
Sorry, I meant with `transform-runtime`'s `core-js` option

docs/preset-env.md
babramczyk: I don't think I say it is..? Unless you're saying I'm implying that it is -- in which case, could you suggest a better way of describing it?

zloirock: > I don't think I say it is..?

> ... the following external polyfill packages:
> - `core-js`: ...
> - `regenerator`: ...
---
It's runtime, not a polyfill.

docs/preset-env.md
babramczyk: I have a line below that reads:

```md
> NOTE: Previously, these would be installed via `@babel/polyfill`, which is now deprecated. The above method of installing these polyfill packages is the recommended way.
```

I don't see this as any more stripped down than the line you commented on, just moved down to be more contextual. Do you want me to move this line up?

docs/preset-env.md
babramczyk: That's what I was trying to do by removing the different versions. I found the fact that the documentation had two different versions of `core-js` confusing -- why are there two? Which one should I be using? It made me do unnecessary research into `core-js`, which I don't think a user of `@babel/preset-env` should have to do.

Are there specific use cases for 2 vs 3? If so, I'd be happy to add notes about them 🙂

babramczyk:
> Could we also add a note that it is incompatible with `transform-runtime`? (ref: [babel/babel#10271](https://github.com/babel/babel/issues/10271))

Do you have wording in mind on this? Might be easier than me trying to come up with something and having it be inaccurate 😅

docs/preset-env.md
vdh: Shouldn't this be `regenerator-runtime` instead of just `regenerator`?

babramczyk: Gotcha, thank you.

babramczyk: Thank you -- fixed.

docs/preset-env.md
```diff
@@ -229,40 +229,45 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
 
-This option configures how `@babel/preset-env` handles polyfills.
+This option configures how `@babel/preset-env` imports polyfills and runtime support into your generated code.
 
-When either the `usage` or `entry` options are used, `@babel-preset-env` will add direct references to `core-js` modules as bare imports (or requires). This means `core-js` will be resolved relative to the file itself and needs to be accessible.
+Specifically, this option supports the following external packages:
 
-Since `@babel/polyfill` was deprecated in 7.4.0, we recommend directly adding `core-js` and setting the version via the [`corejs`](#corejs) option.
+- `core-js`: For standard JS built-ins
+- `regenerator-runtime`: For generator and `async`/`await` support
```
nicolo-ribaudo: Note that this isn't needed for `async`/`await` if your browser supports generators.

docs/preset-env.md
```diff
@@ -229,40 +229,45 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
+import "regenerator-runtime/runtime";
 ```
 
 **Out (different based on environment)**
 
 ```js
+// If the target envs don't support these methods
 import "core-js/modules/es.string.pad-start";
 import "core-js/modules/es.string.pad-end";
+// If the target envs don't support generators or `async`/`await`
```
nicolo-ribaudo:
```suggestion
// If the target envs don't support generators
```

jonkoops:
One thing that I find unclear about the documentation of `useBuiltIns: 'usage'` is whether or not I need to manually `import 'regenerator-runtime/runtime'` or if this mode will automatically add the import for me if needed.

Could this be clarified in this PR as well, to avoid future confusion?

docs/preset-env.md
nicolo-ribaudo: Can you also mention this here?

babramczyk: Isn't that assumed for this whole section, since `preset-env`'s entire purpose is to fill in for browsers' shortcomings.

i.e. `core-js` isn't needed if your browser supports all JS built-ins that it brings

nicolo-ribaudo: My point was that `regenerator-runtime` isn't always needed by browsers that don't support `async`/`await`, while this sentence implies the opposite.

`regenerator-runtime` is only needed for browsers that don't support generators.

babramczyk: I see what you mean, but don't know exactly what text would make sense.

Do you have a suggestion?

nicolo-ribaudo: Maybe this?

```suggestion
- `regenerator-runtime`: For generators and `async`/`await` in browsers without generators support
```

babramczyk: Done. Changed up your suggestion a little. I don't think everyone knows that `async`/`await` is built on generators, so I just grouped them together so it would be clear for everyone

docs/preset-env.md
```diff
@@ -229,40 +229,45 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
 #### `useBuiltIns: 'entry'`
 
 > NOTE: Only use `import "core-js";` and `import "regenerator-runtime/runtime";` once in your whole app.
-> If you are using `@babel/polyfill`, it already includes both `core-js` and `regenerator-runtime`: importing it twice will throw an error.
+> If you are using `@babel/polyfill` (which is deprecated), it already includes both `core-js` and `regenerator-runtime`: importing it twice will throw an error.
 > Multiple imports or requires of those packages might cause global collisions and other issues that are hard to trace.
 > We recommend creating a single entry file that only contains the `import` statements.
 
-This option enables a new plugin that replaces the `import "core-js/stable";` and `import "regenerator-runtime/runtime"` statements (or `require("corejs")` and `require("regenerator-runtime/runtime")`) with individual requires to different `core-js` entry points based on environment.
+This option enables a new plugin that replaces the `import "core-js"` and `import "regenerator-runtime/runtime"` statements with only what's needed from each of them, depending on the target environments.
```
existentialism:
```suggestion
This option enables a plugin that replaces the `import "core-js"` and `import "regenerator-runtime/runtime"` statements with only what your target environments need from each of them.
```

docs/preset-env.md
```diff
@@ -229,40 +229,45 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
 
-Since `@babel/polyfill` was deprecated in 7.4.0, we recommend directly adding `core-js` and setting the version via the [`corejs`](#corejs) option.
+- `core-js`: For standard JS built-ins
+- `regenerator-runtime`: For generator and `async`/`await` support (for browsers that lack support)
 
-```sh
-npm install core-js@3 --save
-
-# or
+Since these packages do not come with Babel or `@babel/preset-env` itself, they'll need to be installed as production-level dependencies. This way, when the Babel-generated code imports them, they'll be accessible (in `node_modules`).
```
existentialism: Wonder if we should be more explicit than "production-level", something like:

> installed a `dependencies` (not `devDependencies`)

docs/preset-env.md
```diff
@@ -229,40 +229,45 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
 
-This option configures how `@babel/preset-env` handles polyfills.
+This option configures how `@babel/preset-env` imports polyfills and runtime support into your generated code.
 
-When either the `usage` or `entry` options are used, `@babel-preset-env` will add direct references to `core-js` modules as bare imports (or requires). This means `core-js` will be resolved relative to the file itself and needs to be accessible.
+Specifically, this option supports the following external packages:
 
-Since `@babel/polyfill` was deprecated in 7.4.0, we recommend directly adding `core-js` and setting the version via the [`corejs`](#corejs) option.
+- `core-js`: For standard JS built-ins
+- `regenerator-runtime`: For generator and `async`/`await` support (for browsers that lack support)
```
nicolo-ribaudo:
```suggestion
- `regenerator-runtime`: For generator and `async`/`await` support (for browsers that lack support for generators)
```

Again, it's important here to note that if a browser supports generators you can use `async` _without_ needing `regenerator-runtime` :sweat_smile:

PR #2070, opened 2019-07-21
Fix `npx babel-node` usage

### LINKED ISSUE #2083: Documentation: wrong rlwrap command line
On https://babeljs.io/docs/en/next/babel-node.html the syntax to start with rlwrap is wrong.

It must be 

```
rlwrap npx @babel/node
```
and not the other way around. Same for the second example

JLHwung:
Fixes #2083 

This PR reverts #2016, and deals with an [issue](https://github.com/babel/babel/issues/10246) reported on babel.

Per [npx synopsis](https://github.com/npm/npx#description), `npx` will execute the `<command>` from `node_modules/.bin`, if the command is not found, it will query `<command>` package from npm registry and install it. When `<command>` contains `@scope`, it will [strip off](https://github.com/npm/npx/blob/latest/parse-args.js#L138) the `@scope`. Therefore `npx @babel/node` simply run `node` without any babel-register hooks.

As `@babel/node` provides binary command `babel-node`, we should run `npx babel-node` to invoke it.

For those prefer to install dependencies on the fly, another example is offered:
```
npx --package=@babel/{core,node,preset-env} babel-node
```

cc @Banou26 @soulofmischief

Unrelated: we also correct the `rlwrap` usage here: `rlwrap` is not a npm package so we cannot invoke `rlwrap` from `npx`.

### DIFF
--- a/docs/node.md
+++ b/docs/node.md
@@ -35,7 +35,7 @@ npx babel-node
 
 If you prefer not to install `@babel/node` and `@babel/core`, you can install them on-the-fly:
 ```sh
-npx --package=@babel/{core,node,preset-env} babel-node
+npx -p @babel/core -p @babel/node babel-node
 ```
 
 Evaluate code.
--- a/docs/node.md
+++ b/docs/node.md
@@ -27,8 +27,15 @@ it will compile ES6 code before running it.
 
 Launch a REPL (Read-Eval-Print-Loop).
 
+> You should install `@babel/node` and `@babel/core` first before `npx babel-node`, otherwise `npx` will install out-of-dated legacy `babel-node` 6.x.
+
 ```sh
 npx babel-node
+```
+
+Install basic dependencies on-the-fly and launch a REPL
+```sh
+npx --package=@babel/{core,node,preset-env} babel-node
 ```
 
 Evaluate code.
--- a/docs/node.md
+++ b/docs/node.md
@@ -33,7 +33,7 @@ Launch a REPL (Read-Eval-Print-Loop).
 npx babel-node
 ```
 
-Install basic dependencies on-the-fly and launch a REPL
+If you prefer not to install `@babel/node` and `@babel/core`, you can install them on-the-fly:
 ```sh
 npx --package=@babel/{core,node,preset-env} babel-node
 ```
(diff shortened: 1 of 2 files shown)


marcofugaro:
Please merge this PR! It has solved a lot of headaches

docs/node.md
```diff
@@ -27,32 +27,39 @@ it will compile ES6 code before running it.
...
+> You should install `@babel/node` and `@babel/core` first before `npx babel-node`, otherwise `npx` will install out-of-dated legacy `babel-node` 6.x.
+
+```sh
+npx babel-node
+```
+
+Install basic dependencies on-the-fly and launch a REPL
 ```sh
-npx @babel/node
+npx --package=@babel/{core,node,preset-env} babel-node
```
nicolo-ribaudo: This won't use `@babel/preset-env`, right? If it doesn't, it should be removed to avoid confusion

nicolo-ribaudo: Also, I find this more clear, since not everyone knows about `{ }` (and I don't know if they work in every shell)

```
npx -p @babel/core -p @babel/node babel-node
```

docs/node.md
```diff
@@ -27,32 +27,39 @@ it will compile ES6 code before running it.
 
 Launch a REPL (Read-Eval-Print-Loop).
 
+> You should install `@babel/node` and `@babel/core` first before `npx babel-node`, otherwise `npx` will install out-of-dated legacy `babel-node` 6.x.
+
+```sh
+npx babel-node
+```
+
+Install basic dependencies on-the-fly and launch a REPL
```
nicolo-ribaudo:
```suggestion
If you prefer not to install `@babel/node` and `@babel/core`, you can install them on-the-fly:
```

docs/node.md
JLHwung: > If it doesn't, it should be removed to avoid confusion

True, we still need a babel config to use `@babel/preset-env`.

> I find [expanding brace expressions] more clear

True, especially if we don't include `@babel/preset-env` in the example. Addressed in d1b18ff.

Sidenote: Brace expansion is first [introduced](https://en.wikipedia.org/wiki/C_shell) in `csh` and then copied to other Unix shells.

PR #2074, opened 2019-07-25
feat: add v8intrinsic language extensions

JLHwung:
Related: https://github.com/babel/babel/pull/10148

/cc @mathiasbynens I couldn't find any official docs on v8 Compiler Intrinsics other than the source code you provided before. Could you spare some time write about it? I am more than happy to reference one in our docs.


mathiasbynens:
The closest things to `%V8Intrinsic` docs is: https://v8.dev/blog/elements-kinds#debugging

PR #2077, opened 2019-07-31
Update plugin-transform-typescript.md

orta:
👋 - I wanted to send more folks in the direction of this doc, and I think it needs a fresh look from the perspective of someone who doesn't know TypeScript well.

The `namespace` caveats section is kinda overwhelming for someone who doesn't know what the feature is, so I've added a prefix summerizing what it is.

### DIFF
--- a/docs/plugin-transform-typescript.md
+++ b/docs/plugin-transform-typescript.md
@@ -55,9 +55,9 @@ require("@babel/core").transform("code", {
 
 ## Caveats
 
-Because there are features of the TypeScript language which rely on the full type-system to be available to make changes at runtime. This section of caveats is quite long, however, it's worth noting that a few of these features are only found in older TypeScript codebases and have modern JavaScript equivilents which you are probably already using.
+Because there are features of the TypeScript language which rely on the full type-system to be available to make changes at runtime. This section of caveats is quite long, however, it's worth noting that a few of these features are only found in older TypeScript codebases and have modern JavaScript equivalents which you are probably already using.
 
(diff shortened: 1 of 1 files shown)


orta:
Cool, I just gave it a more thorough run-through 👍

docs/plugin-transform-typescript.md
```diff
@@ -4,11 +4,79 @@ title: @babel/plugin-transform-typescript
...
+require("@babel/core").transform("code", {
+  plugins: ["@babel/plugin-transform-typescript"]
+});
+```
+
 
 ## Caveats
 
-* Since Babel does not type-check, invalid TypeScript may successfully get transformed, and often in unexpected or invalid ways. 
+Because there are features of the TypeScript language which rely on the full type-system to be available to make changes at runtime. This section of caveats is quite long, however, it's worth noting that a few of these features are only found in older TypeScript codebases and have modern JavaScript equivilents which you are probably already using.
```
Kriyszig: Possible typo:  equiv**a**lents instead of equiv**i**lents

docs/plugin-transform-typescript.md
```diff
@@ -4,11 +4,79 @@ title: @babel/plugin-transform-typescript
...
+
+1. Since Babel does not type-check, code which is syntacically correct, but would fail the TypeScript type-checking may successfully get transformed, and often in unexpected or invalid ways.
```
Kriyszig: Possible typo: syntac**t**ically instead of syntacically

orta:
Fixed typoes - thanks @Kriyszig 

This should be good to go now

orta:
Alright folks, we should get this in ;)

xtuc:
I will go ahead a merge this, thanks for your work.

PR #2080, opened 2019-08-09
Script to generate docs for types and updated docs/types.md

### LINKED ISSUE #2079: t.tsUnknownType is documented but doesn't exist
## Bug Report

**Current Behavior**
Looking at the documentation for type there is a [tsUnknownType](https://babeljs.io/docs/en/babel-types.html#tsunknowntype), but calling the function gives a `TypeError: t.tsUnknownType is not a function`

**Input Code**
```js
const generate = require("@babel/generator").default;
const t = require("@babel/types");

const ast = t.tsUnknownType([t.tsAnyKeyword()]);

const { code } = generate(ast);
console.log(code);
```

**Expected behavior/code**
I believe the output should be `unknown`.

**Babel Configuration (.babelrc, package.json, cli command)**
N/A, this is more a tooling bug than babel behavior bug

**Environment**
- Babel version(s): 7.5.5
- Node/npm version: Node v10.12.0 / yarn 1.17.3
- OS: Ubuntu 18.04
- How you are using Babel: `api`

**Possible Solution**
Searching through this repo yields nothing, it might not exist (yet)?

deificx:
I ended up writing a script to generate this documentation based of type definitions found in `node_modules/@babel/types/lib/index.d.ts`. It does so by looking through the properties of the export from `@babel/types`, seeing which functions (when turned to string, which is not a great idea) includes a call to `_builder.default`, then it looks into the `index.d.ts` file above, yet again using Regex and string matching, to extract the types.

This script is run by another script `scripts/generate-docs-types-versions.js` which creates a directory for each version, installs that version and generates the `types.md` using `scripts/generate-docs-types.js`. It's not exactly pretty, but it works.

I tried to keep line spacing and such similar to keep the diff as small as possible.

I'm not sure what to do with the version doc for version-6.26.3, as I'm not able to locate the babel package for that on npm.

## Notable changes
- Call signature example only includes required arguments
- Alphabetical ordering for the functions and aliases
- Always includes `Aliases`, even if there are none.
- Assumed default values based on type
  - Array<any> (default: `[]`)
  - boolean (default: `false`)
  - string (default: `""`)
  - default (default: `null`)
- `t.tsUnknownType(types)` "renamed" `t.tsUnknownKeyword()`, fixes #2079 
- `t.import()` has disappeared.

### DIFF
--- a/scripts/generate-docs-types-versions.js
+++ b/scripts/generate-docs-types-versions.js
@@ -1,0 +1,72 @@
+const fs = require("fs");
+const path = require("path");
+const { spawn } = require("child_process");
+const { promisify } = require("util");
+
+const mkdir = promisify(fs.mkdir);
+const cp = promisify(fs.copyFile);
+
+const work = (cwd, cmd, argv) =>
+  new Promise((resolve, reject) => {
+    const worker = spawn(cmd, argv, { cwd, shell: true });
+
+    worker.stdout.on("data", data => {
+      console.log(`stdout: ${data.toString()}`);
+    });
+
+    worker.stderr.on("data", data => {
+      console.log(`stderr: ${data.toString()}`);
+    });
+
+    worker.on("close", code => {
+      if (code) {
+        reject();
+      } else {
+        resolve();
+      }
+    });
+  });
+
+(async () => {
+  try {
+    await mkdir(path.resolve(__dirname, "tmp"));
+  } catch (error) {
+    console.log("failed creating directory tmp");
+    console.log(error);
+  }
+  const latest = "7.5.5";
+  const typesVersions = ["7.0.0", "7.2.0", "7.4.0", latest];
+  for (const babelVersion of typesVersions) {
+    const location = path.resolve(__dirname, "tmp", babelVersion);
+    try {
+      await mkdir(location);
+    } catch (error) {
+      console.log("failed creating directory:", location);
+      console.log(error);
+    }
+    await work(location, "yarn", ["init", "-y"]);
+    await work(location, "yarn", ["add", `@babel/core@${babelVersion}`]);
+    await cp(
+      path.resolve(__dirname, "generate-docs-types.js"),
+      path.join(location, "generate-docs-types.js")
+    );
+    await work(location, "node", [
+      "generate-docs-types.js",
+      babelVersion === latest
+        ? ""
+        : `--babel-types-id=version-${babelVersion}-babel-types`,
+    ]);
+    const dest =
+      babelVersion === latest
+        ? path.resolve(__dirname, "..", "docs", "types.md")
+        : path.resolve(
+            __dirname,
+            "..",
+            "website",
+            "versioned_docs",
+            `version-${babelVersion}`,
+            "types.md"
+          );
+    await cp(path.join(location, "types.md"), dest);
+  }
+})();
--- a/scripts/generate-docs-types.js
+++ b/scripts/generate-docs-types.js
@@ -1,0 +1,279 @@
+const fs = require("fs");
+const path = require("path");
+const t = require("@babel/types");
+const { promisify } = require("util");
+
+const readfile = promisify(fs.readFile);
+const writefile = promisify(fs.writeFile);
+
+const callRegex = /(\w+)(\??:)\s(.+)/;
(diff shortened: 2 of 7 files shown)


deificx:
Not sure why there was merge conflicts here, but it looks like masters diff didn't like the re-ordering of the methods. Did a search to see that I didn't remove anything, but haven't checked for duplicates.

scripts/generate-docs-types.js
```diff
@@ -0,0 +1,287 @@
...
+      "index.d.ts"
+    ),
+    "utf-8"
+  );
+  const typeDefLines = typeDefs.split("\n");
+  const aliases = findAliases(typeDefLines);
+  const sections = [];
+
+  for (const type in t) {
+    const isBuilder = /_builder\.default/.test(t[type].toString());
```
tanhauhau: I think we should include not just the builder, but as well as `assert*` and `is*` utility functions

scripts/generate-docs-types.js
```diff
@@ -0,0 +1,287 @@
...
+  const assert = String(`assert${type}`).toLocaleLowerCase();
+  for (const type in t) {
+    if (type.toLowerCase() === assert) {
+      see.assert = type;
+    }
+    if (type.toLowerCase() === is) {
+      see.is = type;
+    }
+  }
+  return `See also \`t.${see.is}(node, opts)\` and \`t.${
```
tanhauhau: what if `see.is` or `see.assert` is not available, wouldn't it be `See also and ` ?
maybe a join with `" and "` would be better?

tanhauhau: and if we adding `is` and `assert` section, maybe this can be a link to the section

scripts/generate-docs-types-versions.js
```diff
@@ -0,0 +1,72 @@
...
+
+(async () => {
+  try {
+    await mkdir(path.resolve(__dirname, "tmp"));
+  } catch (error) {
+    console.log("failed creating directory tmp");
+    console.log(error);
+  }
+  const latest = "7.5.5";
+  const typesVersions = ["7.0.0", "7.2.0", "7.4.0", latest];
```
nicolo-ribaudo: We shouldn't hard-code this, otherwise we need to change it every time we release a new version.
I would prefer to always generate the docs in the `docs/types.md` file, using Babel's `master` branch. Maybe we coulld add a cli parameter to this script to use a local copy of Babel:
```
node scripts/generate-docs-types-versions.js --path ../babel/packages/babel-types
```

scripts/generate-docs-types.js
```diff
@@ -0,0 +1,287 @@
+const fs = require("fs");
+const path = require("path");
+const t = require("@babel/types");
```
nicolo-ribaudo: Can we make this configurable, to be able to generate them from `master`?

```
node scripts/generate-docs-types.js --path ../babel/packages/babel-types
```

scripts/generate-docs-types.js
```diff
@@ -0,0 +1,287 @@
...
+const { promisify } = require("util");
+
+const readfile = promisify(fs.readFile);
+const writefile = promisify(fs.writeFile);
```
nicolo-ribaudo: You can do `const fs = require("fs").promises` instead

scripts/generate-docs-types-versions.js
```diff
@@ -0,0 +1,72 @@
+const fs = require("fs");
```
nicolo-ribaudo: Do we need this file? Ideally we would only need to generate docs for the new current version in the future.

deificx:
Good suggestions, I'll hopefully get around to picking this up again soon. Time flies.

deificx:
> Good suggestions, I'll hopefully get around to picking this up again soon. Time flies.

It does indeed, but better late than never I've heard. 

The conflicts seem to be where there is additional data about when something was added, I might need to handle that case as well then.

In addition to generating the documents I've also tried to create some examples. They aren't always great, like f.ex. `any extends any ? any : any`. But in general they should give some sort of idea about what will happen when you call the code. It fails in a couple of cases because arguments that are indeed required does not appear so in type. That is something that can be fixed later in separate issues.

PR #2093, opened 2019-09-03
docs(v7-migration): add core-js v3 path for importing `flatMap`

tinymachine:
The `@latest` version of core-js is now v3, which has a different path to specific polyfills than v2.


tinymachine:
You're welcome! Strange thing, though: when I visit the reply preview site, I don't see the update.

https://deploy-preview-2093--babel.netlify.com/docs/en/v7-migration

Not sure why. Any ideas?

PR #2101, opened 2019-09-18
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -8,7 +8,7 @@ id: faq
 In order to comply with the specification, the iterator's return method must be
 called on errors. An alternative is to enable [loose mode](plugin-transform-for-of.md#loose)
 but please note that there are **many** caveats to be aware of if you enable
-loose mode and that you're willingly choosing to be spec non compliant.
+loose mode and that you're willingly choosing not to comply with the spec.
 
 Please see [google/traceur-compiler#1773](https://github.com/google/traceur-compiler/issues/1773) and
 [babel/babel#838](https://github.com/babel/babel/issues/838) for more information.
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -8,7 +8,7 @@ id: faq
 In order to comply with the specification, the iterator's return method must be
 called on errors. An alternative is to enable [loose mode](plugin-transform-for-of.md#loose)
 but please note that there are **many** caveats to be aware of if you enable
-loose mode and that you're willingly choosing to be spec incompliant.
+loose mode and that you're willingly choosing to be spec non compliant.
 
 Please see [google/traceur-compiler#1773](https://github.com/google/traceur-compiler/issues/1773) and
 [babel/babel#838](https://github.com/babel/babel/issues/838) for more information.
--- a/js/repl/ExternalPlugins.js
+++ b/js/repl/ExternalPlugins.js
@@ -53,7 +53,7 @@ export default class ExternalPlugins extends React.Component<Props, State> {
     return (
       <button
         className={currentStyles.modalButton}
-        disbled={isLoading}
+        disabled={isLoading}
         onClick={this.handleOpenModal}
       >
         {isLoading ? "Loading Plugin..." : "Add Plugin"}
--- a/js/repl/ExternalPlugins.js
+++ b/js/repl/ExternalPlugins.js
@@ -178,7 +178,7 @@ const currentStyles = {
     a,
     a:visited {
       color: #fff;
-      cursor: poitner;
+      cursor: pointer;
     }
   `,
 };
--- a/website/blog/2017-10-05-babel-turns-three.md
+++ b/website/blog/2017-10-05-babel-turns-three.md
@@ -10,7 +10,7 @@ share_text: "Babel Turns Three"
 
 > Happy Birthday Babel! 🎂 (Sept 28)
 
-Babel has really come a long way since [Sebastian](https://github.com/kittens) started the project only 3 years ago! A while back it was renamed from [6to5 to Babel](https://babeljs.io/blog/2015/02/15/not-born-to-die); for good reason as it has signficantly contributed to the use of ES2015+ by [many companies](https://babeljs.io/users), libraries, and developers alike.
+Babel has really come a long way since [Sebastian](https://github.com/kittens) started the project only 3 years ago! A while back it was renamed from [6to5 to Babel](https://babeljs.io/blog/2015/02/15/not-born-to-die); for good reason as it has significantly contributed to the use of ES2015+ by [many companies](https://babeljs.io/users), libraries, and developers alike.
 
 <!--truncate-->
 
--- a/website/blog/2017-10-05-babel-turns-three.md
+++ b/website/blog/2017-10-05-babel-turns-three.md
@@ -72,7 +72,7 @@ We've accomplished a lot, but we have to continue to work in multiple ways to ma
 
 ## Future Ideas
 
-- Providing a ["metadata" plugin](https://github.com/babel/notes/issues/34) for users to grep for patterns in their codebases. This will be useful to figure out how people are using certain proposals (similar to what was done in [tc39/proposal-optional-chaining#17](https://github.com/tc39/proposal-optional-chaining/issues/17) for coffeescript regarding `?.`)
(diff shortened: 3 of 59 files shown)


js/repl/ExternalPlugins.js
```diff
@@ -178,7 +178,7 @@ const currentStyles = {
     a,
     a:visited {
       color: #fff;
-      cursor: poitner;
+      cursor: pointer;
```
jsoref: This is a behavior change...

js/repl/ExternalPlugins.js
```diff
@@ -53,7 +53,7 @@ export default class ExternalPlugins extends React.Component<Props, State> {
     return (
       <button
         className={currentStyles.modalButton}
-        disbled={isLoading}
+        disabled={isLoading}
```
jsoref: This is a behavior change

website/blog/2018-07-19-whats-happening-with-the-pipeline-proposal.md
```diff
@@ -149,7 +149,7 @@ promise |> await #;
 
 If there are any operators, parentheses (including for method calls), brackets, or anything other than identifiers and dot punctuators, then a topic token is necessary. This avoids footguns and eliminate ambiguity when not using a topic token.
 
-Smart pipelines thus resolve the issue of async in an integrative way, allowing all possible expressions to be emedded in a pipeline; not only `await`, but also `typeof`, `yield`, and another other operator desired.
+Smart pipelines thus resolve the issue of async in an iterative way, allowing all possible expressions to be embedded in a pipeline; not only `await`, but also `typeof`, `yield`, and another other operator desired.
```
jsoref: I'm not sure `iterative` is the right way. I changed a similarly misspelled word to `integrate`...

website/blog/2017-12-27-nearing-the-7.0-release.md
```diff
@@ -61,9 +61,9 @@ Please reach out to help and to post issues when trying to update! This is a gre
 - Updated [TC39 proposals](https://github.com/babel/proposals/issues)
   - Numeric Separator: `1_000`
   - Optional Chaining Operator: `a?.b`
-  - `import.meta` (parseble)
+  - `import.meta` (parsable)
```
jsoref: I'm assuming this is a correct change

docs/faq.md
```diff
@@ -8,7 +8,7 @@ id: faq
 In order to comply with the specification, the iterator's return method must be
 called on errors. An alternative is to enable [loose mode](plugin-transform-for-of.md#loose)
 but please note that there are **many** caveats to be aware of if you enable
-loose mode and that you're willingly choosing to be spec incompliant.
+loose mode and that you're willingly choosing to be spec non compliant.
```
nicolo-ribaudo: This doesn't seem correct

website/blog/2018-07-19-whats-happening-with-the-pipeline-proposal.md
nicolo-ribaudo: cc @mAAdhaTTah

docs/faq.md
jsoref: "choosing not to comply with the spec"

nicolo-ribaudo: Yeah, better :+1:

website/blog/2018-07-19-whats-happening-with-the-pipeline-proposal.md
mAAdhaTTah: In this case, no, "integrative" was intended. The idea is the Smart Pipelines solves async by solving all expressions in pipelines (vs. F#-style, which has one-off special handling for `await`). "Holistic" could be an alternative word choice, or we could rework this sentence so it's clearer, but "iterative" is not the intended meaning here.

jsoref: Would `integrated` work?

I think I'll go with `holistic`, that at least makes some sense, although it'd be nice if you could explain a bit more here.

mAAdhaTTah: To my ear, "integrative" sounded better, but I wasn't really sure why. I checked out the definitions of each:


>integrated | ˈin(t)əˌɡrādəd |
>adjective
>with various parts or aspects linked or coordinated: an integrated and high-quality public transportation system.


>integrative | ˈintəɡrādiv |
>adjective
>serving or intending to unify separate things: an integrative approach to learning.

I _think_ "integrative" matches better what I'm trying to get at here. There are several operators that can be embedded into a Smart pipeline. Instead of solving for each of them individually, we solve for them by solving "expressions in the pipeline" as a whole, hence "unify[ing] separate things." They're not so much "linked or coordinated" –they're not working with each other, exactly, so much as being handled as a unified whole, as expressions.

This is in contrast with F#, which solves it with a one-off implementation for `await`.  I always considered the `await` usage in F# to be as if you were treating it like a unary function, even though that's not really what it is, but it doesn't treat expressions as a unified hole.

Holistic kinda works too:

>holistic | hōˈlistik |
>adjective
>chiefly Philosophy characterized by comprehension of the parts of something as intimately interconnected and explicable only by reference to the whole.

Smart treats all expressions as a single whole, rather than F# treating await like a single component part. That said, having read the definitions, I think "integrative" makes the most sense, but given that I had to look it up, it's not really easing the clarity of my communication here. We could perhaps rewrite this graf to be something like:

>Smart pipelines thus resolve the issue of async by solving for expressions as a whole, enabling any of them to be embedded in a pipeline; not only `await`, but also `typeof`, `yield`, and any other operator desired.

Thoughts?

existentialism: I'm fine with leaving it `integrative`

website/blog/2017-10-05-babel-turns-three.md
```diff
@@ -72,7 +72,7 @@ We've accomplished a lot, but we have to continue to work in multiple ways to ma
 
 ## Future Ideas
 
-- Providing a ["metadata" plugin](https://github.com/babel/notes/issues/34) for users to grep for patterns in their codebases. This will be useful to figure out how people are using certain proposals (similar to what was done in [tc39/proposal-optional-chaining#17](https://github.com/tc39/proposal-optional-chaining/issues/17) for coffescript regarding `?.`)
+- Providing a ["metadata" plugin](https://github.com/babel/notes/issues/34) for users to grep for patterns in their codebases. This will be useful to figure out how people are using certain proposals (similar to what was done in [tc39/proposal-optional-chaining#17](https://github.com/tc39/proposal-optional-chaining/issues/17) for coffeescript regarding `?.`)
```
existentialism: Since we're already touching it:

```suggestion
- Providing a ["metadata" plugin](https://github.com/babel/notes/issues/34) for users to grep for patterns in their codebases. This will be useful to figure out how people are using certain proposals (similar to what was done in [tc39/proposal-optional-chaining#17](https://github.com/tc39/proposal-optional-chaining/issues/17) for CoffeeScript regarding `?.`)
```

website/blog/2018-07-27-removing-babels-stage-presets.md
```diff
@@ -16,7 +16,7 @@ We didn't make this decision lightly and wanted to show the context behind the i
 
 ## Some History
 
-A Babel preset is a shareable list of plugins.
+A Babel preset is a sharable list of plugins.
```
existentialism: Not a huge deal, but `shareable` seems fine?

website/blog/2018-07-19-whats-happening-with-the-pipeline-proposal.md
jsoref: Dropped. It's odd, I don't think I've ever encountered this word. I did a search w/ google and it seems to be mostly used in Health/Medicine.

website/blog/2018-07-27-removing-babels-stage-presets.md
jsoref: Both were used. I tend to pick the more dominant spelling w/in a repo or the one that would be dominant for a region. In this case, the short form was used 2:1. Notably, the short form is an en-US variant (and from memory this repo feels en-GB), and it looks like it's on the receding side. I'm happy to switch the `sharable`s to `shareable`.

existentialism: Yeah... we're wildly inconsistent there, either way, not a big deal 😄

existentialism:
@jsoref thanks!

PR #2108, opened 2019-10-16
Update plugin-proposal-logical-assignment-operators.md

scplay:
add nullish coalescing assignment example in / out


nicolo-ribaudo:
Could you put it in a separate example and write that the nullish coalescing operator plugin must be enabled?

PR #2114, opened 2019-10-26
Add docs parser error recovery

nicolo-ribaudo:
https://github.com/babel/babel/pull/10363


hzoo:
cc @duailibe if prettier was interested in this once it's out

PR #2117, opened 2019-10-30
fix: guess file extension from preset

### LINKED ISSUE #1666: Include file extension in REPL config options

JLHwung:
In this PR we guess the REPL file extension from the loaded presets. By doing so we could apply proper TypeScript transform when `typescript` preset is enabled.

Compared to #2019, this PR does _not_ introduce `filename` option as we don't output error message that will point to `filename` configuration when parsing a TypeScript without transform enabled. Also we are not going to expose `filename` option in a such limited UI space.

Manually tested on local build.

Closes #2019 
Fixes #1666

### DIFF
--- a/js/repl/types.js
+++ b/js/repl/types.js
@@ -141,3 +141,5 @@ export type Transition = {
   visitorType: string,
   currentNode?: string,
 };
+
+export type SupportedFileExtension = ".js" | ".jsx" | ".ts" | ".tsx";
--- a/js/repl/compile.js
+++ b/js/repl/compile.js
@@ -6,7 +6,12 @@ declare var prettier: any;
 declare var prettierPlugins: any;
 
 import Transitions from "./Transitions";
-import type { CompileConfig, Transition } from "./types";
+import type {
+  BabelPresets,
+  CompileConfig,
+  Transition,
+  SupportedFileExtension,
+} from "./types";
 
 type Return = {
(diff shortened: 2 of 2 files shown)


js/repl/types.js
```diff
@@ -141,3 +141,5 @@ export type Transition = {
   visitorType: string,
   currentNode?: string,
 };
+
+export type SupportedFileExtension = ".js" | ".jsx" | ".ts" | ".tsx";
```
JLHwung: Do we need `.mjs` when `config.sourceType === "module"`?

xtuc: That would be more consistent with the Node.js experience, I would say yes.

nicolo-ribaudo: Well, it isn't observable to users.

JLHwung: True. Given that we don't have `.mjs` specific transforms in the preset, we don't have to add it here.

PR #2122, opened 2019-11-06
Fix uncleared floating videos

arku:
### Stuff done

- Clear floated videos responsively
- Uses `calc` to define the video container's widths instead of hardcoding.

##### Before

image

##### After

image

##### Before

image

##### After

image

/cc @nicolo-ribaudo

### DIFF
--- a/website/static/css/video.css
+++ b/website/static/css/video.css
@@ -20,7 +20,7 @@
   margin-right: 5px;
   box-sizing: border-box;
   float: left;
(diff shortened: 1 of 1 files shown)


existentialism:
Thanks!

As a follow up, we should consider using flex + `align-items: stretch` to get rid of this case:

image

arku:
@existentialism Thanks for the quick review!

Yeah, we should migrate to flexbox.

PR #2125, opened 2019-11-08
Update "Babel's Funding Plans" blog post to include Sebastian employment

sebmck:
I don't think this was intentional but it was disappointing to see the financial contributions of the companies I've worked for ignored, which I would consider fairly substantial since Babel would be nothing without it. It represents what was then a sustainable model for my work on Babel.

Happy to make changes, and discuss whether I am even considered a "maintainer" but the project and funding did not start when Henry quit his job.

### DIFF
--- a/website/blog/2019-11-08-babels-funding-plans.md
+++ b/website/blog/2019-11-08-babels-funding-plans.md
@@ -83,6 +83,7 @@ And long term:
 
 We know that as maintainers we need to do outreach for our project and for open source as a whole. Our efforts up to this point have consisted of the following:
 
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Yeah this is fair. IIRC also @jridgewell had some working hours dedicated to Babel?

jridgewell:
I have 20% at Google, but that's split between my all my OSS work, TC39, and V8 devrel things. So, unfortunately, not a lot is dedicated to Babel.

hzoo:
Yeah it wasn't intentional, added that section at the end. We could also add Amjad/Yuriy (they worked on Babel/minify) for work as well right?

I think we can still include that Justin, since it doesn't have to be the current thing but just what we had in the past

hzoo:
I guess we weren't going to add all the history but give an example of some things? Also fine with removing it too.

hzoo:
Would you want to start up something like a HISTORY.md @sebmck? Could be a good reference for others (and us) in terms of things like when big things happened, changes in scope/thinking, etc

PR #2130, opened 2019-11-21
Enable `darcula` theme in REPL when dark colour scheme is preferred

JLHwung:
In this PR we bump CodeMirror to 5.49.2 so that the dark theme `darcula` is included. When the UA supports `(prefers-color-scheme)` media query, we will enable `darcula` theme in the CodeMirror panel, otherwise it will fallback to the `default` theme.

I am open to other [dark themes](https://codemirror.net/demo/theme.html#default) as long as we can reach consensus. In this PR `darcula` is chosen because it is the default dark scheme available in WebStorm.

Screenshots:
image

### DIFF
--- a/js/repl/styles.js
+++ b/js/repl/styles.js
@@ -7,6 +7,7 @@ const colors = {
   errorForeground: "#ff0000",
 
   foregroundLight: "#aaaaaa",
+  foregroundDark: "#777777",
 
   // Inspired by: Nuclide's "One Dark" theme
   inverseBackground: "#21252b",
--- a/js/repl/styles.js
+++ b/js/repl/styles.js
@@ -1,25 +1,46 @@
 // @flow
+import { preferDarkColorScheme } from "./Utils";
 
-const colors = {
+const lightColors = {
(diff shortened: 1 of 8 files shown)


js/repl/styles.js
```diff
@@ -12,14 +13,35 @@ const colors = {
   inverseBackground: "#21252b",
   inverseBackgroundDark: "#181a1f",
   inverseBackgroundLight: "#292d36",
-  inverseForeground: "#ffffff",
+  inverseForeground: "#faffff",
```
JLHwung: The `inverseForeground` is modified from `#fff` to be `lab(100, 0, -5)` so that it shares the same colour channel with the `inverseBackground` which is `lab(15, 0, -5)`.

existentialism:
I'm fine with landing this, though I don't think the sidebar matches the color scheme... but that's a different problem as it already doesn't match the header bar.

JLHwung:
> I like that theme.
> 
> If I enable dark-mode it won't be used until the website is reloaded, but it's probably ok.

Yeah, that will be better, we can address it in a separate PR.

PR #2132, opened 2019-11-22
Update dead link & add description

zero41120:
Original link to the great reading was dead.
Update to the current master branch link.
Mark the manual navigation location in case the link dies again.


nicolo-ribaudo:
Thanks!

PR #2144, opened 2019-12-11
Add note about the conflict with useBuiltins

tarjei:
Hi, this is a stab at #10121. 

As I understand it, preset-env with useBuiltins set to `usage` conflicts with using the runtime helper, though it is not totally clear to me how.


nicolo-ribaudo:
The problem is only when using `useBuiltIns` and `@babel/plugin-transform-runtime`'s `corejs` option, because it's not predictable which of the two polyfills would be applied to the different JavaScript features.

Using `@babel/plugin-transform-runtime` only for the helpers is safe.

Issue #2142, opened 2019-12-04
Update Optional Chaining documentation to reflect the return from undefined to true


mpaarating:
Per https://github.com/babel/babel/issues/10805 and the subsequent [PR](https://github.com/babel/babel/pull/10806) , When using optional delete to delete a property of a nullish base, the delete operation returns now returns true. Should a distinction be made in the docs to support this change?

JLHwung:
@mpaarating Good catch! PR's welcome!

mpaarating:
Great, I'll start working on this ASAP!

mpaarating:
@JLHwung After looking into the docs for `@babel/plugin-proposal-optional-chaining`, there isn't a mention of what the current behavior is (undefined return for nullish base). As such, I don't think that additional documentation need to be added. What are your thoughts?

JLHwung:
@mpaarating I think it will be great to add a section like _delete nested properties_ because the behaviour is changed after 7.8 (to be aligned to the spec) and people may reply on the old behaviour. So it should be clarified in the docs.

PR #2145, opened 2019-12-13
Update optional chaining documentation with deletion change

mpaarating:
# Description
Per #2142, the `7.8` release includes a change to the return value of a `delete` operation on a nullish base. Previously, it had returned `undefined`, but now returns `true`. This PR adds documentation that indicates this change.

## Screenshot
[Image from Gyazo](https://gyazo.com/8fd017b150ea3f333c6652c807c873d5)


nicolo-ribaudo:
Thanks!

PR #2147, opened 2019-12-22
parser.md: fix the name of the range property

thorn0:
The property is named `range`, not `ranges`: https://github.com/babel/babel/blob/master/packages/babel-parser/src/parser/node.js#L16


nicolo-ribaudo:
Thanks!

PR #2155, opened 2020-01-11
Update link to plugin-features.js

andersk:
It was moved in babel/babel#10899.

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -195,7 +195,7 @@ An array of plugins to always include.
 
 Valid options include any:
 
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
Thanks! Could you also update it in the `website/versioned_docs` folder?

andersk:
How do you want `versioned_docs` to be treated? It seems wrong for those to ever link to `babel/{tree,blob}/master` since that doesn’t reflect the version they’re supposed to be for. Should `version-7.4.0` link to `babel/{tree,blob}/v7.4.0`? Or `v7.4.5`? Or some kind of `v7.4.x` branch that doesn’t seem to exist there but maybe should? Am I overthinking this?

PR #2150, opened 2020-01-06
Unify docs about .mjs, .cjs, .js and .json config files

nicolo-ribaudo:
Docs for https://github.com/babel/babel/pull/10903 and https://github.com/babel/babel/pull/10783, follow up to https://github.com/babel/website/pull/2106.

PR #2156, opened 2020-01-12
Add 7.8 in version

rajasekarm:
Updated latest version.


nicolo-ribaudo:
Thanks!
We actually have a script in the `website/` package which bumps the version and generates the versioned docs: `cd website && npm run version`.

existentialism:
FWIW, I would have prefer we ran the script instead of landing this... as it's nice to have _just_ the files that changed from version to version instead of every doc getting added to the `version-7.8.0` folder (like https://github.com/babel/website/tree/master/website/versioned_docs/version-7.7.0).

nicolo-ribaudo:
The script copied every file because in https://github.com/babel/website/pull/2150 I modified every doc.
I should work on backporting the `### With a configuration file (Recommended)` changes and then re-run the script to clean up the 7.8.0 folder.

existentialism:
Ahh... yeah, not critical, just a nicety.

PR #2157, opened 2020-01-16
Fix preset-env when using Babel 7.8.0+

nicolo-ribaudo:
Try opening this url in a private window, to have a clean state: https://babeljs.io/repl/#?presets=env&version=7.8.3

You will see an error message about `@babel/preset-env` being registered twice. This is because `@babel/standalone` already includes it, and then we are re-loading it.

In this PR I chose to load `env-standalone` together with `@babel/standalone` when using an older version. This will increase the loading time for old Babel version, but it improves removes the `isLoaded`/`isLoading` logic which isn't needed anymore with the new bundled preset.

**EDIT**: ~~It still doesn't work.~~
https://deploy-preview-2157--babel.netlify.com/repl#?presets=env&version=7.8.3 Try enabling and disabling the preset a few times.

**EDIT**: It's working now

### DIFF
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -295,7 +295,6 @@ class Repl extends React.Component<Props, State> {
         babelState.errorMessage =
           babelState.errorMessage || envState.errorMessage;
       } else {
-        alert("REG!");
         await this._workerApi.registerEnvPreset();
       }
     }
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -98,6 +98,21 @@ function toCamelCase(str) {
     });
 }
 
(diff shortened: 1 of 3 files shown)


js/repl/Repl.js
```diff
@@ -275,7 +286,19 @@ class Repl extends React.Component<Props, State> {
...
+
+    if (compareVersions(babelState.version, "7.8.0") == -1) {
+      const envState = await this._loadPresetEnvStandalone();
+
+      if (envState.didError) {
+        babelState.didError = true;
+        babelState.errorMessage =
+          babelState.errorMessage || envState.errorMessage;
+      } else {
+        alert("REG!");
```
nicolo-ribaudo: Todo: remove

PR #2158, opened 2020-01-18
fix: fix t.jsxFunction to t.jSXFunction

AshoneA:
Is this jsxFunction wrong that jsx use all lower-case? When I copy t.jsxFunction always throw a not function error, I'm new about babel, or use lower-case have other means?

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -1127,7 +1127,7 @@ Aliases: `Flow`, `FlowType`
 
 ### jSXAttribute
 ```javascript
-t.jsxAttribute(name, value)
+t.jSXAttribute(name, value)
 ```
 
 See also `t.isJSXAttribute(node, opts)` and `t.assertJSXAttribute(node, opts)`.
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
They are defined: https://github.com/babel/babel/blob/341964bd4e21af6b4e42f500431ef4ad192f0f20/packages/babel-types/src/builders/generated/index.js#L559

The `jSX*` version will be removed soon (Babel 8.0.0), and I don't think that we should document it.

`jsx*` have been introduced in Babel 7 (https://github.com/babel/babel/pull/6967). Maybe you are still using an older version?

AshoneA:
> They are defined: https://github.com/babel/babel/blob/341964bd4e21af6b4e42f500431ef4ad192f0f20/packages/babel-types/src/builders/generated/index.js#L559
> 
> The `jSX*` version will be removed soon (Babel 8.0.0), and I don't think that we should document it.
> 
> `jsx*` have been introduced in Babel 7 ([babel/babel#6967](https://github.com/babel/babel/pull/6967)). Maybe you are still using an older version?

Oh, gosh! I should install @babel/types instead of babel-types, thank you for your answer!

PR #2160, opened 2020-01-20
Cleanup 7.8.0 versioned_docs

nicolo-ribaudo:
https://github.com/babel/website/pull/2156#issuecomment-574408195

The two commits should be reviewed one-by-one.
The first backports all the changes related to the "Usage - .babelrc" headers introduced in #2150.
In the second commit I re-run `yarn run version` starting from a clean state and only keep the new generated files.


existentialism:
:shipit: !

PR #2161, opened 2020-01-21
Refine the descriptions of -i and -o options in babel-node

exKAZUu:
The current implementation (https://github.com/babel/babel/blob/master/packages/babel-register/src/node.js#L141-L144) ignores any file outside current working directory.
I feel this behavior without any documentation is confusing. Actually, an issue (https://github.com/babel/babel/issues/7566) is previously created due to the lack of documentation.


nicolo-ribaudo:
Could you backport this to `website/versioned_docs/version-7.8.0/node.md` and `website/versioned_docs/version-7.0.0/node.md`?

exKAZUu:
Thank you for the review. I've backported it.

nicolo-ribaudo:
Last thing, sorry for not noticing it when I first reviewed this PR :sweat_smile:

Looking at the source code I noticed that the default behavior is only used when neither `--ignore` not `--only` are specified:
https://github.com/babel/babel/blob/2057d2b159b0ad1376c9701206cf59419369be75/packages/babel-register/src/node.js#L140

Could you mention it? (Also it would be good to write a similar description for `--only`, even if it wasn't documented yet)

exKAZUu:
@nicolo-ribaudo I've added the description of `-o` option.

docs/node.md
```diff
@@ -80,7 +80,8 @@ npx babel-node --debug --presets es2015 -- script.js --debug
...
 | `-e, --eval [script]`  |                             | Evaluate script                                                                                                                             |
 | `-p, --print`          |                             | Evaluate script and print result                                                                                                            |
-| `-i, --ignore [regex]` | `node_modules`              | Ignore all files that match this regex when using the require hook                                                                          |
+| `-i, --ignore [regex]` | `node_modules`              | Ignore all files that match this regex when using the require hook. If both -i and -o are omitted, any file outside current working directory is ignored too. |
+| `-o, --only [regex]`   | current working directory   | Include all files that match this regex when using the require hook. If both -i and -o are omitted, any file outside current working directory is ignored too. |
```
nicolo-ribaudo: Nit: `Only include the files` would be more explicit, since `Include all files` doesn't imply that what is not matched is excluded. Apart from that, the desc looks good :+1:

exKAZUu: I see. How about `Include only the files that match this regex when using the require hook and exclude the other files. If both -i and -o are omitted, any file outside current working directory is ignored too.` ?

(I'm not a native English speaker, so if you have a good alternative, please share it. Sorry in advance.)

nicolo-ribaudo: I'm not a native speaker either :stuck_out_tongue: 

Your proposed description seems ok to me.

docs/node.md
```diff
@@ -80,7 +80,8 @@ npx babel-node --debug --presets es2015 -- script.js --debug
 | ---------------------- | --------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- |
 | `-e, --eval [script]`  |                             | Evaluate script                                                                                                                             |
 | `-p, --print`          |                             | Evaluate script and print result                                                                                                            |
-| `-i, --ignore [regex]` | `node_modules`              | Ignore all files that match this regex when using the require hook                                                                          |
+| `-i, --ignore [regex]` | `node_modules`              | Ignore all files that match this regex when using the require hook. If both -i and -o are omitted, any file outside current working directory is ignored too. |
```
existentialism:
```suggestion
| `-i, --ignore [regex]` | `node_modules`              | Ignore all files that match this regex when using the require hook. If both `-i` and `-o` are omitted, files outside the current working directory are also ignored. |
```

docs/node.md
```diff
@@ -80,7 +80,8 @@ npx babel-node --debug --presets es2015 -- script.js --debug
...
+| `-o, --only [regex]`   | current working directory   | Include only the files that match this regex when using the require hook and exclude the other files. If both -i and -o are omitted, any file outside current working directory is ignored too. |
```
existentialism:
```suggestion
| `-o, --only [regex]`   | current working directory   | Only include (and exclude all other) files that match this regex when using the require hook. If both `-i` and `-o` are omitted, files outside the current working directory are also ignored. |
```

existentialism:
Thanks @exKAZUu!

PR #2168, opened 2020-02-03
docs: Update MatchPattern function description

antonku:
The purpose of this PR is to update the description of `MatchPattern` function, specifically:
- Change `callee` parameter name to `caller`
- Describe `dirname` property that is a part of the `context`

Link to the source code that confirms the validity of these updates:

https://github.com/babel/babel/blob/779f00b8e9ce568ece4859501dd79682798c7fbf/packages/babel-core/src/config/config-chain.js#L698-L700

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -650,7 +650,7 @@ A root path to include on generated module names.
 
 ### `MatchPattern`
 
-Type: `string | RegExp | (filename: string | void, context: { callee: { name: string } | void, envName: string ) => boolean`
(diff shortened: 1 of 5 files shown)


antonku:
Here is a repo with the test that proofs that the correction suggested in the PR is applicable to different `@babel/core@7.x` versions: https://github.com/antonku/matchpattern-test

You can verify it by cloning the repository, installing the dependencies and running `npm run test`

PR #2173, opened 2020-02-12
docs: add caveats for commonjs modules transform

JLHwung:
Closes https://github.com/babel/babel/issues/11127

### DIFF
--- a/docs/plugin-transform-modules-amd.md
+++ b/docs/plugin-transform-modules-amd.md
@@ -4,7 +4,7 @@ title: @babel/plugin-transform-modules-amd
 sidebar_label: transform-modules-amd
 ---
 
-This plugin transforms ECMAScript modules to [AMD](https://github.com/amdjs/amdjs-api/blob/master/AMD.md). The _syntax_ of Import/Export Statement (`import "./mod.js"`) and Import Expression (`import('./mod.js')`) are transformed to their AMD representation. Note that Babel is unaware of different resolution algorithms between implementations of ECMAScript modules and AMD.
+This plugin transforms ECMAScript modules to [AMD](https://github.com/amdjs/amdjs-api/blob/master/AMD.md). Note that only the _syntax_ of import/export statements (`import "./mod.js"`) and import expressions (`import('./mod.js')`) is transformed, as Babel is unaware of the different resolution algorithms between implementations of ECMAScript modules and AMD.
 
 ## Example
 
@@ -61,4 +61,3 @@ require("@babel/core").transform("code", {
 ### Options
 
 See options for `@babel/plugin-transform-modules-commonjs`.
-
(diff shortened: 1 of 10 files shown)


SimenB:
This doesn't fully cover it though - I'd be specific about `URL`. Both node and the browser (at least Chrome) supports it, but transpilation breaks it. Linking Node docs doesn't really tell the whole picture.

nicolo-ribaudo:
We should mention that Babel only transforms the syntax of import statements/expressions and that it doesn't know about the resolution algorithms used by the different platforms (Node.js, in this case).

SimenB:
That approach seems better, yeah 👍

JLHwung:
@nicolo-ribaudo We don't have a dedicated plugin page for `proposal-dynamic-import`: https://babeljs.io/docs/en/babel-plugin-proposal-dynamic-import I suggest we create it in another PR if we are good with the phrasing.

docs/plugin-transform-modules-amd.md
```diff
@@ -4,6 +4,8 @@ title: @babel/plugin-transform-modules-amd
 sidebar_label: transform-modules-amd
 ---
 
+This plugin transforms ECMAScript modules to [AMD](https://github.com/amdjs/amdjs-api/blob/master/AMD.md). The _syntax_ of Import/Export Statement (`import "./mod.js"`) and Import Expression (`import('./mod.js')`) are transformed to their AMD representation. Note that Babel is unaware of different resolution algorithms between implementations of ECMAScript modules and AMD.
```
existentialism: Maybe:

```suggestion
This plugin transforms ECMAScript modules to [AMD](https://github.com/amdjs/amdjs-api/blob/master/AMD.md). Note that only the _syntax_ of import/export statements (`import "./mod.js"`) and import expressions (`import('./mod.js')`) is transformed, as Babel is unaware of the different resolution algorithms between implementations of ECMAScript modules and AMD.
```

?

docs/plugin-transform-modules-umd.md
```diff
@@ -4,6 +4,10 @@ title: @babel/plugin-transform-modules-umd
 sidebar_label: transform-modules-umd
 ---
 
+This plugin transforms ES2015 modules to [UMD](https://github.com/umdjs/umd). Note that only the _syntax_ of import/export statements (`import "./mod.js"`) is transformed, as Babel is unaware of different resolution algorithms between implementations of ES2015 modules and UMD.
+
+:warning This plugin does not support dynamic import (`import('./lazy.js')`).
```
nicolo-ribaudo: Nit: the warning sign isn't properly displayed, I think that you should directly paste the emoji.

nicolo-ribaudo:
@JLHwung Docusaurus doesn't support GitHub emojis, you need to copy-paste the warning character :stuck_out_tongue: 
https://deploy-preview-2173--babel.netlify.com/docs/en/babel-plugin-transform-modules-umd

PR #2175, opened 2020-02-14
Add corejs option to preset env

danez:
Simply adds a new option to select which core-js version to target.

Screenshot 2020-02-14 at 13 26 20

### DIFF
--- a/js/repl/ReplOptions.js
+++ b/js/repl/ReplOptions.js
@@ -502,6 +502,19 @@ class ExpandedContainer extends Component<Props, State> {
                   Built-ins
                 </LinkToDocs>
                 <select
+                  value={envConfig.corejs}
+                  className={styles.envPresetSelect}
+                  onChange={this._onEnvPresetSettingChange("corejs")}
+                  disabled={
+                    !envConfig.isEnvPresetEnabled ||
+                    !envConfig.isBuiltInsEnabled ||
+                    runtimePolyfillState.isEnabled
+                  }
+                >
+                  <option value="2">core-js 2</option>
+                  <option value="3.6">core-js 3.6</option>
+                </select>
+                <select
                   value={envConfig.builtIns}
                   className={styles.envPresetSelect}
                   onChange={this._onEnvPresetSettingChange("builtIns")}
@@ -998,6 +1011,7 @@ const styles = {
     maxWidth: "7rem",
     fontWeight: 400,
     color: colors.textareaForeground,
+    margin: "0 0 0 0.75rem",
 
     "&:disabled": {
       opacity: 0.5,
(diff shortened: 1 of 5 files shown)


js/repl/compile.js
```diff
@@ -58,6 +58,7 @@ export default function compile(code: string, config: CompileConfig): Return {
   let useBuiltIns = false;
   let spec = false;
   let loose = false;
+  let corejs = "2";
```
JLHwung:
```suggestion
  let corejs = "3";
```
nit: let's make the defaults be modern

js/repl/ReplOptions.js
```diff
@@ -501,6 +501,19 @@ class ExpandedContainer extends Component<Props, State> {
...
+                  className={styles.envPresetSelect}
+                  onChange={this._onEnvPresetSettingChange("corejs")}
+                  disabled={
+                    !envConfig.isEnvPresetEnabled ||
+                    !envConfig.isBuiltInsEnabled ||
+                    runtimePolyfillState.isEnabled
+                  }
+                >
+                  <option value="2">core-js 2</option>
+                  <option value="3">core-js 3</option>
```
JLHwung: I feel like it is worthy to add
```html
<option value="3.6">core-js 3.6</option>
```
or simply replace core-js 3 by core-js 3.6 because 3.6 has more features support than core-js 3.0

danez: Oh I didn't know this makes a difference.

js/repl/ReplOptions.js
```diff
@@ -501,6 +501,19 @@ class ExpandedContainer extends Component<Props, State> {
...
+                  className={styles.envPresetSelect}
+                  onChange={this._onEnvPresetSettingChange("corejs")}
+                  disabled={
+                    !envConfig.isEnvPresetEnabled ||
+                    !envConfig.isBuiltInsEnabled ||
+                    runtimePolyfillState.isEnabled
+                  }
+                >
+                  <option value="2">core-js 2</option>
+                  <option value="3.6">core-js 3</option>
```
nicolo-ribaudo: I think that the UI should show 3.6, to make it clearer that the corresponding option is `corejs: "3.6"` and not `corejs: 3`.

js/repl/replUtils.js
```diff
@@ -155,6 +155,7 @@ export const persistedStateToEnvConfig = (
     node: envPresetDefaults.node.default,
     version: persistedState.version,
     builtIns: envPresetDefaults.builtIns.default,
+    corejs: "2",
```
nicolo-ribaudo: Why hard-coded?

danez: I changed it to do the same as the builtins option.

js/repl/ReplOptions.js
danez: changed.

PR #2181, opened 2020-02-15
Update usage.md

76784:
Correct the file name extension to reflect the content.


nicolo-ribaudo:
I'd prefer to edit the example rather than the filename, since `babel.config.json` has the benefit that it's simpler.

PR #2185, opened 2020-02-28
docs: add onlyRemoveTypeImports to (preset|plugin-transform)typescript

kaicataldo:
Refs https://github.com/babel/babel/pull/11173#issuecomment-591707329

This PR adds documentation for the new `onlyRemoveTypeImports` configuration option in the upcoming 7.9.0 release.

Questions:
1. This is my first time working in this repo. Did I miss anything?
2. I fixed up the docs so that they matched and did a little editing. Is it okay that I didn't backport these changes to the older versions of the docs?

### DIFF
--- a/docs/plugin-transform-typescript.md
+++ b/docs/plugin-transform-typescript.md
@@ -153,19 +153,27 @@ If you have existing code which uses the TypeScript-only [namespace][namespace] 
 
 `boolean`, defaults to `false`
 
+Forcibly enables `jsx` parsing. Otherwise angle brackets will be treated as typescript's legacy type assertion `var foo = <string>bar;`. Also, `isTSX: true` requires `allExtensions: true`.
+
 ### `jsxPragma`
 
 `string`, defaults to `React`
 
-Replace the function used when compiling JSX expressions.
-
-This is so that we know that the import is not a type import, and should not be removed
+Replace the function used when compiling JSX expressions. This is so that we know that the import is not a type import, and should not be removed.
 
 ### `allowNamespaces`
 
 `boolean`, defaults to `false` but will default to `true` in the [future](https://github.com/babel/notes/blob/master/2019/05/21.md#prs).
 
-> You can read more about configuring plugin options [here](https://babeljs.io/docs/en/plugins#plugin-options)
+Enables compilation of TypeScript namespaces.
+
+### `onlyRemoveTypeImports`
+
+`boolean`, defaults to `false`
+
+When set to `true`, the transform will only remove [type-only imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-exports) (introduced in TypeScript 3.8). This should only be used if you are using TypeScript >= 3.8.
+
+> You can read more about configuring plugin options [here](https://babeljs.io/docs/en/plugins#plugin-options).
 
 ## TypeScript Compiler Options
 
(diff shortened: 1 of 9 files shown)


existentialism:
> I fixed up the docs so that they matched and did a little editing. Is it okay that I didn't backport these changes to the older versions of the docs?

It's not critical, but I personally think we should backport :)

kaicataldo:
@existentialism Which parts do you think we should backport? Everything but the new option?

nicolo-ribaudo:
We should also add the equivalent tsc option to the "TypeScript Compiler Options" section.

docs/plugin-transform-typescript.md
```diff
@@ -198,6 +206,8 @@ equivalents in Babel can be enabled by some configuration options or plugins.
   ```
 - `--importHelpers`
   This is the equivalent of the `@babel/plugin-transform-runtime` package.
+- `---importsNotUsedAsValues`
+  You can the `onlyRemoveTypeImports` option to replicate this behavior. `onlyRemoveTypeImports: true` is equivalent to `importsNotUsedAsValues: preserve`, while `onlyRemoveTypeImports: false` is equivalent to `importsNotUsedAsValues: remove`. There is no equivalent for `importsNotUsedAsValues: error`.
```
kaicataldo: Does this seem correct/clear?

kaicataldo:
Thanks for the comments! Ready for re-review :)

docs/plugin-transform-typescript.md
nicolo-ribaudo: Maybe we can't mention that `onlyRemoveTypeImports: false` isn't 100% accurate because we don't do any type analysis, and link to https://github.com/babel/babel/issues/8361 as an example? Also, it's the reason why we don't support `importsNotUsedAsValues: error`.

kaicataldo: Do you mind clarifying what you mean?

nicolo-ribaudo: Sure!

Consider this example:
```ts
// dep.ts

export type T = string;
```
```ts
// main.ts

import { T } from "./dep";
export { T };
```

`onlyRemoveTypeImports: false` leave `main.ts` as-is, because we don't have the type information necessary to understand that `T` is a type.
However, TypeScript's `importsNotUsedAsValues: remove` transforms it to an empty file because it knows that `T` is a type.

docs/plugin-transform-typescript.md
```diff
@@ -153,19 +153,27 @@ If you have existing code which uses the TypeScript-only [namespace][namespace]
 
 `boolean`, defaults to `false`
 
+Forcibly enables `jsx` parsing. Otherwise angle brackets will be treated as typescript's legacy type assertion `var foo = <string>bar;`. Also, `isTSX: true` requires `allExtensions: true`.
```
existentialism: Nit:

```suggestion
Forcibly enables `jsx` parsing. Otherwise angle brackets will be treated as TypeScript's legacy type assertion `var foo = <string>bar;`. Also, `isTSX: true` requires `allExtensions: true`.
```

docs/plugin-transform-typescript.md
existentialism:
```suggestion
  You can use the `onlyRemoveTypeImports` option to replicate this behavior. `onlyRemoveTypeImports: true` is equivalent to `importsNotUsedAsValues: preserve`, while `onlyRemoveTypeImports: false` is equivalent to `importsNotUsedAsValues: remove`. There is no equivalent for `importsNotUsedAsValues: error`.
```

kaicataldo:
Thanks for updating this!

PR #2201, opened 2020-03-13
add new options to `plugin-transform-react-jsx` docs

lunaruan:
This PR adds `automatic | classic` runtime options and `importSource` to the `plugin-transform-react-jsx` docs.

### DIFF
--- a/docs/plugin-transform-react-jsx.md
+++ b/docs/plugin-transform-react-jsx.md
@@ -8,7 +8,7 @@ sidebar_label: transform-react-jsx
 
 ### React Automatic Runtime
 
-Automatic runtime is a feature avaliable in v7.9.0. With this runtime enabled, the functions that JSX compiles to will be imported automatically.
+Automatic runtime is a feature available in v7.9.0. With this runtime enabled, the functions that JSX compiles to will be imported automatically.
 
 **In**
 
@@ -158,7 +158,7 @@ const profile = Preact.h(
 
 ### Fragments
 
-Fragments are a feature available in React 16.2.0+.
+[Fragments](https://reactjs.org/docs/fragments.html) are a feature available in React 16.2.0+.
 
 #### React
 
@@ -186,7 +186,7 @@ const descriptions = items.map(item =>
 );
 ```
 
-#### Custom for Classic Runtime
+#### Customizing with the Classic Runtime
 
 **In**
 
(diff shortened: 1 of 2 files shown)


docs/plugin-transform-react-jsx.md
```diff
@@ -145,11 +247,14 @@ With options:
...
-      "pragma": "Preact.h", // default pragma is React.createElement
-      "pragmaFrag": "Preact.Fragment", // default is React.Fragment
-      "throwIfNamespace": false // defaults to true
-    }]
+    [
+      "@babel/plugin-transform-react-jsx",
+      {
+        "throwIfNamespace": false, // defaults to true
+        "runtime": "automatic", // defaults to classic
+        "importSource": "Preact" // defaults to react
```
nicolo-ribaudo: I think that we should avoid using this example until Preact supports it.

```suggestion
        "importSource": "custom-jsx-library" // defaults to react
```

nicolo-ribaudo:
Can you also update https://github.com/babel/website/blob/master/docs/preset-react.md?

docs/plugin-transform-react-jsx.md
```diff
@@ -6,41 +6,137 @@ sidebar_label: transform-react-jsx
...
+      className: "profile",
+    }),
+    _jsx("h3", {
+      children: [user.firstName, user.lastName].join(" "),
+    }),
+  ],
+});
+```
+
+#### Custom for Automatic Runtime
```
existentialism: Howabout:

```suggestion
#### Customizing the Runtime Import
```

docs/plugin-transform-react-jsx.md
```diff
@@ -6,41 +6,137 @@ sidebar_label: transform-react-jsx
...
+  null,
   React.createElement("img", { src: "avatar.png", className: "profile" }),
   React.createElement("h3", null, [user.firstName, user.lastName].join(" "))
 );
 ```
 
-### Custom
+### React Classic Runtime
+
+If you can't use auto importing, you can default to classic runtime, which is the default behavior for v7 and prior.
```
existentialism: I think dropping the first "default" reads better, thoughts?

```suggestion
If you do not want (or cannot use) auto importing, you can use the classic runtime, which is the default behavior for v7 and prior.
```

docs/plugin-transform-react-jsx.md
```diff
@@ -6,41 +6,137 @@ sidebar_label: transform-react-jsx
...
+```javascript
+const profile = React.createElement(
+  "div",
+  null,
+  React.createElement("img", { src: "avatar.png", className: "profile" }),
+  React.createElement("h3", null, [user.firstName, user.lastName].join(" "))
+);
+```
+
+### Custom for Classic Runtime
```
existentialism: If we like the suggestion above, this can also match it:

```suggestion
### #### Customizing the Classic Runtime Import
```

docs/plugin-transform-react-jsx.md
lunaruan: Oooh this sounds a lot better

PR #2219, opened 2020-03-26
fix(repl): bump prettier to version 2

### LINKED ISSUE #2218: REPL + prettify should not throw on let = 42
## Bug Report

Updates: ( Thanks to @nicolo-ribaudo )

When the `Prettify` setting of REPL is enabled
```js
let = 42;
```
should not throw.

=== Below is original issue report ===

- [ ] I would like to work on a fix!

**Current Behavior**
The transformed result of the following snippet throws syntax error

```
Unexpected token (2:5)
  1 | var _ref = {};
> 2 | let = _ref["let"];
    |     ^
  3 | _ref;
```

**Input Code**
- [REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=BQbwBANgpgLmC-YC8YTwJQG4g&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=script&lineWrap=true&presets=env&prettier=true&targets=&version=7.9.0&externalPlugins=%40babel%2Fplugin-transform-destructuring%407.8.8)

```js
({ let } = {});
```

**Expected behavior/code**
It should be transformed correctly with `let` assigned to `undefined`

**Babel Configuration (babel.config.js, .babelrc, package.json#babel, cli command, .eslintrc)**
- Filename: `babel.config.js`

```js
{
  "plugins": ["@babel/plugin-transform-destructuring"]
}
```

**Environment**: REPL

**Possible Solution**
Check if `id` is an identifier named `let`, if it is, build a variable declaration instead of an assignment expression.

https://github.com/babel/babel/blob/a901bd50b16b76a87e65508d2a94f098f5745c76/packages/babel-plugin-transform-destructuring/src/index.js#L101

JLHwung:
Fixes #2218.

### DIFF
--- a/js/repl/PluginConfig.js
+++ b/js/repl/PluginConfig.js
@@ -79,7 +79,7 @@ const pluginConfigs: Array<PluginConfig> = [
     baseUrl: "https://unpkg.com",
     label: "Prettify",
     package: "prettier",
-    version: "1.13.0",
+    version: "2",
     files: ["standalone.js", "parser-babylon.js"],
   },
 ];


js/repl/PluginConfig.js
```diff
@@ -79,7 +79,7 @@ const pluginConfigs: Array<PluginConfig> = [
     baseUrl: "https://unpkg.com",
     label: "Prettify",
     package: "prettier",
-    version: "1.13.0",
+    version: "2",
```
JLHwung: This will ensure we are always using latest prettier 2. We can always pin it if any non-major release of prettier breaks our REPL.

JounQin: The babel parser filename has been renamed, related #2274.

PR #2222, opened 2020-04-07
Fix warning about corejs in provided config

mdamien:
When running with the config provided in the usage guide there's a warning.

```
WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.

You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:

  npm install --save core-js@2    npm install --save core-js@3
  yarn add core-js@2              yarn add core-js@3
```

This follows the warning recommendation and sets the `corejs` setting

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -35,6 +35,7 @@ The entire process to set this up involves:
              "safari": "11.1",
            },
            "useBuiltIns": "usage",
+           "corejs": "3.0.0",
          }
        ]
      ]
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -35,7 +35,7 @@ The entire process to set this up involves:
              "safari": "11.1",
            },
            "useBuiltIns": "usage",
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -35,6 +35,7 @@ The entire process to set this up involves:
              "safari": "11.1",
            },
            "useBuiltIns": "usage",
+           "corejs": "3.0.0",
```
nicolo-ribaudo, line 38:
```suggestion
           "corejs": "3.6.4",
```

let's use the last version in the example.

mdamien:
updated, thanks @nicolo-ribaudo

PR #2223, opened 2020-04-09
polish: update landing page minirepl config

JLHwung:
Update the landing page mini-repl settings to a modern one.
Added external helpers and loose mod to minimize the code size, since it is mini-repl.

### DIFF
--- a/js/minirepl.js
+++ b/js/minirepl.js
@@ -109,7 +109,10 @@ function compileCode(sourceEditor, targetEditor) {
 
   try {
     transformed = Babel.transform(sourceEditor.getValue(), {
-      presets: ["react", ["env", { targets: "defaults", loose: true }]],
+      presets: [
+        "react",
+        ["env", { targets: "defaults, not ie 11, not ie_mob 11", loose: true }],
+      ],
       plugins: [["external-helpers", { helperVersion: "7.100.0" }]],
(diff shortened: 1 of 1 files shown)


js/minirepl.js
```diff
@@ -109,7 +109,8 @@ function compileCode(sourceEditor, targetEditor) {
 
   try {
     transformed = Babel.transform(sourceEditor.getValue(), {
-      presets: ["es2015", "es2016", "es2017"],
+      presets: ["react", ["env", { targets: "defaults", loose: true }]],
```
hzoo, line 112: I suppose defaults still compiles basically everything? Maybe in the future we want to make targets a simple dropdown even for the landing page 🤔

JLHwung: FYI here is what will be compiled on "defaults": https://github.com/babel/babel/blob/master/packages/babel-preset-env/test/fixtures/debug/browserslists-defaults/stdout.txt

And `"defaults not ie": https://github.com/babel/babel/blob/master/packages/babel-preset-env/test/fixtures/debug/browserslists-defaults-not-ie/stdout.txt

I am good to remove `ie` here, WDYT?

existentialism: Yeah, I'm fine with dropping:

image

PR #2225, opened 2020-04-09
update landing page code example

JLHwung:
As a follow up to #2223 , after we updated the mini repl preset to a more modernized one, it turns out some of our code examples does not get transformed at all.

This PR update code examples to showcase some of the latest ES features that we still need to apply babel transforms.

### DIFF
--- a/js/minirepl.js
+++ b/js/minirepl.js
@@ -3,10 +3,9 @@
 import debounce from "lodash.debounce";
 
 const miniReplExamples = [
-  "[1, 2, 3].map(n => n ** 2);",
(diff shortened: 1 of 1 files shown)


existentialism:
@hzoo I think that'll require more design as we aren't really surfacing what targets we're using in the mini-repl?

hzoo:
> @hzoo I think that'll require more design as we aren't really surfacing what targets we're using in the mini-repl?

Definitely, so far I'm thinking a very simple version with maybe 3 options, compile everything/IE, modern, current browser

PR #2228, opened 2020-04-20
hotfix: use unpkg.com

JLHwung:
cc @xtuc


JLHwung:
@xtuc Yeah we can revert this PR once `bundle.run` is working again. Let's partially bring back REPL.

PR #2234, opened 2020-04-27
Add docs for the `allowArrayLike` options

nicolo-ribaudo:


docs/plugin-transform-destructuring.md
```diff
@@ -101,6 +101,17 @@ var _z = z,
...
 > You can read more about configuring plugin options [here](https://babeljs.io/docs/en/plugins#plugin-options)
 
+### `allowArrayLike`
+`boolean`, defaults to `false`
+
+This option allows destructuring array-like objects using the array destructuring syntax.
+
+An array-like object is an object with a `length` property: for example, `{ 0: "a", 1: "b", length: 2 }`. Note that, like real arrays, array-like objects can have "holes": `{ 1: "a", length: 3 }` is equivalent to `[ (hole), "a", (hole) ]`.
+
+While is not spec-compliant to destructure array-like objects as if they were arrays, there are many objects that would be _iterables_ in modern browsers with `Symbol.iterator` support. Some notable examples are the DOM collections, like `document.querySelectorAll("img.big")`. Those objects are the main use case for this option.
```
existentialism, line 111:
```suggestion
While it is _not_ spec-compliant to destructure array-like objects as if they were arrays, there are many objects that would be _iterables_ in modern browsers with `Symbol.iterator` support. Some notable examples are the DOM collections, like `document.querySelectorAll("img.big")`, which are the main use case for this option.
```

PR #2246, opened 2020-05-12
Fix package name for plugin-proposal-class-props

paularmstrong:
**Problem:** The name in the docs referencing `@babel/plugin-proposal-class-properties` was listed as `@babel/proposal-class-properties`, which doesn't exist.

**Solution:** Fix the name.


existentialism:
@paularmstrong thanks!

PR #2250, opened 2020-05-14
Updated plugin acceptance docs for include/exclude options in preset-env

JMarkoski:
The docs about include ([https://babeljs.io/docs/en/babel-preset-env#include](https://babeljs.io/docs/en/babel-preset-env#include) say that plugins prefixed with `plugin-` are valid, which is false. The only valid plugin names are those prefixed with `@babel/`, `@babel/plugin-`, `babel-`, `babel-plugin-`. The problem is with the regex normalizing polyfill names here: [https://github.com/babel/babel/blob/9a52019019527be8c179ea7c9025fa2355f16170/packages/babel-preset-env/src/normalize-options.js#L104](https://github.com/babel/babel/blob/9a52019019527be8c179ea7c9025fa2355f16170/packages/babel-preset-env/src/normalize-options.js#L104). Also in versions `6.23.3` and `7.0.0` only `babel-plugin-` prefix is accepted. 
The PR updates every doc version of `preset-env` about which plugin prefixes are accepted in the `include` and `exclude` options.

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -205,7 +205,7 @@ An array of plugins to always include.
 
 Valid options include any:
 
-- [Babel plugins](https://github.com/babel/babel/blob/master/packages/babel-compat-data/scripts/data/plugin-features.js) - with `@babel/` (`@babel/transform-spread`), with `@babel/plugin-` (`@babel/plugin-transform-spread`), with `babel-` (`babel-transform-spread`), with `babel-plugin-` (`babel-plugin-transform-spread`) and without prefix (`transform-spread`) are supported.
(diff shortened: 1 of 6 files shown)


existentialism:
Find it hard to parse:

image


Maybe work better as nested list items?

JMarkoski:
I agree it's somewhat hard to read, but I'm not sure what you are referring to how I should modify it, may you please be more specific? Thanks.

JMarkoski:
@existentialism What do you think about the new version?

existentialism:
My initial thought was something like:

```
Valid options include any:

* Babel plugins - both full and shorthand names are supported, for example the following are functionally equivalent:
  * @babel/plugin-transform-spread
  * @babel/transform-spread
  * babel-transform-spread
  * transform-spread
```

Thoughts?

JMarkoski:
Yes, that's way better and more readable, I agree. Will change it to that format.

PR #2257, opened 2020-05-23
Fix pipeline proposer picker

xixixao:
If you ever select a different pipeline proposal than minimal, you cannot select minimal again, cause the option value prop was misspelled.

May all your bug fixes be this easy.


nicolo-ribaudo:
Thanks!

PR #2274, opened 2020-06-09
fix: prettier babel parser filename

JounQin:
`parser-babylon.js` is deprecated in favor of `parser-babel.js`

### DIFF
--- a/js/repl/PluginConfig.js
+++ b/js/repl/PluginConfig.js
@@ -80,7 +80,7 @@ const pluginConfigs: Array<PluginConfig> = [
     label: "Prettify",
     package: "prettier",
     version: "2",
-    files: ["standalone.js", "parser-babylon.js"],
+    files: ["standalone.js", "parser-babel.js"],
   },
 ];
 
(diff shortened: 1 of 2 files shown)


JLHwung:
Error is thrown on https://deploy-preview-2274--babel.netlify.app/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=O4o&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2%2Cenv&prettier=true&targets=&version=7.10.4&externalPlugins=

```
Couldn't resolve parser "babylon". Parsers must be explicitly added to the standalone bundle.
```

@JounQin Can you take a look?

JounQin:
@JLHwung Sorry I just noticed that error, should have been fixed, please help to review again.

JounQin:
@JLHwung Any time to review?

PR #2280, opened 2020-06-14
Update presets

zjffun:
> As of Babel v6, all the yearly presets have been deprecated. We recommend using @babel/preset-env instead.


existentialism:
@zjffun thanks!

PR #2286, opened 2020-07-05
Fix typo in docs/plugin-proposal-class-properties

leesiongchan:


docs/plugin-proposal-class-properties.md
```diff
@@ -100,7 +100,7 @@ For an explanation of the consequences of using either, see [Definition vs. Assi
   }
 ```
 
-Without `{ "loose": true }`, the above code will compile to the following, using `Object.defineProperty`:
+Without `{ "loose": false }`, the above code will compile to the following, using `Object.defineProperty`:
```
existentialism, line 103: Thanks for the PR, but this is actually correct... if loose is false, it will use `Object.defineProperty`. The "without" might be confusing, maybe this is clearer as:

> When `loose` is set to `false`, the above code will compile to the following:

And subsequently,

> When `loose` is set to `true`, the above code will compile to the following:

Thoughts?

leesiongchan: Ohhhhh! I am so sorry!!! Ok I think this is better

existentialism: No need to apologize!

Would you like to update this PR with those changes?

leesiongchan: Yea I will do that

trusktr, line 103:
```suggestion
When `loose` is set to `false`, the above code will compile to the following, using `Object.defineProperty`:
```

liuxingbaoyu (approved):

PR #2288, opened 2020-07-06
docs: link getting started page on flow/typescript preset

### LINKED ISSUE #2287: Plugins need to be a lot more information in the docs
**Is your feature request related to a problem?**

I decided to try out Rollup in my project. It currently uses Webpack. Upon attempting to get it to work, I come across the following error:

`Add @babel/preset-flow (https://git.io/JfeDn) to the 'presets' section of your Babel config to enable transformation.`

I don't understand why it's asking me to install preset-flow or what it is since I've never had to use it before, so I take a look at the docs:

https://babeljs.io/docs/en/babel-preset-flow

It has a few examples but literally no explanation as to what it actually does. The examples are minimal. I get a basic understanding from them, but if someone is less experienced than myself, then they're going to struggle.

This isn't the first time I've felt this way, and I'm sure I'm not the only one who thinks the plugins need a lot more information, starting with a basic explanation of "what does this do", because the examples simply aren't enough.

**Describe the solution you'd like**

- MVP of what does this do
- Ideally more examples

**Describe alternatives you've considered**.

Googling for answers. Nothing comes up.

JLHwung:
We linked to React Get Started Page: https://babeljs.io/docs/en/babel-preset-react#installation

I think it makes sense to extends to TypeScript/Flow.

Fixes #2287

### DIFF
--- a/docs/preset-flow.md
+++ b/docs/preset-flow.md
@@ -4,7 +4,7 @@ title: @babel/preset-flow
 sidebar_label: flow
 ---
 
-This preset includes the following plugins:
+This preset is recommended if you use [Flow](https://flow.org/en/docs/getting-started/), a static type checker for JavaScript code. It includes the following plugins:
 
 - [@babel/plugin-transform-flow-strip-types](plugin-transform-flow-strip-types.md)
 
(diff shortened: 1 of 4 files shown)


existentialism:
It might also be helpful to just add a short description atop the page like:

> This preset is recommended if you use [Flow](), a static type checker for JavaScript code.

PR #2300, opened 2020-07-22
Move me to contributors, add JLHwung, order core by name

danez:
Well as you noticed I am not really active in babel, I'm struggling with live and the pandemic is not helping with that. But what I wanted to say: I don't feel it's fair to be represented as core member anymore. I'm happy to help in the future and I'm always reachable, but I'm already so far away from the project and it feels that will not change any time soon. 

I also noticed that Huáng is not represented on the team page, but looking at the babel commits I only see his face :joy: So I guess you should be added.

### DIFF
--- a/website/data/team.yml
+++ b/website/data/team.yml
@@ -2,12 +2,15 @@ core:
   - name: Babel
     github: babel
     twitter: babeljs
-  - name: Daniel Tschinder
-    github: danez
-    twitter: TschinderDaniel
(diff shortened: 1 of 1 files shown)


JLHwung:
@danez Thanks for catching that!

existentialism:
> I'm struggling with live and the pandemic is not helping with that

@danez definitely feeling the same, if you ever need anything don't hesitate to reach any of us!

PR #2308, opened 2020-07-30
Create shorter url for blog posts, redirect old

hzoo:


_redirects
```diff
@@ -75,3 +75,18 @@ https://babel.netlify.com/* https://babeljs.io/:splat 301!
 /docs/plugins/undeclared-variables-check/ /docs/en/babel-plugin-undeclared-variables-check
 
 /docs/en/next/tools/*               /setup
+
+/7.11.0 /blog/2020/07/30/7.11.0
```
hzoo, line 79: Will have to do this manually later but not a big deal

_redirects
```diff
@@ -75,3 +75,18 @@ https://babel.netlify.com/* https://babeljs.io/:splat 301!
...
+/7.7.0 /blog/2019/11/05/7.7.0
+/7.6.0 /blog/2019/09/05/7.6.0
+/7.5.0 /blog/2019/07/03/7.5.0
+/7.4.0 /blog/2019/03/19/7.4.0
+/7.3.0 /blog/2019/01/21/7.3.0
+/7.2.0 /blog/2018/12/03/7.2.0
+/7.1.0 /blog/2018/09/17/7.1.0
+/7.0.0 /blog/2018/08/27/7.0.0
+/6.23.0 /blog/2017/02/13/6.23.0
+/6.23.3 /blog/2017/02/13/6.23.0
```
JLHwung, line 92: Are we gonna doing the same for `7.10.5`, etc.?

hzoo: I guess we could, I only did this because of the versions page though.

hzoo: there is something called splats which you could try https://docs.netlify.com/routing/redirects/redirect-options/#splats

`/news/*  /blog/:splat`

package.json
```diff
@@ -56,7 +55,6 @@
     "eslint-plugin-markdown": "^1.0.0-beta.3",
     "eslint-plugin-prettier": "^3.0.1",
     "eslint-plugin-react": "7.13.0",
-    "flow-bin": "^0.98.1",
```
JLHwung, line 59: I don't think we can remove flow check here. We can move flow check to CI though it should be done in another PR.

PR #2310, opened 2020-08-04
Update plugin-transform-property-literals demo

huhaixiao:
Previous demo is not working, because foo keys are not reserved words


JLHwung (approved):
Thanks.

PR #2321, opened 2020-08-15
Update index.js

hzoo:
Removes partial list of sponsors, we should consider just not showing any of them too, not necessary


hzoo:
Just a quick change, basically can't rely on doing this automatically anymore since we'd have to manually check everything

hzoo:
Hopefully builds, never used phone to make a change, at least it's just removal so should be good

PR #2368, opened 2020-10-01
Hacktoberfest

nicolo-ribaudo:
Hopefully this is enough? If we say that we won't accept their PRs hopefully the spam will end.


felquis (approved):
Hope this message helps new comers. In general, I believe the https://www.firsttimersonly.com/ approach is a good one for long-term, better drive interested people to the right spot.
 
as

xtuc (approved):

JLHwung (approved):

PR #2377, opened 2020-10-01
IMPROVEDDOCs

shashimaurya1:
hey i just did some sort of  change which indeed not needed mandatory. hope so u accept my pull request as a beginner for now. upvote


nicolo-ribaudo:
@shashimaurya1 If this is just for the t-shirt, there is a better way to do it: just create a repository on your own profile, and open PRs to that repository. That is not ethic, but at least you don't open these PRs that don't add any value to the project.

PR #2381, opened 2020-10-02
Mention mobileToDesktop in preset-env docs

eps1lon:
To avoid potential confusion like https://github.com/babel/babel/issues/12123


JLHwung (approved):
Thanks.

PR #2382, opened 2020-10-02
Update package.json

adarsh2508:


nicolo-ribaudo:
It's already a few lines above.

PR #2388, opened 2020-10-06
Fixed broken links in docs

acagastya:
There are some broken links of the pattern `/docs/FOO` which don't work and give 404.  The should be instead linked to `/docs/en/FOO`.


acagastya:
I believe there might me even more such broken links, but I haven't looked.

JLHwung:
The language `en` should not be visible to markdown links (Since there is not an `en` folder` in `/docs`). Instead we can add more redirects in https://github.com/babel/website/blob/master/_redirects 
```
/docs/usage /docs/en/usage
/docs/presets /docs/en/presets
```

acagastya:
@JLHwung Okay.  I have submitted #2392 for that, and I am closing this now.

PR #2399, opened 2020-10-14
Clarify typescript preset/plugin scope

### LINKED ISSUE #2311: export * as ns syntax with @babel/preset-typescript
## Bug Report
**Current behavior**
TypeScript 3.8 shipped with support for [`export * as ns` Syntax](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8/#export-star-as-namespace-syntax) which should correctly transpile the following:

**Input Code**
``` ts
export * as api from './api'
```

**Expected behavior**
Using the code above with just [`@babel/preset-typescript`](https://github.com/babel/babel/tree/main/packages/babel-preset-typescript) will throw an error.

In the meanwhile I have been using the [`@babel/plugin-proposal-export-namespace-from`](https://github.com/babel/babel/tree/main/packages/babel-plugin-proposal-export-namespace-from) plugin.

Ideally the functionality of the [`@babel/plugin-proposal-export-namespace-from`](https://github.com/babel/babel/tree/main/packages/babel-plugin-proposal-export-namespace-from) plugin could be merged into [`@babel/preset-typescript`](https://github.com/babel/babel/tree/main/packages/babel-preset-typescript).

JLHwung:
Constantly we have feature requests like https://github.com/babel/babel/issues/10690 https://github.com/babel/babel/issues/12030 to include staged 3 features that are supported in TypeScript compilers. Since we are not making `preset-typescript` a `preset-whatever-tsc-can-do`, we should clarify this point on the docs.

Resolves https://github.com/babel/babel/issues/12030
Fixes #2311

### DIFF
--- a/docs/plugin-transform-typescript.md
+++ b/docs/plugin-transform-typescript.md
@@ -6,7 +6,9 @@ sidebar_label: Typescript Plugin
 
 > **NOTE**: This plugin is included in `@babel/preset-typescript`
 
-This plugin adds support for the syntax used by the [TypeScript programming language][ts]. However, this plugin does not add the ability to type-check the JavaScript passed to it. For that, you will need to install and set up TypeScript.
+This plugin adds support for the types syntax used by the [TypeScript programming language][ts]. However, this plugin does not add the ability to type-check the JavaScript passed to it. For that, you will need to install and set up TypeScript.
+
+Note that although the TypeScript compiler `tsc` actively supports certain JavaScript proposals such as optional chaining (`?.`), nullish coalescing (`??`) and class properties (`this.#x`), this preset does not include these features because they are not the types syntax available in TypeScript only. We recommend using `preset-env` with `preset-typescript` if you want to transpile these features.
 
 ## Example
 


nicolo-ribaudo:
Note that I wouldn't be opposed to it being `preset-parsing-whatever-tsc-can-parse-but-only-transforming-types`, since it doesn't  give us higher backward compat problems than `shippedProposals`, and it's closer to what people expect.

nicolo-ribaudo (approved):

docs/preset-typescript.md
```diff
@@ -34,10 +34,15 @@ npm install --save-dev @babel/preset-typescript
 
 ### With a configuration file (Recommended)
 
+Note that although TypeScript compiler `tsc` actively supports new JavaScript proposals such as optional chaining (`?.`), nullish coalescing (`??`) and class properties (`this.#x`), this preset does not include these features because they are not
```
existentialism, line 37:
```suggestion
Note that although the TypeScript compiler `tsc` actively supports certain JavaScript proposals such as optional chaining (`?.`), nullish coalescing (`??`) and class properties (`this.#x`), this preset does not include these features because they are not
```

docs/preset-typescript.md
```diff
@@ -34,10 +34,15 @@ npm install --save-dev @babel/preset-typescript
...
+the types syntax available in TypeScript only. We recommend using `preset-env` with `preset-typescript` if you want to transpile these features.
```
existentialism, line 38: The following might be confusing:

> this preset does not include these features because they are not the types syntax available in TypeScript only

Maybe something like:

> this preset does not include these features as it is currently only meant to support the TypeScript types syntax.

Thoughts?

PR #2407, opened 2020-10-25
Fix syntax error which breaks syntax highlighting

joelvh:
The missing quote causes syntax highlighting issues in the docs website.


JLHwung (approved):
Thanks!

PR #2416, opened 2020-11-04
Add docs for top-level targets

nicolo-ribaudo:
These docs are mostly copied from `preset-env`'s docs, except for a few small edits.

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -324,6 +324,115 @@ a set of operations as independent compilation passes.
 Note: This option may be removed in future Babel versions as we add better
 support for defining ordering between plugins.
 
+## Output targets
+
+### `targets`
+
+Type: `string | Array<string> | { [string]: string }`<br />
+Default: `{}`<br />
+Placement: Allowed in Babel's programmatic options, or in config files<br />
+Added in: `v7.13.0`<br />
+
+Describes the environments you support/target for your project.
+
+This can either be a [browserslist-compatible](https://github.com/ai/browserslist) query (with [caveats](#ineffective-browserslist-queries)):
+
+```json
+{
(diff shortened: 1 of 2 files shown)


docs/options.md
```diff
@@ -324,6 +324,115 @@ a set of operations as independent compilation passes.
...
+
+> _Please note_: when specifying both `browsers` and the esmodules target, they will be intersected.
+
+```json
+{
+  "targets": {
+    "esmodules": true
+  }
+}
+```
```
AjayPoshak, line 395: I believe we should warn users who have set both `esmodules` and `browsers` options in their config.

docs/preset-env.md
```diff
@@ -82,7 +82,7 @@ For more information on setting options for a preset, refer to the [preset optio
 
 ### `targets`
 
-`string | Array<string> | { [string]: string }`, defaults to `{}`.
+`string | Array<string> | { [string]: string }`, defaults to the top-level `targets` option if no browserslist-related option is specified in `@babel/preset-env`'s docs, otherwise to `{}`.
```
AjayPoshak, line 85: `nitpick`
```
 `@babel/preset-env`'s config, otherwise to `{}`
```

docs/options.md
nicolo-ribaudo: This is now well supported, since Babel will intersect them

docs/preset-env.md
nicolo-ribaudo:
```suggestion
`string | Array<string> | { [string]: string }`, defaults to the top-level `targets` option if no browserslist-related option is specified in `@babel/preset-env`'s options, otherwise to `{}`.
```

PR #2417, opened 2020-11-04
Fix incorrect import paths in @babel/plugin-transform-react-jsx docs

dcastil:
This caused quite some confusion on my side. 😄


nicolo-ribaudo:
Whops, thanks!

PR #2423, opened 2020-11-24
Add note about specifying minor for corejs property

ewanharris:
When specifying just `3` it can lead to a situation where expected polyfills are missing due to changes released in a core-js minor version, e.g. `Promise.allSettled` was moved to stable, but won't be polyfilled by `corejs: 3` but will by `corejs: 3.2`.

By specifying the corejs version in a `3.minor` fashion it ensures that the right polyfills are imported. This is already stated in the core-js docs, but it should be mirrored in the docs for preset-env

References:

core-js issue: zloirock/core-js/issues/793
core-js docs: https://github.com/zloirock/core-js#babelpreset-env

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -392,7 +392,9 @@ Don't add polyfills automatically per file, and don't transform `import "core-js
 
 Added in: `v7.4.0`
 
-`2`, `3` or `{ version: 2 | 3, proposals: boolean }`, defaults to `2`.
+`2`, `3`, `3.minor`, or `{ version: 2 | 3, proposals: boolean }`, defaults to `2`.
+
+> Note: Specifying the `corejs` property with a minor value e.g. `3.7` will ensure that changes in a new minor core-js version are imported. Not specifying a minor version can lead to expected polyfills being missing.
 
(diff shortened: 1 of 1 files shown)


docs/preset-env.md
```diff
@@ -392,7 +392,9 @@ Don't add polyfills automatically per file, and don't transform `import "core-js
 
 Added in: `v7.4.0`
 
-`2`, `3` or `{ version: 2 | 3, proposals: boolean }`, defaults to `2`.
+`2`, `3`, `3.minor`, or `{ version: 2 | 3, proposals: boolean }`, defaults to `2`.
```
ewanharris, line 395: This probably isn't the best way to document it, but I couldn't figure out a better way

nicolo-ribaudo: Maybe we can just put "`string` or `{ version: string, proposals: boolean }` here, and then list the possible versions more in prose?

JLHwung: We should recommend string, e.g. `"3.10"`, for minor version usage here, otherwise we can't differentiate 3.1 and 3.10.

ewanharris: I'll update it to state `string` or `{ version: string, proposals: boolean }` and then update the note to expand upon that

JLHwung (approved):

ewanharris:
This was ultimately done by #2438 so closing

PR #2425, opened 2020-12-01
Clarify docs about env "useBuiltIns" behavior and common uses.

AndrewSouthpaw:
I think a lot of people wind up at this section of the docs and can get pretty easily overwhelmed by all the details about polyfills, core-js, etc.

My intent was to provide some more language around how the system works, and what people "probably" will want to choose the majority of the time, to guide them in the right direction.

I also switched the order of the detailed sections on `usage` and `entry` so that it reflects the order of the options presented (usage, entry, false).

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -264,18 +264,85 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
 
 `"usage"` | `"entry"` | `false`, defaults to `false`.
 
-This option configures how `@babel/preset-env` handles polyfills.
+This option configures how `@babel/preset-env` handles polyfills and whether optimizations are applied to only include polyfills that are needed.
 
-When either the `usage` or `entry` options are used, `@babel/preset-env` will add direct references to `core-js` modules as bare imports (or requires). This means `core-js` will be resolved relative to the file itself and needs to be accessible.
-
-Since `@babel/polyfill` was deprecated in 7.4.0, we recommend directly adding `core-js` and setting the version via the [`corejs`](#corejs) option.
+When specifying polyfilling behavior, you will want to include `core-js` and probably `regenerator-runtime` dependencies:
 
 ```sh
-npm install core-js@3 --save
+npm install --save core-js regenerator-runtime
+```
 
-# or
+You'll also need to set the core-js version via the [`corejs`](#corejs) option:
 
-npm install core-js@2 --save
+```json
+{
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "corejs": 3
+      }
+    ]
+  ]
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/preset-env.md
```diff
@@ -264,18 +264,85 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
...
+      {
+        "corejs": 3
+      }
+    ]
+  ]
+}
+```
 
-# or
+Generally, `"usage"` meets the needs of most users by only including the polyfills that are needed.
```
existentialism, line 290: This might be a tricky recommendation for users who have dependencies in `node_modules` that rely on certain built-ins being available that aren't run through Babel.

docs/preset-env.md
```diff
@@ -264,18 +264,85 @@ This option is useful for "blacklisting" a transform like `@babel/plugin-transfo
 
 `"usage"` | `"entry"` | `false`, defaults to `false`.
 
-This option configures how `@babel/preset-env` handles polyfills.
+This option configures how `@babel/preset-env` handles polyfills and whether optimizations are applied to only include polyfills that are needed.
```
existentialism, line 267: What about something like:

```suggestion
This option configures `@babel/preset-env` to only include the polyfills needed by your target environment(s). This can either include all polyfills needed to mimic a ES2020 environment (`entry`), or only include polyfills as they are needed/used in each file (`usage`).
```

docs/preset-env.md
AndrewSouthpaw: I hadn't even considered that. Sounds like a nuance worth mentioning, I'll include that...

docs/preset-env.md
AndrewSouthpaw: Riffing on that a bit more, how about this? I dropped mention of ES2020 because that will become outdated eventually.

```suggestion
This option configures `@babel/preset-env` to only include the polyfills needed by your target environment(s) to mimic the latest JS-spec environment. Options are to include polyfills as they are needed/used in each file (`usage`) or include polyfills at the entrypoint to your application (`entry`).
```

Given the nuance you mention, I'm not exactly clear on what are the pros/cons of each? If `usage` wouldn't help people in the case about `node_modules` you mentioned above, is there a reason people wouldn't always just use `entry`? If I understand that aspect more, then I think we could get a better synopsis out to users.

existentialism: There are certainly people who _do_ compile their `node_modules` (or if you're sure that your deps don't need to be polyfilled) to get the full benefit of `usage`, it's just a caveat that people may not be aware of when they're setting up Babel for their project.

AndrewSouthpaw: That totally makes sense. I'm just trying to understand, when you say "the full benefit of `usage`", what is that benefit? Are there benefits of `usage` over `entry`? I think that would be good to showcase.

PR #2434, opened 2020-12-15
fix: increase the limits on recent builds to CircleCI API defaults

JLHwung:
We have added so many CircleCI tasks that the standalone build task is not included in limit=10 jobs.

This PR increases `limit` to CircleCI defaults: https://circleci.com/docs/api/v1/#recent-builds-for-a-single-project

Fixed issue: Can not access Babel REPL main build (https://babel.dev/repl/build/main) even the build is successful.

### DIFF
--- a/js/repl/CircleCI.js
+++ b/js/repl/CircleCI.js
@@ -43,7 +43,7 @@ export async function loadLatestBuildNumberForBranch(
   repo: ?string,
   branch: string,
   jobName: string,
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Oh I was trying to use it a few minutes ago and just assumed "probably it has been broken for a long time" :joy:

nicolo-ribaudo (approved):

PR #2436, opened 2021-01-03
Extensions does not work with a quote delimited string.

ryanpcmcquen:
As I discovered in my own project. When using `.ts` files alongside `.js` files, passing a quote delimited string to `--extensions`, causes Babel to ignore plain `.js` files:

extensions_string_issue

Commands are as follows:
```
"build": "babel src/ -s --extensions .js,.jsx,.ts,.tsx -d lib/",
"build-broken": "babel src/ -s --extensions '.js,.jsx,.ts,.tsx' -d lib/",
```

There may be other areas that need to be updated ... or perhaps this is just a bug?

### DIFF
--- a/docs/preset-typescript.md
+++ b/docs/preset-typescript.md
@@ -8,7 +8,7 @@ This preset is recommended if you use [TypeScript](https://www.typescriptlang.or
 
 - [@babel/plugin-transform-typescript](plugin-transform-typescript.md)
 
-> You will need to specify `--extensions ".ts"` for `@babel/cli` & `@babel/node` cli's to handle `.ts` files.
+> You will need to specify `--extensions .ts` for `@babel/cli` & `@babel/node` cli's to handle `.ts` files.
 
 ## Example
 


ryanpcmcquen:
Here is an exact commit where this can be repro'd:
https://github.com/ryanpcmcquen/inkdrop_wiki_links/commit/1428c56aa21933f3e80a2c90d695c74146439d9a

JLHwung:
I don't think we have to change `--extensions ".ts"` to `--extensions .ts`

`@babel/cli` does not support comma-separated `--extensions` string, if you want to provide variadic options

```
--extensions .ts .js
```
or
```
--extensions .ts --extensions .js
```

See also https://github.com/tj/commander.js/#variadic-option

ryanpcmcquen:
Well, this works:
```
--extensions .js,.jsx,.ts,.tsx
```

This does not:
```
--extensions '.js,.jsx,.ts,.tsx'
```
And this does not:
```
--extensions ".js,.jsx,.ts,.tsx"
```

Despite comma separated strings not being supported, they have propagated well across the internet, even landing in an official Microsoft project promoting Babel with TypeScript:

https://github.com/microsoft/TypeScript-Babel-Starter/blob/d7f039380418fcd4d59e6bd4ab3fd8e405d07196/package.json#L10

```
    "build:js": "babel src --out-dir lib --extensions \".ts,.tsx\" --source-maps inline"
```

They are on Microsoft's site as well:

https://devblogs.microsoft.com/typescript/typescript-and-babel-7/

Other various sites have picked up on this syntax now for it to become somewhat idiomatic ... since the quote delimiting is superfluous, is there any reason to keep it in the documentation?

nicolo-ribaudo:
We'll then need a new tests to make sure that comma separated extensions don't accidentally stop working.

JLHwung:
> Here is an exact commit where this can be repro'd:

Thanks for the reproduction repo. However, I can not reproduce this issue: `yarn build-broken` works for me.

```
yarn run v1.22.10
warning wiki_links@0.2.0: The engine "inkdrop" appears to be invalid.
$ babel src/ -s --extensions '.js,.jsx,.ts,.tsx' -d lib/
Successfully compiled 4 files with Babel (654ms).
✨  Done in 0.95s.
```

> @babel/cli does not support comma-separated --extensions string

I was wrong. Babel does support comma separated `extensions`: https://github.com/babel/babel/blob/d3a7cd5e8d5f99fb0b465a5c1d7209d5e1e6ab03/packages/babel-cli/src/babel/options.ts#L354

PR #2439, opened 2021-01-16
Remove yearly presets from the REPL

nicolo-ribaudo:
They have been deprecated since 7.0.0, and they are not even available anymore on npm (the npm packages are empty).

Since existing repl urls and existing settings saved in localStorage still have those presets, I didn't fully remove the option (to avoid breaking existing URLs). Instead, they are just hidden if they are not already enabled.

I also moved the "targets" option to the top: if we are removing the yearly presets it should have a more prominent role, since it's the main way to choose what to transpile. Also, this will better align with the top-level option of Babel 7.13.0.

Preview: https://deploy-preview-2439--babel.netlify.app/repl
Preview with the es2017 preset enabled: https://deploy-preview-2439--babel.netlify.app/repl#?presets=es2017

### DIFF
--- a/js/repl/ReplOptions.js
+++ b/js/repl/ReplOptions.js
@@ -24,18 +24,20 @@ import type {
 } from "./types";
 
 const PRESET_ORDER = [
-  "es2015",
-  "es2015-loose",
-  "es2016",
-  "es2017",
-  "stage-0",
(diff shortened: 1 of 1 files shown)


hzoo (approved):
!!

PR #2444, opened 2021-02-02
polish: mention that we also support foo for @babel/preset-foo

JLHwung:
When triaging https://github.com/babel/babel/issues/12739, I realize that we didn't put that `foo` will also be resolved to `@babel/preset-foo`.


JLHwung:
Nvm, just realize that only works for `@babel/helper-fixtures` 🤦‍♂️

PR #2446, opened 2021-02-06
Add og:image

38elements:
Since `og:image` is not setted, logo is not displayed at Twitter card.
This change adds `og:image`. Since Twitter card does not support SVG, this change adds png image.
https://ogp.me/

https://docusaurus.io/docs/en/site-config.html#ogimage-string


JLHwung (approved):
Thanks!

PR #2448, opened 2021-02-16
Update index.md

AnYiLin:
ES6++

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -13,7 +13,7 @@ Babel is a toolchain that is mainly used to convert ECMAScript 2015+ code into a
 
 ```js
 // Babel Input: ES2015 arrow function
-[1, 2, 3].map((n) => n + 1);
+[1, 2, 3].map(n => n + 1);
 
 // Babel Output: ES5 equivalent
 [1, 2, 3].map(function(n) {


JLHwung:
Can you clarify the motivation of this PR? `(n) => n + 1` and `n => n + 1` is equivalent.

AnYiLin:
The latter can better reflect the characteristics of es6, I think.

JLHwung:
We use [prettier](https://prettier.io/docs/en/options.html#arrow-function-parentheses) to enforce a consistent code style among docs. The option `arrowParens` instructs `prettier` whether an arrow paren should be avoided if possible. The default value was changed from `avoid` (`n => n + 1`) to `always` (`(n) => n + 1`) in v2.0.0. Here is the rationale:

> At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option.

Although I am not going to accept this PR, thanks for contributing to Babel docs anyway!

AnYiLin:
> We use [prettier](https://prettier.io/docs/en/options.html#arrow-function-parentheses) to enforce a consistent code style among docs. The option `arrowParens` instructs `prettier` whether an arrow paren should be avoided if possible. The default value was changed from `avoid` (`n => n + 1`) to `always` (`(n) => n + 1`) in v2.0.0. Here is the rationale:
> 
> > At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option.
> 
> Although I am not going to accept this PR, thanks for contributing to Babel docs anyway!

I got it. Thanks.

PR #2453, opened 2021-02-21
Clarify options to polyfill proposals

cdoublev:
Please forgive me if I'm misunderstanding the current version (english is not my native language), but it let me thinks that I have to choose one of the three described options to polyfill stable features.


nicolo-ribaudo:
The clarification is correct, thanks!

PR #2459, opened 2021-02-25
docs: @babel/node does not support ignore/only in config file

JLHwung:
Context: 
- https://github.com/babel/babel/issues/12038
- https://github.com/babel/babel/issues/11892

Since programmatic options have higher priority than config file, the `ignore` and `only` in config file will never be supported because `@babel/register` specifies these options:

https://github.com/babel/babel/blob/039eb2710d54d03f02240106808cd9f0ef1732d8/packages/babel-register/src/node.js#L144-L164

Before `@babel/core` can offer programmatic options as defaults, state it clearly that they are currently not supported.


nicolo-ribaudo (approved):
One day I'd like to fix this, but it's low priority and requires a lot of effort.

docs/node.md
```diff
@@ -76,14 +77,14 @@ npx babel-node --inspect --presets @babel/preset-env -- script.js --inspect
...
+| `-x, --extensions`     | `".js",".jsx",".es6",".es","cjs"` | List of extensions to hook into                                                                                                                                                                |
+| `--presets`            | `[]`                              | Comma-separated list of [presets](presets.md) (a set of plugins) to load and use.                                                                                                              |
+| `--plugins`            | `[]`                              | Comma-separated list of [plugins](plugins.md) to load and use.                                                                                                                                 |
+| `--config-file [path]` | `[]`                              | Path to the babel config file to use. Defaults to working directory `babel.config.json` or `babel.config.js`. The following options in config file are not supported: `ignore`, `only`.        |
```
existentialism, line 89:
```suggestion
| `--config-file [path]` | `[]`                              | Path to the babel config file to use. Defaults to working directory `babel.config.json` or `babel.config.js`. The following options in config files are not supported: `ignore`, `only`.        |
```

existentialism (approved):

PR #2464, opened 2021-03-03
Fix REPL builtins option

### LINKED ISSUE #2460: builtIns option is not loaded from query parameters
REPL does not use `builtIns` query parameter when provided, that causes some of repl link to not be correct.
This parameter is set by repl itself

https://babeljs.io/repl#?browsers=defaults&build=&builtIns=entry&spec=false&loose=false&code_lz=PTAuFMGdSA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.13.7&externalPlugins=

Currently buildins value is always read from defaults

https://github.com/babel/website/blob/a36a96e914dda58189be5d534a444e5edb0aaae4/js/repl/replUtils.js#L159

image

simmilar issue is present for `corejs` and instead of reading from `envPresetDefaults` we should get those values from `persistedState` 

https://github.com/babel/website/blob/a36a96e914dda58189be5d534a444e5edb0aaae4/js/repl/replUtils.js#L160

JLHwung:
Fixes #2460

### DIFF
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -18,6 +18,7 @@ import TimeTravelSlider from "./TimeTravelSlider";
 import {
   babelConfig,
   envPresetConfig,
+  envPresetDefaults,
   shippedProposalsConfig,
   pluginConfigs,
   runtimePolyfillConfig,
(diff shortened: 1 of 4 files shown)


js/repl/Repl.js
```diff
@@ -567,13 +577,12 @@ class Repl extends React.Component<Props, State> {
 
     const presetsArray = this._presetsToArray();
 
-    const builtIns = envConfig.isBuiltInsEnabled && envConfig.builtIns;
```
JLHwung, line 570: The are moved to `_onOptionChange` so the `envPresetDefaults.builtIns.default` will be also sync to UI.

hzoo (approved):

PR #2467, opened 2021-03-16
docs: allowNamespaces now defaults to true

fytriht:
`allowNamespaces` now defaults to true

### DIFF
--- a/docs/plugin-transform-typescript.md
+++ b/docs/plugin-transform-typescript.md
@@ -71,7 +71,7 @@ class A {
 
 ### `allowNamespaces`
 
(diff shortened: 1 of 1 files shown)


JLHwung:
We should add a history table that it defaults to true since Babel 7.13.0

fytriht:
> We should add a history table that it defaults to true since Babel 7.13.0

It's added https://github.com/babel/website/commit/f73a7ed37682a07c2da3bfaba41279a2d92090fc#diff-630f53ac9773f45ae7d5df8a3a6a74bfb6648dc8598794f9572a6a40d357caa4R81

nicolo-ribaudo (approved):

PR #2471, opened 2021-04-02
update old sponsors

hzoo:
> accidentally did a push (thought I was on a branch so reverted it)

Just updates so it doesn't have our older sponsors since we haven't updated this in a long time. Will make a change after this to actually calculate amounts since not everyone does the monthly amount (either one offs, yearly, periodic donations).

Would also be good to show previous sponsors anyway in a different way (either below or a different view as appreciation), what webpack has is pretty nice.

preview
image


website/pages/en/index.js
```diff
@@ -99,6 +99,9 @@ const SponsorTier = props => {
   );
   return (
     <div>
+      <h3>{props.title}</h3>
+      <div>are currently pledging {props.amount} per month</div>
```
hzoo, line 103: adding a short description

website/pages/en/index.js
```diff
@@ -189,26 +190,14 @@ const Hero = ({ language }) => (
...
+          <a href="https://github.com/babel/babel/releases/tag/v7.13.0">
+            changelog
+          </a>{" "}
+          for more details!
         </span>
       </div>
 
       <MiniRepl language={language} />
-
-      <h3>Special Sponsor</h3>
```
hzoo, line 198: This is gone (could do like a this could be you but don't need to)

website/static/css/index.css
```diff
@@ -141,51 +141,6 @@
   }
 }
 
-.sponsors-work .cards {
```
hzoo, line 144: unused styles

website/siteConfig.js
```diff
@@ -35,7 +35,7 @@ const users = loadYaml("./data/users.yml").map(user => ({
   image: `/img/users/${user.logo}`,
 }));
 
-const sponsorsManual = loadYaml("./data/sponsors.yml").map(sponsor => ({
+const sponsorsManual = (loadYaml("./data/sponsors.yml") || []).map(sponsor => ({
```
nicolo-ribaudo, line 38: Do we still need this `sponsorsManual` logic?

nicolo-ribaudo: Well maybe we might need it again in the future

nicolo-ribaudo (approved):
Thanks!

website/siteConfig.js
hzoo: Yeah the only reason why I kept!

Issue #2179, opened 2020-02-15
Add typescript code-blocks


sidntrivedi012:
In the screenshot below, we can see the input typescript code is marked as javascript in the code block. This may cause confusion to the new contributors. Is there any way to change it? Do we have to manually change the markdown everywhere? 

Screenshot source - https://babeljs.io/docs/en/babel-plugin-transform-typescript

![Screenshot 2020-02-15 at 11 45 07 PM](https://user-images.githubusercontent.com/31186013/74593575-7e04b700-5052-11ea-9d1e-93f944334738.png)

Issue #2244, opened 2020-05-12
Change the shorthands from babel-plugin-… to @babel/plugin-…


dilyanpalauzov:
https://babeljs.io/docs/en/plugins#plugin-shorthand says:
> If the name of the package is prefixed with babel-plugin-, you can use a shorthand: … (skip this prefix)

https://babeljs.io/docs/en/presets#preset-shorthand says:
> If the name of the package is prefixed with babel-preset-, you can use a shorthand: (skip this prefix)

As plugins and presets are commonly not prefixed with `babel-plugin-` or `babel-preset-`, but with `@babel/plugin-` and `@babel/preset-` change the shorthand options to `@babel/preset-` and `@babel/plugin-`.

JLHwung:
I think we can list both `@babel/plugin` and `babel-plugin` as supported prefixes because
1) `@babel/core` supports both of them
2) Conventionally community plugins that are not hosted in `babel` repo start with `babel-plugin-`: e.g. `babel-plugin-istanbul`, `babel-plugin-dynamic-import-node` and a few others to name.

dilyanpalauzov:
> @babel/core supports both of them

I am not sure what this means concerning the current state.

Calling `babel-cli --preset='typescript'` does not work (cannot find module 'babel-preset-typescript'), while `babel-cli --preset='@babel/preset-typescript` does work.

hzoo:
Not sure if behavior changed but testing `babel-cli --presets='typescript'` (preset should be presets) will give the error

> Did you mean "@babel/typescript"?

which is expected since you need the org `@babel` part assuming `@babel/preset-typescript` was installed

But the example may be confusing if you are using Babel's presets vs one on npm so we can change that

PR #2472, opened 2021-04-02
fixup sidebar, modidy parts in plugins/presets

hzoo:
changed sidebar a bit, will do more in a next pr.
- showing the full name of package (removed the sidebar_label, which caused some misc formatting but should be ok)

image

- only updated docs/presets and doc/plugins pages so can ignore the others mostly.
- Fixes https://github.com/babel/website/issues/2179 (just change javascript to typescript code block in 2 places)

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -15,9 +15,6 @@ https://babel.netlify.com/* https://babeljs.io/:splat 301!
 /7/                                 https://old.babeljs.io/repl/build/7.0/
 /7.html                             https://old.babeljs.io/repl/build/7.0/
 
-/docs/en/plugins#pluginpreset-paths /docs/en/plugins#using-a-plugin
-/docs/en/presets#preset-paths /docs/en/presets#using-a-preset
-
 # after docusaurus rewrite
 /php/                               https://old.babeljs.io/php/
 /docs/usage/                        /docs/en/usage/
--- a/_redirects
+++ b/_redirects
@@ -15,6 +15,9 @@ https://babel.netlify.com/* https://babeljs.io/:splat 301!
 /7/                                 https://old.babeljs.io/repl/build/7.0/
 /7.html                             https://old.babeljs.io/repl/build/7.0/
 
+/docs/en/plugins#pluginpreset-paths /docs/en/plugins#using-a-plugin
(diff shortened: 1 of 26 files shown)


_redirects
```diff
@@ -15,6 +15,9 @@ https://babel.netlify.com/* https://babeljs.io/:splat 301!
 /7/                                 https://old.babeljs.io/repl/build/7.0/
 /7.html                             https://old.babeljs.io/repl/build/7.0/
 
+/docs/en/plugins#pluginpreset-paths /docs/en/plugins#using-a-plugin
+/docs/en/presets#preset-paths /docs/en/presets#using-a-preset
```
hzoo, line 19: not sure if this works?

docs/plugins.md
```diff
@@ -97,37 +113,6 @@ These plugins apply transformations to your code.
 - [throw-expressions](plugin-proposal-throw-expressions.md)
 - [private-property-in-object](plugin-proposal-private-property-in-object.md)
 
-### Minification
```
hzoo, line 100: we probably don't need to recommend this

docs/plugins.md
```diff
@@ -169,48 +154,6 @@ Your `.babelrc`:
...
-
-You can also specify an relative/absolute path to your plugin.
-
-```json
-{
-  "plugins": ["./node_modules/asdf/plugin"]
-}
-```
-
-### Plugin Shorthand
```
hzoo, line 190: this is confusing and we recommend using the full path now anyway so can remove? https://github.com/babel/website/issues/2244

Have a link to options.md#name-normalization (it's sorta hidden) but it's there

_redirects
nicolo-ribaudo: I don't think so, the hash is not sent to the server.
We can add an empty div with `id="preset-paths"` next to the `using-a-preset` section.

nicolo-ribaudo (commented):
Thanks!

nicolo-ribaudo (approved):

_redirects
hzoo: Awesome! 
It's a little below the section, but it's not a big deal if we don't either way I guess?

PR #2473, opened 2021-04-03
redo current monthly sponsors calculation, mostly from webpack

hzoo:
> it's not that different from current

The new sponsors download graphql code is via https://github.com/webpack/webpack.js.org/blob/master/src/utilities/fetch-supporters.js. I think it is a more reliable than the old way of using "active" and "tier" because this would do it based on transactions within the last period of time (say a month or year) vs. just because someone happens to sign up at a certain tier (I recall some issues with it earlier but maybe it's fixed), either way haven't looked at any of it in a long time.

I would also like to show the total amounts donated (say for handshake or anyone else who has donated previously but just not currently), could just do it at the bottom, on another page, or make it a toggle. I forgot docu v1 doesn't really have interactivity so we'll need to move out of that.

### DIFF
--- a/scripts/download-sponsors.js
+++ b/scripts/download-sponsors.js
@@ -6,60 +6,189 @@ const fs = require("fs");
 
 const graphqlEndpoint = "https://api.opencollective.com/graphql/v2";
 
-const graphqlQuery = `{
+// all from webpack: https://github.com/webpack/webpack.js.org/blob/master/src/utilities/fetch-supporters.js
+const { promisify } = require("util");
+const asyncWriteFile = promisify(fs.writeFile);
+const REQUIRED_KEYS = ["totalDonations", "slug", "name"];
+const sponsorsFile2 = `${__dirname}/../website/data/sponsors.json`;
+
+// https://github.com/opencollective/opencollective-api/blob/master/server/graphql/v2/query/TransactionsQuery.ts#L81
+const graphqlPageSize = 1000;
+
+const membersGraphqlQuery = `query account($limit: Int, $offset: Int) {
   account(slug: "babel") {
-    orders(status: ACTIVE, limit: 1000) {
-      totalCount
+    members(limit: $limit, offset: $offset) {
       nodes {
-        tier {
-          slug
-        }
-        fromAccount {
+        account {
           name
           slug
           website
           imageUrl
           description
         }
+        totalDonations {
+          value
+        }
+        createdAt
       }
     }
   }
 }`;
 
-const sponsorsFile = `${__dirname}/../website/data/sponsors.json`;
-
-console.log("Downloading sponsors data from Open Collective...");
-
-fetch(graphqlEndpoint, {
-  method: "POST",
-  headers: { "Content-Type": "application/json" },
-  body: JSON.stringify({ query: graphqlQuery }),
-})
-  .then(res => res.json())
-  .then(res => res.data.account.orders.nodes)
-  .then(nodes =>
-    nodes
-      .filter(node => !!node.tier)
-      .map(node => ({
-        tier: node.tier.slug,
-        name: node.fromAccount.name,
-        slug: node.fromAccount.slug,
-        website: node.fromAccount.website,
-        avatar: node.fromAccount.imageUrl,
-        twitterHandle: node.fromAccount.twitterHandle,
-        description: node.fromAccount.description,
-      }))
-  )
-  .then(sponsors => JSON.stringify(sponsors, null, 2))
-  .then(sponsorsJson => {
-    fs.writeFile(sponsorsFile, sponsorsJson, err => {
-      if (err) {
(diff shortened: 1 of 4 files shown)


scripts/download-sponsors.js
```diff
@@ -6,60 +6,189 @@ const fs = require("fs");
...
+const getAllNodes = async (graphqlQuery, getNodes, time = "year") => {
+  const body = {
+    query: graphqlQuery,
+    variables: {
+      limit: graphqlPageSize,
+      offset: 0,
+      dateFrom: new Date(
+        time === "year"
+          ? new Date().setFullYear(new Date().getFullYear() - 1) // data from last year
+          : new Date().setMonth(new Date().getMonth() - 1) // data from last month
```
hzoo, line 78: this is a bit different: I want to query from either the last month or year

website/pages/en/index.js
```diff
@@ -78,29 +78,24 @@ const GetStarted = ({ language }) => {
...
   );
 };
 
 const SponsorTier = props => {
-  const tierSponsors = siteConfig.sponsors.filter(
-    sponsor => sponsor.type == props.type && sponsor.tier === props.tier
-  );
+  let { min, max } = props;
+  const tierSponsors = siteConfig.sponsors.filter(sponsor => {
+    let value = Math.max(sponsor.monthly, sponsor.yearly / 12);
```
hzoo, line 88: Figured this could account for if you are currently donating per month, or if you just started this month, or stopped but still donated an equivalent amount the past year?

nicolo-ribaudo:
> I would also like to show the total amounts donated (say for handshake or anyone else who has donated previously but just not currently), could just do it at the bottom, on another page, or make it a toggle.

What about showing it on hover, with the `title` attribute?

> I forgot docu v1 doesn't really have interactivity so we'll need to move out of that.

We can use normal client-side JS, like we do for the minirepl.

scripts/download-sponsors.js
```diff
@@ -6,60 +6,189 @@ const fs = require("fs");
 
 const graphqlEndpoint = "https://api.opencollective.com/graphql/v2";
 
-const graphqlQuery = `{
+// all from webpack: https://github.com/webpack/webpack.js.org/blob/master/src/utilities/fetch-supporters.js
+const { promisify } = require("util");
+const asyncWriteFile = promisify(fs.writeFile);
```
nicolo-ribaudo, line 11: We can just use `writeFile` from the `fs/promises` module.

scripts/download-sponsors.js
```diff
@@ -6,60 +6,189 @@ const fs = require("fs");
...
+const REQUIRED_KEYS = ["totalDonations", "slug", "name"];
+const sponsorsFile2 = `${__dirname}/../website/data/sponsors.json`;
```
nicolo-ribaudo, line 13: Why `sponsorsFile2` and not `sponsorsFile`?

scripts/download-sponsors.js
```diff
@@ -6,60 +6,189 @@ const fs = require("fs");
...
+      },
+    }).then(response => response.json());
+    const nodes = getNodes(result.data);
+    allNodes = [...allNodes, ...nodes];
+    body.variables.offset += graphqlPageSize;
+    if (nodes.length < graphqlPageSize) {
+      return allNodes;
+    } else {
+      // sleep for a while
+      await new Promise(resolve => setTimeout(resolve, 100));
```
nicolo-ribaudo, line 102: Does it fail without this?

scripts/download-sponsors.js
hzoo: carry over when I was testing with the old one 😛

scripts/download-sponsors.js
hzoo: Oh I didn't change this part, not sure yet

nicolo-ribaudo (approved):

scripts/download-sponsors.js
```diff
@@ -2,64 +2,190 @@
...
+    transactionsGraphqlQuery,
+    data => data.transactions.nodes
+  );
+  for (const transaction of transactionsYear) {
+    if (!transaction.amountInHostCurrency) continue;
+    const amount = transaction.amountInHostCurrency.value;
+    if (!amount || amount <= 0) continue;
+    const supporter = supportersBySlug.get(transaction.fromAccount.slug);
+    if (!supporter) continue;
+    supporter.yearlyDonations += amount;
```
JLHwung, line 161: Not related: I wished decimal proposal were stage 4.

JLHwung (approved):

nicolo-ribaudo (approved):

scripts/download-sponsors.js
nicolo-ribaudo: If it causes problems we can switch it to use `1` = `1 cent` rather than `1$`

PR #2475, opened 2021-04-05
Modify the language type of some codes (in babel-preset-typescript)

dev-itsheng:
When I checked the documentation of [`babel-preset-typescript`](https://babeljs.io/docs/en/babel-preset-typescript), I noticed some problems with the language of some sample codes:

image

After TypeScript code is converted, JavaScript should be generated.

image

The typed code should be TypeScript.

So I try to change it. Next, if I have time, I will check all the code formats in the webpage and try to fix the errors

### DIFF
--- a/docs/preset-typescript.md
+++ b/docs/preset-typescript.md
@@ -19,7 +19,7 @@ const x: number = 0;
 
 **Out**
 
-```typescript
+```javascript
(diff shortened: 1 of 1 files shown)


JLHwung (approved):
Thanks!

PR #2477, opened 2021-04-08
update plugin pages with new sidebar, mark syntax plugins, mention pr…

hzoo:
…eset-env for older transforms

changes a lot of files but mostly to add plugin pages to a sidebar

- Move this section https://babeljs.io/docs/en/plugins#transform-plugins to it's own page

image

- most plugins are a part of a sidebar now (https://babeljs.io/docs/en/babel-plugin-transform-arrow-functions isn't currently)

image

- Check that the plugins that are already a part of preset-env have a note at the top saying so, so people don't add them when they don't need to. Same with react/flow/typescript

image

image

- add syntax only note for syntax plugins that are missing

image


docs/plugin-proposal-async-generator-functions.md
```diff
@@ -64,13 +66,14 @@ function forEach(ai, fn) {
...
+[Try it Out in the REPL](<https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=stage-3&code=async%20function*%20genAnswers()%20%7B%0A%20%20var%20stream%20%3D%20%5B%20Promise.resolve(4)%2C%20Promise.resolve(9)%2C%20Promise.resolve(12)%20%5D%3B%0A%20%20var%20total%20%3D%200%3B%0A%20%20for%20await%20(let%20val%20of%20stream)%20%7B%0A%20%20%20%20total%20%2B%3D%20await%20val%3B%0A%20%20%20%20yield%20total%3B%0A%20%20%7D%0A%7D%0A%0Afunction%20forEach(ai%2C%20fn)%20%7B%0A%20%20return%20ai.next().then(function%20(r)%20%7B%0A%20%20%20%20if%20(!r.done)%20%7B%0A%20%20%20%20%20%20fn(r)%3B%0A%20%20%20%20%20%20return%20forEach(ai%2C%20fn)%3B%0A%20%20%20%20%7D%0A%20%20%7D)%3B%0A%7D%0A%0Avar%20output%20%3D%200%3B%0AforEach(genAnswers()%2C%20function(val)%20%7B%20output%20%2B%3D%20val.value%20%7D)%0A.then(function%20()%20%7B%0A%20%20console.log(output)%3B%20%2F%2F%2042%0A%7D)%3B&experimental=true&loose=false&spec=false&playground=true&stage=0>)
```
JLHwung, line 76: The new link does not work because html entities like `%2C` would not get unescaped when wrapped in `<>`.

I have copied the code and generated a new link.
```
https://babel.dev/repl#?browsers=ie%2011&build=&builtIns=usage&spec=false&loose=false&code_lz=IYZwngdgxgBAZgV2gFwJYHsICoYHMCmEAghCAO74BOIAFAJQwDeAUDDAG7CUwjKX7AAtjAC8MANowACpXSDUIfADp-IdABt2-GgBY6AGmmz5ilfjWbtATgNG5C5ao1aaARgBMDALoBuVhy4YZHRkYHVRGAAGPzY4dG5gMmBUZBgadXxUznD0OB4-AUEGFjY2YNDwgGoxROSssJjSsFR8dQATIJCG_wBfZj7mRBQMCHh4gFFgKAALGmTDOAhi_35kBEpR5KUIfAAPZHolZGnCGiGoNEw0ymXSmFQ8mgBCSiU2zHxbu_gIGhvGu6rdajOKUSYzOaoBZLAEwPpsHp0PwDTjcdAIZAABwxEWigwmU1mBGIpAo1HoCyQFxGNGyxRg6KxOOqAXUSmyCHwcLozCOJ1-50uo3oTH8UEwFmU6nQuBojOxyCRMAA9MqYDp3P0kUA&debug=false&forceAllTransforms=false&shippedProposals=true&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.13.15&externalPlugins=
```

docs/plugin-transform-remove-undefined.md
```diff
@@ -54,12 +54,12 @@ babel --plugins babel-plugin-transform-remove-undefined script.js
...
 require("@babel/core").transformSync("code", {
-  plugins: ["babel-plugin-transform-remove-undefined"]
+  plugins: ["babel-plugin-transform-remove-undefined"],
 });
 ```
 
 ## Options
 
-+ `tdz` - Detect usages before declaration/initialization in let/const(throws) and var(void 0)
+- `tdz` - Detect usages before declaration/initialization in let/const(throws) and var(void 0)
```
nicolo-ribaudo, line 63: It took me too long to realize why the deleted line is marked with `+` in the git diff and the new line with `-` :joy:

docs/plugin-transform-unicode-escapes.md
```diff
@@ -1,21 +1,19 @@
...
 
 **In**
 
 ```javascript
-var \u{1d49c} = "\u{Babe1}";
+var 𝒜 = "\u{Babe1}";
 
-console.log(\u{1d49c});
+console.log(𝒜);
 ```
```
nicolo-ribaudo, line 17: This change should be reverted, to show unicode escapes in the input.

docs/plugin-transform-unicode-escapes.md
```diff
@@ -1,21 +1,19 @@
 ---
 id: babel-plugin-transform-unicode-escapes
 title: @babel/plugin-transform-unicode-escapes
-sidebar_label: transform-unicode-escapes
+sidebar_label: unicode-escapes
 ---
 
-Compile ES2015 Unicode escapes to ES5
```
nicolo-ribaudo, line 7: Why was this deleted?

hzoo: Didn't seem that helpful as a description but can add it back, we should add more to all of these maybe

nicolo-ribaudo: Yeah the name and example are probably enough

nicolo-ribaudo (approved):

PR #2479, opened 2021-04-11
Update usage.md

danBamikiya:
Updated usage.md to not recommend the use of [`@babel/polyfill`](https://babeljs.io/docs/en/babel-polyfill), which is deprecated. Solves #2171

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -253,7 +253,7 @@ For example:
   ]
 }
 ```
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo (commented):
Thanks! Maybe we can mention that `regenerator-runtime` is only needed if you are transpiling generators.

danBamikiya:
@nicolo-ribaudo did just that!

JLHwung (approved):
Thanks!

Issue #2240, opened 2020-05-12
Promote core-js/stable instead of @babel/polyfill


dilyanpalauzov:
https://babeljs.io/docs/en/ says “Babel is a JavaScript compiler: Polyfill features that are missing in your target environment (through @babel/polyfill)”

https://babeljs.io/docs/en/usage says “Running these commands to install the packages: npm install --save @babel/polyfill… Polyfill: As of Babel 7.4.0, this package has been deprecated in favor of directly including core-js/stable (to polyfill ECMAScript features) and regenerator-runtime/runtime (needed to use transpiled generator functions)”

https://babeljs.io/docs/en/babel-polyfill repeats “ As of Babel 7.4.0, this package has been deprecated in favor of directly including core-js/stable (to polyfill ECMAScript features) and regenerator-runtime/runtime (needed to use transpiled generator functions)”

• Change https://babeljs.io/docs/en/ to state “… (through core-js/stable)”

• Update the documentation to state first how to poliyfill with core-js and then how for old Babel to use @babel/polyfill.

JLHwung:
@dilyanpalauzov Thanks for the report, yeah we should have maintained consistency everywhere. Can you draft a PR?

dilyanpalauzov:
No, I do not understand how it works.

PR #2480, opened 2021-04-11
Update index.md to promote core-js/stable instead of @babel/polyfill

danBamikiya:
Change https://babeljs.io/docs/en/ to state “… (through core-js/stable)” instead of "... (through @babel/polyfill". Solves #2240


nicolo-ribaudo:
Thanks! Maybe we could say "through a third-party polyfill such as core-js", since it's not the only one?

danBamikiya:
@nicolo-ribaudo just did that!

nicolo-ribaudo (approved):

PR #2484, opened 2021-04-12
add helper-compilation-targets docs

JLHwung:
Drafted a doc on `@babel/helper-compilation-targets`.

Related: https://github.com/babel/babel/pull/13131

[Preview](https://deploy-preview-2484--babel.netlify.app/docs/en/babel-helper-compilation-targets)

### DIFF
--- a/docs/helper-compilation-targets.md
+++ b/docs/helper-compilation-targets.md
@@ -4,7 +4,7 @@ title: @babel/helper-compilation-targets
 sidebar_label: helper-compilation-targets
 ---
 
-`@babel/helper-compilation-targets` is a helper set working with compilation targets and compat tables. It is used by `@babel/preset-env` to determine which plugin should be enabled based on the give [`targets`](options.md#targets) option.
+`@babel/helper-compilation-targets` is a helper package that works with compilation targets (browsers or other environments like node) and compat tables (knowing what version supports a specific syntax). It is used by `@babel/preset-env` to determine which plugin should be enabled based on the [`targets`](options.md#targets) option.
 
 ```javascript
 import {
--- a/docs/helper-compilation-targets.md
+++ b/docs/helper-compilation-targets.md
@@ -1,0 +1,240 @@
+---
+id: babel-helper-compilation-targets
+title: @babel/helper-compilation-targets
+sidebar_label: helper-compilation-targets
+---
+
+`@babel/helper-compilation-targets` is a helper set working with compilation targets and compat tables. It is used by `@babel/preset-env` to determine which plugin should be enabled based on the give [`targets`](options.md#targets) option.
+
+```javascript
+import {
+  filterItems,
+  getTargets,
+  isRequired,
+} from "@babel/helper-compilation-targets";
+```
+
+## filterItems
+
+```typescript
+function filterItems(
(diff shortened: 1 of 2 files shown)


hzoo:
Awesome! Maybe could mention other tools using it or why you would as an integration?

docs/helper-compilation-targets.md
```diff
@@ -0,0 +1,240 @@
...
+
+**Example**
+
+```javascript
+// Return the default compilation targets
+// returns {}
+getTargets();
+```
+
+An empty compilation target is equivalent to [force all transforms](preset-env.md#forceAllTransforms). The default compilation targets will changed to browserlists [`defaults, not IE 11`](https://runkit.com/jlhwung/605cd58b2c44c6001a463717) in Babel 8.
```
JLHwung, line 152:
```suggestion
An empty compilation target is equivalent to [force all transforms](preset-env.md#forceAllTransforms). The default compilation targets will be changed to browserlists query [`defaults, not IE 11`](https://runkit.com/jlhwung/605cd58b2c44c6001a463717) in Babel 8.
```

hzoo:
Not the best place to put it, but I think we can make another sidebar section for various helper packages including this one

https://github.com/babel/website/blob/70e3ceabd41030658581f9c802ab58a74c030927/website/sidebars.json#L25-L35 

Helper Packages or Babel Helper Packages? another one that people use would be the add import one

JLHwung:
> Helper Packages or Babel Helper Packages? another one that people use would be the add import one

I am good for "Helper Packages", since it is shorter.

docs/helper-compilation-targets.md
```diff
@@ -0,0 +1,243 @@
+---
+id: babel-helper-compilation-targets
+title: @babel/helper-compilation-targets
+sidebar_label: helper-compilation-targets
+---
+
+`@babel/helper-compilation-targets` is a helper set working with compilation targets and compat tables. It is used by `@babel/preset-env` to determine which plugin should be enabled based on the give [`targets`](options.md#targets) option.
```
hzoo, line 7:
```suggestion
`@babel/helper-compilation-targets` is a helper package that works with compilation targets (browsers or other environments like node) and compat tables (knowing what version supports a specific syntax). It is used by `@babel/preset-env` to determine which plugin should be enabled based on the [`targets`](options.md#targets) option.
```

docs/helper-compilation-targets.md
```diff
@@ -0,0 +1,243 @@
...
+
+  // A set of plugins that should always be excluded if `excludes` is empty
+  defaultExcludes?: Array<string> | null,
+
+  // A map from transform plugin to syntax plugin for backward compatibility with older `@babel/parser` versions
+  pluginSyntaxMap?: Map<string, string | null>
+): Set<string>; // A set of enabled plugins
+```
+
+Given a compat data table `list` (i.e. `@babel/compat-data`) and [browser targets](preset-env.md#targets) `targets`, return a set of required plugins.
```
hzoo, line 41: Maybe we need a doc for @babel/compat-data later

hzoo (approved):
Great start, love the examples and good approach for us to continue documenting some of these

PR #2490, opened 2021-04-13
Update babel types docs

JLHwung:
Update babel-types docs, now generated from https://github.com/babel/babel/pull/13148.

Several docs issue are now addressed:

- use `tsArrayType` instead of `tSArrayType` as API titles. They are awkward.
- specify clearly that the fields listed below signature is the AST shape. It is not the list of builder parameters. It has caused confusions before: (https://github.com/babel/babel/issues/12319)
- place Alias after AST node shape since it is a metadata of the AST rather than the builder

[Preview](https://deploy-preview-2490--babel.netlify.app/docs/en/babel-types)

For tracking purpose, please submit your feedback to https://github.com/babel/babel/pull/13148. This PR will be merged when https://github.com/babel/babel/pull/13148 is merged.


JLHwung:
Merged since https://github.com/babel/babel/pull/13148 is merged.

PR #2493, opened 2021-04-14
remove helpers docs that don't have info/internal

hzoo:
Related to https://github.com/babel/website/pull/2484#issuecomment-818293552

There's only a few helpers we intend others to use, or even any that people would want to use. Most of these just have a "TODO" anyway, if we want to document we can add it back?

> @babel/helper-compilation-targets, @babel/helper-module-imports, @babel/helper-annotate-as-pure seem like the main ones.. even @babel/helper-plugin-utils is just for internal?


JLHwung:
Can we rename it to e.g. `.helper-hoist-variables.md`? If we delete them and decide to add back later, we still have to either dig out the git history or manually prepare the boilerplates.

hzoo:
Makes sense, I guess I'd rather just not have it documented in the sense (I don't believe) we want to necessarily encourage the use of any of these packages other than for ourselves? If we do document it, we should just add it to the readme then. Like https://www.npmjs.com/package/@babel/helper-explode-class https://www.npmjs.com/package/@babel/helper-bindify-decorators has no dependents/downloads. It's not really something are promoting for anyone to use in their plugin?

JLHwung:
> Like https://www.npmjs.com/package/@babel/helper-explode-class https://www.npmjs.com/package/@babel/helper-bindify-decorators has no dependents/downloads.

Good catch! We can archive those helpers if they don't have dependents. For helpers we are using, I think it is important to at least document the intention of this helper. Not only does it help users but also it will benefit other Babel contributors.

hzoo:
> We can archive those helpers if they don't have dependents. For helpers we are using, I think it is important to at least document the intention of this helper. Not only does it help users but also it will benefit other Babel contributors.

I agree! I guess I more saying that I also want to document everything more (including internal use) but it doesn't need to be on the _website_ necessarily, if it's for contributors. I guess that's more of a result of us switching to keeping all docs on the website vs. the readme a while ago.

PR #2502, opened 2021-04-16
Promote assumptions on loose docs

JLHwung:
In this PR we promote the alternative `assumptions` usage on the previous `loose` mode docs. I hope it can facilitate the transition from the vague `loose` mode to explicit assumptions.

This PR covers all the `loose` mode docs except `babel-preset-stage-*` since they are deprecated.

### DIFF
--- a/docs/plugin-proposal-class-properties.md
+++ b/docs/plugin-proposal-class-properties.md
@@ -85,7 +85,7 @@ require("@babel/core").transformSync("code", {
 
 When `true`, class properties are compiled to use an assignment expression instead of `Object.defineProperty`.
 
-> ⚠️ Consider migrate to top level [`setPublicClassFields` assumptions](/assumptions#setPublicClassFields)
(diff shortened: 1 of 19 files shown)


docs/plugin-proposal-class-properties.md
```diff
@@ -85,6 +85,17 @@ require("@babel/core").transformSync("code", {
 
 When `true`, class properties are compiled to use an assignment expression instead of `Object.defineProperty`.
 
+> ⚠️ Consider migrate to top level [`setPublicClassFields` assumptions](/assumptions#setPublicClassFields)
```
hzoo, line 88:
```suggestion
> ⚠️ Consider migrating to the top level [`setPublicClassFields`](/assumptions#setPublicClassFields) assumption
```

Do we need to put the version here?
Actually we should add that too https://babeljs.io/assumptions and https://babeljs.io/docs/en/options#assumptions? 7.13

hzoo (approved):
good idea!

PR #2510, opened 2021-04-28
docs: Correction for plugin-syntax-decorators

rschristian:
[This was changed a few years back](https://github.com/babel/babel/commit/d79b5eeeffab14f89c5be92888965e57fcab6436#diff-201cce1951d30fd696d97469efdf4451f2f406ac45317d947013b70a72c076d3), but this page was never updated to reflect that the `decoratorsBeforeExport` option no longer defaulted to `false`


nicolo-ribaudo (approved):
Thanks!

JLHwung (approved):

PR #2519, opened 2021-05-04
Add Wormhole to users

feross:
I'm Feross, author and maintainer of WebTorrent, StandardJS, and other open source projects. I recently started a company and we're using Babel! ❤️

Wormhole lets you share files with end-to-end encryption and a link that automatically expires. So you can keep what you share private and make sure your stuff doesn't stay online forever.

I optimized the SVG image using SVGO, so it's ready to go.


xtuc (approved):

JLHwung:
Thanks!

PR #2528, opened 2021-05-12
update shippedProposals docs

### LINKED ISSUE #2526: "NOTE: This plugin is included in @babel/preset-env" is misleading, should state that it's under the shippedProposals flag
### Problem
The "call-to-action" at the beginning of plugins like [babel-plugin-proposal-class-properties](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties) is really misleading, it says that the plugin is already _included_ in the preset, however, it doesn't say that it's under an option flag, which happens to be presented only at the very end of the documentation.

### Solution
Change to "NOTE: This plugin is included in @babel/preset-env under the [shippedProposals](https://babeljs.io/docs/en/babel-preset-env#shippedproposals) flag".

I'm happy to send a PR if you agree that it makes sense.

JLHwung:
Fixes #2526

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -502,7 +502,7 @@ The following are currently supported:
 - [Private field brand checks](https://github.com/tc39/proposal-private-fields-in-in)
 
 **Materialized Features**
-These features were behind `shippedProposals` flag in older Babel versions. They are now generaly available.
+These features were behind `shippedProposals` flag in older Babel versions. They are now generally available.
 
 - [class properties](https://github.com/tc39/proposal-class-fields)
 - [numeric separator](https://github.com/tc39/proposal-numeric-separator)
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -498,7 +498,7 @@ The following are currently supported:
 **Features**
 
 - [Class static block](https://github.com/tc39/proposal-class-static-block)
-- [Import assertions](https://github.com/tc39/proposal-import-assertions)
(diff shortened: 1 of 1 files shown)


docs/preset-env.md
```diff
@@ -496,6 +497,13 @@ The following are currently supported:
 
 **Features**
 
+- [Class static block](https://github.com/tc39/proposal-class-static-block)
+- [Import assertions](https://github.com/tc39/proposal-import-assertions)
+- [Private field brand checks](https://github.com/tc39/proposal-private-fields-in-in)
+
+**Materialized Features**
+These features were behind `shippedProposals` flag in older Babel versions. They are now generaly available.
```
existentialism, line 505:
```suggestion
These features were behind `shippedProposals` flag in older Babel versions. They are now generally available.
```

docs/preset-env.md
```diff
@@ -496,6 +497,13 @@ The following are currently supported:
 
 **Features**
 
+- [Class static block](https://github.com/tc39/proposal-class-static-block)
+- [Import assertions](https://github.com/tc39/proposal-import-assertions)
```
nicolo-ribaudo, line 501: Maybe we should note that it's just parsing

JLHwung, line 501:
```suggestion
- [Import assertions](https://github.com/tc39/proposal-import-assertions) (parsing only)
```

existentialism (approved):

PR #2529, opened 2021-05-13
feat: add version toggle to REPL

JLHwung:
image

This PR adds a version toggle to REPL. The `Latest` refers to the latest published `@babel/standalone` versions, it's the default when no `version` is specified in URL. The other versions are the highest patch in each minor releases.

[Preview](https://deploy-preview-2529--babel.netlify.app/repl)


js/repl/past-versions.json
```diff
@@ -0,0 +1,7 @@
+[
+  "7.13.17",
+  "7.12.18",
+  "7.11.6",
+  "7.10.5",
+  "7.9.6"
+]
```
nicolo-ribaudo, line 7: Can we autogenerate this by analyzing the output of some `npm` command?

js/repl/Repl.js
```diff
@@ -612,6 +613,12 @@ class Repl extends React.Component<Props, State> {
     UriUtils.updateQuery(payload);
   };
 
+  _onVersionChange = (e: Event) => {
+    this.state.babel.version = e.target.value;
```
existentialism, line 617: Any reason we're mutating state directly here instead of `this.setState`?

JLHwung: No, I used to mutate state when I learned React since most of time they works. Bad habits.

package.json
```diff
@@ -22,7 +22,8 @@
     "start:docusaurus": "cd website && yarn start",
     "start:repl": "webpack -d -w",
     "start:codemirror6": "cross-env NODE_ENV=development yarn build:codemirror6",
-    "test": "remark docs README.md --quiet"
+    "test": "remark docs README.md --quiet",
+    "version": "node ./scripts/generate-repl-past-versions.mjs && git add ./js/repl/past-versions.json"
```
JLHwung, line 26: I think we can add this step to the netlify build commands, in case we forget to run when new minor/major is cut (very likely), but we probably won't forget to merge the docs for new features so hopefully the versions will catch up soon.

PR #2537, opened 2021-05-20
Fix Options page saying 'objects object'

### LINKED ISSUE #2535: https://babeljs.io/docs/en/options
https://babeljs.io/docs/en/options has the text:

> Options can be passed to Babel in a variety of ways. When passed directly to Babel, you can just pass the **objects** object.

I assume it should say "**options** object"?

thernstig:
Instead correctly say 'options object'

Fixes #2535

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -15,7 +15,7 @@ sidebar_label: Config Options
 - [Option concepts](#options-concepts)
 
 Options can be passed to Babel in a variety of ways. When passed directly to Babel,
-you can just pass the objects object. When Babel is used via a wrapper, it may also be
+you can just pass the options object. When Babel is used via a wrapper, it may also be
 necessary, or at least more useful, to pass the options via [configuration files](config-files.md).
 
 If passing options via `@babel/cli` you'll need to `kebab-case` the names. i.e.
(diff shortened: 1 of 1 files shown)


docs/options.md
```diff
@@ -377,7 +377,7 @@ We recognize this isn’t ideal and will be revisiting this in Babel v8.
 
 Type: `boolean`
 
-You may also target browsers supporting ES Modules (https://www.ecma-international.org/ecma-262/6.0/#sec-modules). When specifying this option, the browsers field will be ignored. You can use this approach in combination with `<script type="module"></script>` to conditionally serve smaller scripts to users (https://jakearchibald.com/2017/es-modules-in-browsers/#nomodule-for-backwards-compatibility).
+You may also target browsers supporting ES Modules (<https://www.ecma-international.org/ecma-262/6.0/#sec-modules>). When specifying this option, the browsers field will be ignored. You can use this approach in combination with `<script type="module"></script>` to conditionally serve smaller scripts to users (<https://jakearchibald.com/2017/es-modules-in-browsers/#nomodule-for-backwards-compatibility>).
```
thernstig, line 380: These changes were done by markdownlint, not sure if you are ok with them? I do recommend using markdownlint :)

thernstig: Reasons why angle brackets are good: https://github.com/updownpress/markdown-lint/blob/master/rules/034-no-bare-urls.md

JLHwung: I don't have strong opinion, but it will be great if we can consistently format the markdown either ways.

JLHwung (approved):

PR #2539, opened 2021-05-22
Add assumptions to repl UI

### LINKED ISSUE #2532: Add `assumptions` to repl
I would like to work on adding assumptions option to repl.  I have attached a screenshot of what I think should be the design.

Screenshot 2021-05-16 at 10 57 55 AM

AjayPoshak:
fixes #2532 

Preview: https://deploy-preview-2539--babel.netlify.app/repl

### DIFF
--- a/js/repl/Repl.js
+++ b/js/repl/Repl.js
@@ -126,6 +126,8 @@ class Repl extends React.Component<Props, State> {
 
     const presetsOptions = persistedStateToPresetsOptions(persistedState);
     const envConfig = persistedStateToEnvConfig(persistedState);
+    envConfig.assumptions =
+      typeof envConfig.assumptions === "undefined" ? {} : envConfig.assumptions;
     // const isPresetsTabExpanded = !!presets.filter(preset => preset !== "env")
     //   .length;
 
@@ -240,6 +242,7 @@ class Repl extends React.Component<Props, State> {
           }
           showOfficialExternalPlugins={state.showOfficialExternalPlugins}
           loadingExternalPlugins={state.loadingExternalPlugins}
+          onAssumptionsChange={this._onAssumptionsChange}
         />
         <div className={styles.wrapperPanels}>
           <div
@@ -571,6 +574,12 @@ class Repl extends React.Component<Props, State> {
     }, this._pluginsUpdatedSetStateCallback);
   };
 
+  _onAssumptionsChange = (value: string, isChecked: boolean) => {
+    const { assumptions } = this.state.envConfig;
+    assumptions[value] = isChecked;
+    this.setState({ envConfig: { ...this.state.envConfig, assumptions } });
+  };
+
   _persistState = () => {
     const { state } = this;
     const { envConfig, plugins } = state;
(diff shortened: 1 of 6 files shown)


js/repl/ReplOptions.js
```diff
@@ -666,7 +694,40 @@ class ExpandedContainer extends Component<Props, State> {
...
+                  assumptions[option] === "undefined"
+                    ? false
+                    : assumptions[option];
+                return (
+                  <label className={styles.envPresetRow}>
+                    <LinkToDocs
+                      className={`${styles.envPresetLabel} ${
+                        styles.highlightWithoutUppercase
+                      }`}
+                      section="assumptions"
```
JLHwung, line 715: We don't have `#assumptions` section on `preset-env`, since `LinkToDocs` is just an anchor element, we can come up with our own hyper link to `assumptions.html#option`, i.e.

```
docs/en/assumptions#objectrestnosymbols
```

JLHwung (commented):
The assumptions should also be persisted to the URL. For example, when a user clicks the assumption `iterableIsArray` and then copies the REPL URL, one will expect that if another person clicks this link, `iterableIsArray` will be enabled.

js/repl/ReplOptions.js
AjayPoshak: done

AjayPoshak:
> The assumptions should also be persisted to the URL. For example, when a user clicks the assumption `iterableIsArray` and then copies the REPL URL, one will expect that if another person clicks this link, `iterableIsArray` will be enabled.

done

AjayPoshak:
@JLHwung Please review.

nicolo-ribaudo:
I tested the repl, and it looks like when I change an assumption the code is not re-compiled? Ideally they should trigger a recompilation with the new config.

js/repl/Repl.js
```diff
@@ -581,6 +591,7 @@ class Repl extends React.Component<Props, State> {
       browsers: envConfig.browsers,
       bugfixes: envConfig.isBugfixesEnabled,
       build: state.babel.build,
+      assumptions: JSON.stringify(envConfig.assumptions),
```
JLHwung, line 594: We should also add `assumptions` to `URL_KEYS`:

https://github.com/babel/website/blob/70fe2426dac96f524638f18bf1955459d37b8929/js/repl/UriUtils.js#L7

AjayPoshak: yup, I have it there https://github.com/babel/website/pull/2539/files#diff-086d46e0a51a52ba06a867e04c535506c6a37540e5b335d2504b1c4895ed4510R29

AjayPoshak:
> I tested the repl, and it looks like when I change an assumption the code is not re-compiled? Ideally they should trigger a recompilation with the new config.

fixed

JLHwung (approved):
Thanks!

js/repl/ReplOptions.js
```diff
@@ -981,6 +1068,11 @@ const styles = {
     fontWeight: "bold",
     color: colors.inverseForeground,
   }),
+  highlightWithoutUppercase: css({
+    fontSize: "0.75rem",
+    fontWeight: "bold",
+    color: colors.inverseForeground,
+  }),
```
fedeci, line 1075: Q: is there any way to avoid code repetition?

fedeci (approved):

PR #2540, opened 2021-05-25
Update `babel-jest` example

nicolo-ribaudo:
Ref: https://github.com/facebook/jest/issues/11444

@SimenB Btw, `babel-jest` could add `export { createTransformer }` (other than the `default` export) to avoid the need of `.default` when used from CommonJS or native ESM.

### DIFF
--- a/docs/config-files.md
+++ b/docs/config-files.md
@@ -249,7 +249,7 @@ your Jest options via the [transform option](https://jestjs.io/docs/en/configura
 
 so all JS files will be processed with your version of `babel-jest` with the option enabled.
 
-> NOTE: When using `babel-jest` &le; 26, you must omit the `.default` part: `require("babel-jest").createTransformer({ ...`.
+> NOTE: When using `babel-jest` &le; 27, you must omit the `.default` part: `require("babel-jest").createTransformer({ ...`.
 
 #### Others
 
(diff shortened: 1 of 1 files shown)


SimenB:
You can also provide this in Jest's config, that might be better than using `createTransformer` directly?

```json
{
  "transform": [".*js", {"rootMode": "upward"}]
}
```

(don't copy this snippet without verifying, written from memory)

docs/config-files.md
```diff
@@ -249,6 +249,8 @@ your Jest options via the [transform option](https://jestjs.io/docs/en/configura
 
 so all JS files will be processed with your version of `babel-jest` with the option enabled.
 
+> NOTE: When using `babel-jest` &le; 27, you must omit the `.default` part: `require("babel-jest").createTransformer({ ...`.
```
nicolo-ribaudo, line 252:
```suggestion
> NOTE: When using `babel-jest` < 27, you must omit the `.default` part: `require("babel-jest").createTransformer({ ...`.
```

Probably this? `&le;` is `<=`.

existentialism: Howabout `&lt;`?

existentialism (approved):

docs/config-files.md
nicolo-ribaudo:
```suggestion
> NOTE: When using `babel-jest` &lt; 27, you must omit the `.default` part: `require("babel-jest").createTransformer({ ...`.
```

PR #2545, opened 2021-06-07
docs: add `noIncompleteNsImportDetection` assumption

fedeci:
Add docs for changes in babel/babel#13290

### DIFF
--- a/docs/assumptions.md
+++ b/docs/assumptions.md
@@ -185,6 +185,21 @@ let name = user?.name;
 
 </div>
 
+## `noIncompleteNsImportDetection`
+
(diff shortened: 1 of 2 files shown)


JLHwung (approved):
We should also add it to REPL after #2539 is merged.

existentialism (approved):

JLHwung (commented):
Can you also add an assumption section (can refer to here) in https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs ? So the `assumption` gets more visibility when people are reading plugin docs.

PR #2548, opened 2021-06-22
add manual github sponsors

hzoo:
Can do this until we figure out the github api to get sponsors data like for OC (it comes through OC as "github" vs individualized which makes sense)


website/siteConfig.js
```diff
@@ -67,6 +66,7 @@ const sponsors = [
         total: sponsor.totalDonations,
       };
     }),
+  ...sponsorsManual,
```
hzoo, line 69: Need to override since duplicate entries (in this case coinbase previous donated on OC earlier/stopped and it's different)

PR #2555, opened 2021-07-14
docs: add attachComment

JLHwung:
Docs PR of https://github.com/babel/babel/pull/13229

### DIFF
--- a/docs/parser.md
+++ b/docs/parser.md
@@ -33,6 +33,7 @@ mind. When in doubt, use `.parse()`.
   <summary>History</summary>
 | Version | Changes |
 | --- | --- |
+| `v7.15.0` | Added `attachComment` |
 | `v7.7.0` | Added `errorRecovery` |
 | `v7.5.0` | Added `allowUndeclaredExports` |
 | `v7.2.0` | Added `createParenthesizedExpressions` |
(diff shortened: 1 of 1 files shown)


docs/parser.md
```diff
@@ -64,6 +65,8 @@ mind. When in doubt, use `.parse()`.
   to set this option to `true` to prevent the parser from prematurely
   complaining about undeclared exports that will be added later.
 
+- **attachComment**: By default, Babel attaches comments to adjacent AST nodes. When this option is set to `false`, comments are not attached. It can provide up to 30% performance improvement when the input code has _many_ comments. You probably don't need to set this option, `@babel/eslint-parser` will set it for you. It is not recommended to use `attachComment: false` with Babel transform, as it removes all the comments in output code, and renders annotations such as `/* istanbul ignore next */` nonfunctional.
```
hzoo, line 68: > You probably don't need to set this option, `@babel/eslint-parser` will set it for you.

Maybe can just say `@babel/eslint-parser` will set it for you if it's used for linting? I guess we could mention we may add an option for specifying it https://github.com/babel/babel/pull/13229#issuecomment-829422588 too if we wanted

hzoo (approved):

docs/parser.md
JLHwung: > babel/babel#13229 (comment)

I think that falls perfectly to the roadmap section. I am afraid add not-yet-implemented ideas to this section may confuse our users.

PR #2560, opened 2021-08-04
Fix GraphQL query in the download-sponsors script

nicolo-ribaudo:
This was a breaking change in the GraphQL API that caused our website builds to fail - https://github.com/opencollective/opencollective-api/pull/6178


nicolo-ribaudo:
CI green!

PR #2570, opened 2021-09-01
move plugins to stage 4, es2022

hzoo:
Ref https://github.com/babel/babel/pull/13717, similar to https://github.com/babel/website/pull/2477

### DIFF
--- a/docs/plugins-list.md
+++ b/docs/plugins-list.md
@@ -19,13 +19,6 @@ sidebar_label: Plugins List
 - [throw-expressions](plugin-proposal-throw-expressions.md)
 - [record-and-tuple](plugin-proposal-record-and-tuple.md)
 
-#### Syntax Only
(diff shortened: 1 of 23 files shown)


JLHwung (approved):

docs/plugins-list.md
```diff
@@ -20,28 +19,33 @@ sidebar_label: Plugins List
 - [throw-expressions](plugin-proposal-throw-expressions.md)
 - [record-and-tuple](plugin-proposal-record-and-tuple.md)
 
+#### Syntax Only
```
JLHwung, line 22: I think they can be placed at the last section. They are only required if you are using an older `@babel/core`, we may want to say that here, too.

existentialism (approved):

JLHwung (approved):

PR #2576, opened 2021-09-12
feat: add download REPL button

fedeci:


js/repl/Repl.js
```diff
@@ -504,6 +507,54 @@ class Repl extends React.Component<Props, State> {
...
+      babel: baseBabelConfig({
+        sourceType,
+        envConfig,
+        presets,
+        plugins: externalPlugins.map(plugin => plugin.name),
+        sourceMap: runtimePolyfillState.isEnabled,
+      }),
+    };
+
+    packageJson.devDependencies["@babel/core"] = "latest";
```
fedeci, line 542: It is using the latest core version but should use the last one available for the selected minor version. Still needs some logic for this.

PR #2579, opened 2021-09-29
modify @babel/env to @babel/preset-env

shenlu89:
In [Usage Guide](https://babeljs.io/docs/en/usage), I can see the presets named "@babel/env". However, I cannot see this name elsewhere in the document, instead of "@babel/preset-env". So I modify all the "@babel/env" to "@babel/preset-env" in Usage Guide page.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -25,7 +25,7 @@ The entire process to set this up involves:
    {
      "presets": [
        [
-         "@babel/env",
+         "@babel/preset-env",
          {
            "targets": {
              "edge": "17",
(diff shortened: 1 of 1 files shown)


JLHwung (approved):
Babel will expand `@scope/env` to be `@scope/preset-env` so both `@babel/preset-env` and `@babel/env` work: See https://babeljs.io/docs/en/options#name-normalization

However I don't oppose such changes as I personally prefer the expanded name for explicitness.

hzoo (approved):
yeah same as @JLHwung, we have since recommended being explicit, no reason to need the old way (it used to just be "env")

PR #2590, opened 2021-10-30
index: 7.15.0 -> 7.16.0

coderaiser:
New version of `Babel` released 🥳 !


nicolo-ribaudo:
Thanks because it's always you who remembers about this :sweat_smile:

PR #2593, opened 2021-10-30
docs: fix misleading documentation

LucaColonnello:
Fix misleading documentation on function bind.
A missing new line made the first example very confusing.

image

**From**

```js
obj::func;
// is equivalent to:
func.bind(obj)::obj.func;
// is equivalent to:
obj.func.bind(obj);
```

**To**

```js
obj::func;
// is equivalent to:
func.bind(obj)

::obj.func;
// is equivalent to:
obj.func.bind(obj);
```

### DIFF
--- a/docs/plugin-proposal-function-bind.md
+++ b/docs/plugin-proposal-function-bind.md
@@ -9,7 +9,9 @@ sidebar_label: function-bind
 ```js
 obj::func;
 // is equivalent to:
-func.bind(obj)::obj.func;
+func.bind(obj)
+
+::obj.func;
 // is equivalent to:
 obj.func.bind(obj);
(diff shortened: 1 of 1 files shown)


LucaColonnello:
Apologise in advance if this happens to not respect some contribution standards (I'm not 100% sure), but it was a very quick one I spot while reading, and it took me a while to understand, until I Googled the function bind proposal.

(Not sure this proposal is still alive, but still, worth correcting it)

JLHwung (approved):
Thanks!

LucaColonnello:
Hey @JLHwung ,
Can this be merged? I can't merge it myself unfortunately!

PR #2599, opened 2021-12-23
Add iofod to users

qkorbit:
This PR adds iofod to the website.

iofod uses Babel to handle ES6+ syntax and npm modules. With Babel we can enjoy using the new ES syntax in our projects and not worry about compatibility with production code on slightly older browsers!

Screenshot:

image

Logo:

logo-white


JLHwung (approved):
Thanks!

a931528933:
Cool thing

PR #2601, opened 2021-12-29
Document how to deal with `core-js` and `regenerator-runtime` injection in Yarn PnP

wojtekmaj:


merceyz:
This shouldn't be documented as a PnP specific issue because it isn't one, you can run into the same issue with Yarn using `node_modules`, pnpm, and npm. See https://github.com/storybookjs/storybook/issues/11255 where it happened when using npm.

nicolo-ribaudo:
The preferred solution should be https://babeljs.io/docs/en/babel-plugin-transform-runtime#absoluteruntime, since it doesn't require you to use aliases or change the Yarn configuration.
And, as @merceyz mentioned, it should be used whenever you are compiling your dependencies and want them to reference all the same `core-js` package.

wojtekmaj:
> The preferred solution should be https://babeljs.io/docs/en/babel-plugin-transform-runtime#absoluteruntime

This option is not available in @babel/preset-env though. So how users of @babel/preset-env could escape this trap?

nicolo-ribaudo:
Migrating from preset-env to the new https://github.com/babel/babel-polyfills packages, which have the `absoluteImports` option.

wojtekmaj:
> By migrating from preset-env to the new https://github.com/babel/babel-polyfills packages

Would that be a piece of advice you think the docs should mention instead of the one I wrote?

> which have the absoluteImports option (https://github.com/babel/babel-polyfills/blob/main/docs/usage.md#absoluteimports) and are already internally used by preset-env.

...or maybe this option could be exposed in preset-env?

PR #2608, opened 2022-01-17
docs: target.node vs browserslist node query

JLHwung:
Clarify difference between targets.node and a browserslist node query.

Context: https://github.com/babel/babel/issues/14167#issuecomment-1014844151

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -397,7 +397,7 @@ Type: `string | "current" | true`.
 
 If you want to compile against the current node version, you can specify `"node": true` or `"node": "current"`, which would be the same as `"node": process.versions.node`.
 
-Alternatively, you can specify node version in a browserslist query:
+Alternatively, you can specify the node version in a browserslist query:
 
 ```jsonc
 {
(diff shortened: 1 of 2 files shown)


docs/options.md
```diff
@@ -395,6 +397,22 @@ Type: `string | "current" | true`.
...
 
+Alternatively, you can specify node version in a browserslist query:
+
+```jsonc
+{
+  "targets": "node 12" // not recommended
+}
+```
+
+In this case, browserslist will interpret it as `^12.0.0`, which is resolved to
```
nicolo-ribaudo, line 408: I find this confusing, because to me `^12.0.0` means "I want to support every version starting from 12.0.0. I think we should just say "browserslist will resolve it to the _latest_ version available", without mentioning `^12.0.0`.

nicolo-ribaudo (approved):

docs/options.md
```diff
@@ -395,6 +397,21 @@ Type: `string | "current" | true`.
...
 
+Alternatively, you can specify node version in a browserslist query:
+
+```jsonc
+{
+  "targets": "node 12" // not recommended
+}
+```
+
+In this case, browserslist will resolve it to the _latest_ version available in the `node-releases` library. Because Node.js may support new language features in minor releases, a program generated for Node.js 12.22 may throw a syntax error on Node.js 12.0. We recommend you always specify a minor version when using browserslist node queries:
```
existentialism, line 408:
```suggestion
In this case, browserslist will resolve it to the _latest_ version available in the `node-releases` library. Because Node.js may support new language features in minor releases, a program generated for Node.js 12.22 may throw a syntax error on Node.js 12.0. We recommend that you always specify a minor version when using node queries with browserslist:
```

PR #2629, opened 2022-02-17
Remove redundant comma

hannut91:
I found two commas here. So I deleted it

See also
  - https://babeljs.io/docs/en/options#output-targets


JLHwung (approved):
Thanks

PR #2639, opened 2022-04-15
[7.18.0] Update docs for `importInterop` option

NickHeiner:
It looks like the `umd` and `amd` plugins also support `importInterop` in the code, but don't mention it in the docs. Should I just copy/paste the entire `importInterop` section from the CJS docs? It's not obvious to me that that will be accurate. :smile:

https://github.com/babel/babel/pull/14456

### DIFF
--- a/docs/plugin-transform-modules-amd.md
+++ b/docs/plugin-transform-modules-amd.md
@@ -4,6 +4,13 @@ title: @babel/plugin-transform-modules-amd
 sidebar_label: AMD
 ---
 
+<details>
+  <summary>History</summary>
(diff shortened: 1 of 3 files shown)


JLHwung:
> Should I just copy/paste the entire importInterop section from the CJS docs?

Please. Thank you!

nicolo-ribaudo:
Or an alternative we could also only link the `importInterop` docs, rather than copying them, so that they don't risk getting out of sync.

NickHeiner:
Linking seems preferable to me. :smile: But I'll do whatever y'all chose.

PR #2640, opened 2022-05-05
Add `transform-react-pure-annotations` to preset-react.md

inokawa:
https://github.com/babel/babel/blob/main/packages/babel-preset-react/src/index.ts

I couldn't find plugin-transform-react-pure-annotations.md under [website/docs](https://github.com/babel/website/tree/main/docs) directory.
Even so it would be worth to list this plugin to prevent duplicate use.

### DIFF
--- a/docs/preset-react.md
+++ b/docs/preset-react.md
@@ -110,7 +110,7 @@ Though the JSX spec allows this, it is disabled by default since React's JSX doe
 
 `boolean`, defaults to `true`.
 
-Enables `@babel/plugin-transform-react-pure-annotations`.
+Enables `@babel/plugin-transform-react-pure-annotations`. It will mark top-level React method calls as pure for tree shaking.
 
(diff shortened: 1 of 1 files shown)


docs/preset-react.md
```diff
@@ -105,6 +106,12 @@ Toggles whether or not to throw an error if a XML namespaced tag name is used. F
 
 Though the JSX spec allows this, it is disabled by default since React's JSX does not currently have support for it.
 
+#### `pure`
+
+`boolean`, defaults to `true`.
+
+Enables `@babel/plugin-transform-react-pure-annotations`.
```
JLHwung, line 113:
```suggestion
Enables `@babel/plugin-transform-react-pure-annotations`. It will mark top-level React method calls as pure for tree shaking.
```

JLHwung (approved):
Thanks!

PR #2641, opened 2022-05-09
config-files: rephrase entries in list of configs

Simek:
# Why

Currently, the list of possible configs can be hard to understand for some of the users.

# How

To address this, and simplify the information a bit, I have changed the initial filenames with wildcard and listed all possible extensions after that, instead of using `json` as a base, which might lead to the creation of files like `babel.config.json.mjs`.

Same has been done to the `.babelrc` entry, however I have also extracted the no extension possibility to the separate line, since this was the most confusing part of the previous entry.

Additionally, to avoid repetition, I have changed the "Supported file extensions" intro sentence, so it link to the "Configuration File Types" section, instead of listing the possible configuration file options again.

### DIFF
--- a/docs/config-files.md
+++ b/docs/config-files.md
@@ -16,10 +16,11 @@ Babel has two parallel config file formats, which can be used together, or indep
 </details>
 
 - Project-wide configuration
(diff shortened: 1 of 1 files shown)


JLHwung (approved):
Thanks.

PR #2643, opened 2022-05-17
Added online version of autoprefixer. 

rahulkumarsingh73690:
Added online version of Autoprefixer tool (Featured on Mozilla). So users can instantly use it without installing it.


nicolo-ribaudo:
Well, that section is specifically about the `.broewserslistrc` file that you use only with the "offline" version of autoprefixer.

PR #2645, opened 2022-05-19
Update docs for regenerator-runtime helper inclusion

nicolo-ribaudo:


docs/caveats.md
```diff
@@ -6,17 +6,18 @@ id: caveats
 ## Polyfills
 
 In order for certain features to work they require certain polyfills. You can satisfy **all**
-Babel feature requirements by using [@babel/polyfill](polyfill.md).
+Babel feature requirements by using a complete polyfill such as [`core-js-bundle`](https://www.npmjs.com/package/core-js-bundle).
```
zloirock, line 9: I'm not sure that it's a good idea to recommend the bundled version. It's *bundled*, so can't be optimized and (now) contains early-stage polyfills. The recommended way to include a complete polyfill is `core-js/actual`.

nicolo-ribaudo:
```suggestion
Babel feature requirements by using a complete polyfill such as [`core-js/actual`](https://www.npmjs.com/package/core-js) or (if you want to load it using a `<script>` tag) [`core-js-bundle`](https://www.jsdelivr.com/package/npm/core-js-bundle).
```

Good suggestion, thanks! I'm also linking to a CDN for `core-js-bundle` so that people can just copy the link.

nicolo-ribaudo: I'll push a new commit soon :+1:

Issue #2651, opened 2022-06-07
bundle.run 502 bad gateway & Plugins cound not be loaded


zhiqingchen:
### repl cound not add plugin
![image](https://user-images.githubusercontent.com/1876158/172353589-d8642ba5-8ed7-48e1-ab2c-51123d95ce8c.png)

### caused by network problem
![image](https://user-images.githubusercontent.com/1876158/172353866-5a3f6c47-f97b-48cb-b5fe-463dc17bd45d.png)

### bundle.run is down
https://bundle.run/@babel/plugin-transform-arrow-functions@7.17.12

liuxingbaoyu:
This problem has been around for a long time.
I am trying to solve it.

nicolo-ribaudo:
I think @xtuc has admin access to bundle.run

EnjoyChan:
any solve way here ？I have met the same problem

liuxingbaoyu:
https://deploy-preview-2652--babel.netlify.app/

You can use it temporarily, although it will be a little slow the first time.

trusktr:
@liuxingbaoyu Nice work. I am really looking forward to this! It will be an easy way to make bug reproductions (I may have found a bug with stage 3 decorators causing private fields not to be readable by class code on which a decorator is placed).

I tried to load `@babel/plugin-proposal-decorators` in the repl at https://deploy-preview-2652--babel.netlify.app/, which succeeds to load the plugin, but then it shows this error:

```
[BABEL] /repl.jsx: The decorators plugin, when .version is '2018-09' or not specified, requires a 'decoratorsBeforeExport' option, whose value must be a boolean. (While processing: "base$0$inherits")
```

Is there a way to specify plugin options?

If not, maybe there needs to be some convention that each plugin can provide a schema of the options, or something, then repl can load that, and add the UI for it.

liuxingbaoyu:
Unfortunately this doesn't seem to be supported currently.
https://github.com/babel/website/issues/1292

PR #2652, opened 2022-06-10
fix `bundle.run` 502

liuxingbaoyu:
Allow trying from multiple sources, and add `https://packd.liuxingbaoyu.xyz`.

Fixes: #2651


liuxingbaoyu:
I am not very sure...

liuxingbaoyu:
Let's merge it, it's better than now anyway. :)

PR #2657, opened 2022-06-25
Clarify that generator does not guarantee the output format

dimameh:
Fixes https://github.com/babel/babel/issues/13617

Added a note that babel-generator does not guarantee the output format


nicolo-ribaudo (approved):
Thanks!

PR #2660, opened 2022-07-12
Update 2015-01-27-2to3.md

Ashtaq:


nicolo-ribaudo:
The old name of Babel was 6to5.

PR #2665, opened 2022-07-31
Update documentation

vitorelourenco:
Through testing i've determined that only paths worked for me, not regexp


lxmarinkovic (approved):

liuxingbaoyu:
Thanks for doing this, but regex is indeed supported here, maybe you've hit a bug. :)

PR #2670, opened 2022-08-16
Restore changes in example

dracos:
24588d64 changed this documentation so that it said nothing changed.
I don't know if it also changed anything else.


nicolo-ribaudo:
Thank you!

PR #2671, opened 2022-08-28
chore: update some dependencies

liuxingbaoyu:
webpack doesn't work on newer versions of nodejs.


liuxingbaoyu:
This seems to require upgrading the yarn1 version in the netlify environment variable settings. Can you help me?
cc @JLHwung @nicolo-ribaudo

JLHwung (commented):
Does it work if you keep `.yarnrc`? I guess older Yarn 1 versions, like 1.13 used in netlify build images, only supports `.yarnrc`.

liuxingbaoyu:
It doesn't seem possible, due to the old version of node.
The default yarn version has been updated, except that it defaults to the version when the project was created.
I guess we can update them together in environment variables.

https://answers.netlify.com/t/default-yarn-version-is-now-1-17/2297/11

JLHwung:
@liuxingbaoyu I have added `YARN_VERSION=1.22.19` environment variable, can you push a new commit? I can't re-run the netlify checks from GitHub UI.

liuxingbaoyu:
It looks like we also need to update the NODE version as yarn3 dropped NODE10 support.

JLHwung:
https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript

Can you create `.node-version` at the project root. We can stay with `latest` until the build is broken.

liuxingbaoyu:
Yarn is working fine now, there are other errors, I will continue to investigate.

liuxingbaoyu:
https://docs.netlify.com/configure-builds/troubleshooting-tips/#build-cache
I can't find the reason at all, can you try this?

JLHwung:
@liuxingbaoyu The test was passing (but failing on CI) probably because our local filesystems are not case-sensitive. `./node_modules/Buffer` and `./node_modules/buffer` both works.

JLHwung (approved):

liuxingbaoyu:
Amazing, thanks!

PR #2672, opened 2022-09-02
Document decorators 2022-03 version.

nicolo-ribaudo:
Docs for https://github.com/babel/babel/pull/14836


JLHwung (approved):
Can you also update the `History` table?

PR #2681, opened 2022-09-20
Migrating from `flow` to `typescript`

liuxingbaoyu:
Currently `prettier` cannot parse `import type`, which I will fix in the future.


js/cm6.mjs
```diff
@@ -1,4 +1,4 @@
-export { EditorState, basicSetup } from "@codemirror/basic-setup";
-export { EditorView } from "@codemirror/view";
-export { oneDark } from "@codemirror/theme-one-dark";
-export { javascriptLanguage } from "@codemirror/lang-javascript";
+export { EditorState, basicSetup } from "@codemirror/basic-setup";
+export { EditorView } from "@codemirror/view";
+export { oneDark } from "@codemirror/theme-one-dark";
+export { javascriptLanguage } from "@codemirror/lang-javascript";
```
nicolo-ribaudo, line 4: What's the diff in this file? :thinking:

liuxingbaoyu: It seems that their newlines are inexplicably changed, I'll restore them later.😄

nicolo-ribaudo (approved):
bye bye flow

JLHwung (approved):

PR #2682, opened 2022-09-21
adding-Emacs-editor

lxmarinkovic:
Hi @existentialism & @nicolo-ribaudo  , I noticed that on the [editors](https://babeljs.io/docs/en/editors) page, we don't have Emacs, although Vim has its place. So having in mind that Vim and Emacs aren’t going anywhere anytime soon, it would be fair for the Emacs crowd to have their place next to Vim on the [editors](https://babeljs.io/docs/en/editors) page :D

Screen Shot 2022-09-21 at 10 08 19 AM

### DIFF
--- a/docs/editors.md
+++ b/docs/editors.md
@@ -21,6 +21,12 @@ take a look at the [Setup](/setup) guide.
 Install [language-babel](https://atom.io/packages/language-babel) package
 and follow the [instructions](https://github.com/gandm/language-babel#installation).
 
+### Emacs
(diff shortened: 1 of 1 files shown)


liuxingbaoyu (approved):

lxmarinkovic:
Hi @babel , would you be so kind to tell me what you think about the Emacs addition to the editors page, and review this pr and eventually merge it if you share my opinion on Emacs

PR #2683, opened 2022-09-21
doc: Modify `babel-cli` documentation

liuxingbaoyu:
ref: https://github.com/babel/babel/pull/14950

http://www.html5rocks.com/en/tutorials/developertools/sourcemaps/ is broken.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -69,7 +69,7 @@ npx babel script.js --watch --out-file script-compiled.js
 
 ### Compile with Source Maps
 
-> **Note:** Since v7.19.3, if this parameter is not specified, `babel-cli` will follow the [configuration files](https://babeljs.io/docs/en/config-files).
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -69,8 +69,10 @@ npx babel script.js --watch --out-file script-compiled.js
 
 ### Compile with Source Maps
 
+> **Note:** Since v7.19.3, if this parameter is not specified, `babel-cli` will follow the [configuration files](https://babeljs.io/docs/en/config-files).
```
nicolo-ribaudo, line 72:
```suggestion
> **Note:** Since v7.19.3, if this parameter is not specified, `@babel/cli` will follow the [configuration files](https://babeljs.io/docs/en/config-files).
```

liuxingbaoyu: Great find! I used to think this looked a little weird but couldn't find the reason.😂

Issue #2687, opened 2022-09-26
"Built-ins" throws on REPL


fisker:
```text
[BABEL] /repl.js: @babel/preset-env: The 'useBuiltIns' option must be either
    'false' (default) to indicate no polyfill,
    '"entry"' to indicate replacing the entry polyfill, or
    '"usage"' to import only used polyfills per file (While processing: "base$0")
```

[link](https://babeljs.io/repl#?browsers=ie%2011&build=&builtIns=usage&corejs=3.21&spec=false&loose=false&code_lz=GYexH4DoEMBcAoC0BGAlEA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&timeTravel=true&sourceType=module&lineWrap=true&presets=env%2Cstage-3&prettier=true&targets=&version=7.19.2&externalPlugins=&assumptions=%7B%7D)

liuxingbaoyu:
Can you provide a link where it went wrong?

fisker:
Added

liuxingbaoyu:
Thanks for the report, the PR is waiting to be merged, you can temporarily use https://deploy-preview-2688--babel.netlify.app/

fisker:
Thank you!

PR #2688, opened 2022-09-26
fix `useBuiltIns`

liuxingbaoyu:
fixes: #2687

### DIFF
--- a/js/repl/compile.ts
+++ b/js/repl/compile.ts
@@ -55,7 +55,7 @@ export default function compile(code: string, config: CompileConfig): Return {
   let compileErrorMessage = null;
   let envPresetDebugInfo = null;
   let sourceMap = null;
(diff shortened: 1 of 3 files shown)


js/repl/compile.ts
```diff
@@ -82,7 +82,7 @@ export default function compile(code: string, config: CompileConfig): Return {
       targets.electron = envConfig.electron;
     }
     if (envConfig.isBuiltInsEnabled) {
-      useBuiltIns = !config.evaluate && !!envConfig.builtIns;
+      useBuiltIns = (!config.evaluate && envConfig.builtIns) as false|"entry"|"usage";
```
liuxingbaoyu, line 85: It seems that ts cannot be inferred normally here.

JLHwung (approved):

PR #2694, opened 2022-10-20
feat: Support decorators 2022-03

liuxingbaoyu:


js/repl/compile.ts
```diff
@@ -167,6 +168,8 @@ export default function compile(code: string, config: CompileConfig): Return {
         : undefined,
     };
 
+    console.info("Babel config", babelConfig);
```
nicolo-ribaudo, line 171: Missed this

nicolo-ribaudo (approved):

js/repl/compile.ts
liuxingbaoyu: Yeah, that's fine for me. I don't mind removing it though.

nicolo-ribaudo: Sorry, I meant "you forgot to remove this" :sweat_smile: 
Or is it intentional? I often use `console.log` in the repl, and I am worried that this would produce a lot of logs.

liuxingbaoyu: This is intentional, but I don't mind removing it at all.
Also I thought this should be `console.debug`, which I forgot about at first. :)

nicolo-ribaudo: Ok, I would prefer to hide it. Also because the `@babel/standalone` config format does not match the normal config format (for example, plugin/preset names), so people wouldn't be able to copy-paste it anyway.

I'm +1 for showing the generated/normalzied json config, but that should probably happen in a way more integrated with the UI.

js/repl/compile.ts
```diff
@@ -167,6 +168,8 @@ export default function compile(code: string, config: CompileConfig): Return {
...
+
     const transformed = Babel.transform(code, babelConfig);
```
liuxingbaoyu, line 173:
```suggestion
    const transformed = Babel.transform(code, babelConfig);
```

js/repl/compile.ts
liuxingbaoyu: Yes, I want to improve it a bit, but it's a little difficult.

PR #2698, opened 2022-10-28
Fix name of plugin-proposal-duplicate-named-capturing-groups-regex

ptomato:
duplicated → duplicate; this is the name the package is published under.


nicolo-ribaudo (approved):
Thank you!

PR #2701, opened 2022-11-10
fix: Allow default assumptions

liuxingbaoyu:


JLHwung (changes requested):
Personally I don't like the idea of intermediate checkbox, aka tri-state checkbox. The `noNewArrows` assumption is the only one that is enabled by default and we planned to disable it in Babel 8, to align with other assumptions. 

I suggest we render the current default setting on UI, i.e. `noNewArrows` is `true` while the others are `false`.

Issue #2705, opened 2022-12-09
https://babeljs.io/versions is out of date


Bosch-Eli-Black:
https://babeljs.io/versions lists the latest version at `7.11.0`, but the latest version is `7.20.0`

PR #2706, opened 2022-12-09
Update the version list and add exception printing

liuxingbaoyu:
Fixes: #2705


existentialism (approved):

website/past-versions.json
```diff
@@ -1,4 +1,12 @@
 [
+  "7.20.0",
+  "7.19.0",
+  "7.18.0",
+  "7.17.0",
+  "7.16.0",
+  "7.15.0",
+  "7.14.0",
+  "7.13.0",
```
liuxingbaoyu, line 9:
```suggestion
  "7.13.0",
  "7.12.0",
```

JLHwung:
The `past-versions.json` has not been updated because we removed versioned docs since Babel 7.11. But it would be nice to keep the link of the release blog post.

I think we can update https://github.com/babel/website/blob/main/website/pages/en/versions.js. The current page still refers to pre-release document, but now the current documentation is live: it shows the version information in the API history table. For those API without such table, you can safely assume that this API is available since 7.11.

Not related to this PR, but the REPL past versions are also out-of-dated: You can run https://github.com/babel/website/blob/main/package.json#L25 to update.

To summarize, we can
1. update the Past Versions sections of the versions page
2. merge the Current version section with the pre-release version
3. update REPL past-versions

liuxingbaoyu:
I think it's ok to remove the main branch, because we only merge PRs when they are about to be released.

PR #2707, opened 2022-12-13
Parser `function.sent`requires plugin

overlookmotel:
The docs for parser currently includes `function.sent` in the list of syntax "already enabled on the latest version of @babel/parser, and cannot be disabled because they are part of the language".

This does not appear to be correct.

With `@babel/parser` v7.20.5 (latest at time of writing):

```js
const {parse} = require('@babel/parser');
parse('function* f() { return function.sent; }');
```

throws error `SyntaxError: This experimental syntax requires enabling the parser plugin: "functionSent". (1:32)`

This PR revises the docs to reflect that `functionSent` plugin must be manually enabled to parse this syntax.


JLHwung (approved):
Thanks!

PR #2710, opened 2023-01-03
feat: Add reset button

liuxingbaoyu:
Fixes: #2709 
image

### DIFF
--- a/js/repl/ReplOptions.tsx
+++ b/js/repl/ReplOptions.tsx
@@ -23,6 +23,7 @@ import type {
 } from "./types";
 
 import type { ReactElement, ChangeEvent } from "react";
+import StorageService from "./StorageService";
 
 const PRESET_ORDER = [
   "react",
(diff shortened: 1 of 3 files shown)


js/repl/ReplOptions.tsx
```diff
@@ -766,22 +767,36 @@ class ExpandedContainer extends Component<Props, State> {
...
-            </select>
-          </div>
-        )}
+
+        <div className={styles.versionRow} title={`v${babelVersion}`}>
+          <button
+            onClick={() => {
+              StorageService.set("replState", "");
+              location.hash = "";
+              location.reload();
```
JLHwung, line 776: nit: Can you extract the click handler into a component property? The `Repl` component should take care of the reset logic.

js/repl/ReplOptions.tsx
```diff
@@ -766,22 +767,36 @@ class ExpandedContainer extends Component<Props, State> {
...
-                    <option value={version} key={index}>
-                      v{version}
-                    </option>
-                  )
-              )}
-            </select>
-          </div>
-        )}
+
+        <div className={styles.versionRow} title={`v${babelVersion}`}>
```
JLHwung, line 771: nit: Since now it contains a version selector and a reset button, can we rename `versionRow`? Also we may remove the `title` property now.

js/repl/ReplOptions.tsx
```diff
@@ -766,22 +767,36 @@ class ExpandedContainer extends Component<Props, State> {
...
+            onClick={() => {
+              StorageService.set("replState", "");
+              location.hash = "";
+              location.reload();
+            }}
+          >
+            Reset
+          </button>
+          <select
+            className={css({
```
JLHwung, line 782: The `className` should have been inherited from `styles.optionSelect`.

js/repl/ReplOptions.tsx
```diff
@@ -766,22 +767,36 @@ class ExpandedContainer extends Component<Props, State> {
...
+              marginLeft: 10,
```
JLHwung, line 783: Conventionally `rem` is adopted in this project. Can you convert `10px` to `rem`?

js/repl/ReplOptions.tsx
JLHwung: I mean

```suggestion
            className={styles.optionSelect}
```
The margin settings have been specified there.

<details>
<summary>If you are confused by the "should have"</summary>
"should have" is a virtual tense, it means "we wish to do something but for some reason we don't" (本應該 in Chinese). It does not mean "we have already done something" (應該已經 in Chinese).
</details>

liuxingbaoyu: image
image

Thank you for your detailed explanation!
`optionSelect` seems to have unexpected effects.

JLHwung: It is somehow expected because we don't have a button style. Now we probably should draft one. Can you take the box settings from the select widget?

liuxingbaoyu: Sorry I'm not sure if this is expected. (I haven't written css before)

JLHwung (approved):

PR #2713, opened 2023-01-25
fix jsx example syntax error

### LINKED ISSUE #2699: doc index page, jsx example has syntax error
at the babel docs' index page, the section about [jsx and react](https://babeljs.io/docs/en/#jsx-and-react), an example snippet goes like 

```jsx
export default function DiceRoll(){
  const [num, setNum] = useState(getRandomNumber());

  const getRandomNumber = () => {
    return Math.ceil(Math.random() * 6);
  };

  return (
    <div>
      Your dice roll:
      {num}
    </div>
  );
};
```

which is not that good. it has a reference used before initiation (getRandomNumber), and a variable declared but never used (setNum).

a better one might be

```jsx
export default function DiceRoll(){
  const getRandomNumber = () => {
    return Math.ceil(Math.random() * 6);
  };

  const [num, setNum] = useState(getRandomNumber());

  const handleClick = () => {
    const newNum = getRandomNumber();
    setNum(newNum);
  };

  return (
    <div>
      Your dice roll: {num}.
      <button onClick={handleClick}>Click to get a new number</button>
    </div>
  );
};
```

fynnfluegge:
Closes #2699

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -44,16 +44,21 @@ and add `@babel/preset-react` to your Babel configuration.
 
 ```jsx
 export default function DiceRoll(){
-  const [num, setNum] = useState(getRandomNumber());
-
   const getRandomNumber = () => {
     return Math.ceil(Math.random() * 6);
   };
 
+  const [num, setNum] = useState(getRandomNumber());
+
+  const handleClick = () => {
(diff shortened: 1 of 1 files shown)


JLHwung (approved):
Thanks!

PR #2714, opened 2023-01-27
Update preset-env.md

epreston:
Clarify exception to "targets" handling for backward compatibility.

https://github.com/babel/babel/issues/15371#issuecomment-1404849205

Closes https://github.com/babel/babel/issues/15371

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -82,7 +82,7 @@ For more information on setting options for a preset, refer to the [preset optio
 
 ### `targets`
 
-`string | Array<string> | { [string]: string }`, defaults to the top-level `targets` option if no [browserslist config sources](https://github.com/ai/browserslist#queries) are specified, otherwise to `{}`.
+`string | Array<string> | { [string]: string }`, defaults to the top-level [`targets` option](options.md#targets) if [`configPath`](#configpath), [`browserslistEnv`](#browserslistenv) and [`ignoreBrowserslistConfig`](#ignorebrowserslistconfig) are not specified, otherwise to `{}`.
 
 For usage, refer to the [`targets` option](options.md#targets) documentation.
 
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -82,9 +82,11 @@ For more information on setting options for a preset, refer to the [preset optio
 
 ### `targets`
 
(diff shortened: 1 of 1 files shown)


epreston:
This clarifies the behaviour of the `targets` setting in `@babel/preset-env`.  The behaviour of `esmodule` and `browsers` "intersection" differs depending on where the settings come from.

- from the `.browserslistrc` - intersection
- from the top level `targets` setting - intersection
- from the local `targets` setting within the `@babel/preset-env` option - no intersection

Further detail is provided in the github issue: https://github.com/babel/babel/issues/15371

docs/preset-env.md
```diff
@@ -82,10 +82,12 @@ For more information on setting options for a preset, refer to the [preset optio
 
 ### `targets`
 
-`string | Array<string> | { [string]: string }`, defaults to the top-level `targets` option if no browserslist-related option is specified in `@babel/preset-env`'s docs, otherwise to `{}`.
+`string | Array<string> | { [string]: string }`, defaults to the top-level `targets` option if no [browserslist config sources](https://github.com/ai/browserslist#queries) are specified, otherwise to `{}`.
```
JLHwung, line 85: "browserslist-related options" are actually more than the config sources. Namely they are `configPath`, `browserslistEnv` and `ignoreBrowserslistConfig`. So I prefer to keep previous wordings, or we can clarify what are these options.

JLHwung:
```suggestion
`string | Array<string> | { [string]: string }`, defaults to the top-level [`targets` option](options.md#targets) if [`configPath`](#configpath), [`browserslistEnv`](#browserslistenv) and [`ignoreBrowserslistConfig`](#ignorebrowserslistconfig) are not specified, otherwise to `{}`.
```

The browserslist config sources can be found in `options.md#targets`, as we have put below:

```md
> For usage, refer to the [`targets` option](options.md#targets) documentation.
```

For record there was a link to the browserslist query before we are promoting the top level `targets` option, which can benefit not only `preset-env` but also other plugins. After that most of the `targets` docs are moved to `options.md#targets`. 

I think this section should only cover the behaviour different to the top-level `targets`, like the note you have put about the `esmodules` intersection. Users are supposed to refer to `options.md#targets` about how `targets` should be configured. Otherwise we will have almost identical two `targets` configuration guides.

epreston: I trust you understand the subtleties more than me. I've left this PR open to modifications so you can tune this appropriately.

PR #2715, opened 2023-02-06
Add docs for 2023-01 decorators

nicolo-ribaudo:
Docs for https://github.com/babel/babel/pull/15405


liuxingbaoyu:
I will add this in the repl.

By the way, the current repl, the decorator only works when `stage-2` is uniquely enabled.
It's a bit strange that an exception will be thrown if both `stage-2` and `stage-3` are enabled.
We probably need to do something to avoid user confusion.

docs/plugin-proposal-decorators.md
```diff
@@ -84,15 +84,17 @@ require("@babel/core").transformSync("code", {
...
 | --- | --- |
+| `v7.21.0` | Added support for `version: "2023-01"` |
 | `v7.19.0` | Added support for `version: "2022-03"` |
 | `v7.17.0` | Added the `version` option with support for `"2021-12"`, `"2018-09"` and `"legacy"` |
 </details>
 
 ### `version`
 
-`"2022-03"`, `"2021-12"`, `"2018-09"` or `"legacy"`. Defaults to `"2018-09"`.
+`"2023-01"`, `"2022-03"`, `"2021-12"`, `"2018-09"` or `"legacy"`.
```
JLHwung, line 94: Is a `version` a required option now? If so it would be a breaking change.

docs/plugin-proposal-decorators.md
```diff
@@ -118,7 +120,7 @@ export @decorator class Bar {}
 export class Foo {}
 ```
 
-This option was added to help tc39 collect feedback from the community by allowing experimentation with both possible syntaxes. The proposal now settled on having decorators after `export`.
+This option was added to help tc39 collect feedback from the community by allowing experimentation with both possible syntaxes. The proposal now settled on allowing decorators either before or after `export`.
```
existentialism, line 123:
```suggestion
This option was originally added to help tc39 collect feedback from the community by allowing experimentation with the proposed syntaxes. The proposal has now settled on allowing decorators either before or after `export`.
```

docs/plugin-proposal-decorators.md
nicolo-ribaudo: No, just `["@babel/plugin-proposal-decorators"]` has always been invalid. `"version"` has always been required, except if `"decoratorsBeforeExports"` was set. Since most people will not need to set it now, it's confusing to say that `"version"` has a default value (however, this behavior is still mentioned a few lines below).

PR #2734, opened 2023-02-24
fix dark mode primary colors

JLHwung:
The dark mode primary colors seem not working.

Currently main:

image

Expected:

image

### DIFF
--- a/website/src/css/custom.css
+++ b/website/src/css/custom.css
@@ -17,7 +17,7 @@
 
   --ifm-blockquote-border-left-width: 4px;
 }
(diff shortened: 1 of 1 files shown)


JLHwung:
Technically they are not new colors, they were included in #2622 and they are derived directly from the brand color. For some reason they are not enabled in #2622 and this PR fixes that.

Personally I prefer themes in this PR.

liuxingbaoyu (approved):

nicolo-ribaudo (approved):
The new colors have much better contrast, it makes it easier to read them.

PR #2739, opened 2023-03-01
sidebar: move plugins under presets

JLHwung:
This is an experimental PR. I would like learn about your opinions.

Previously the plugins have dedicated sidebars, so there are no means to navigate to the plugin pages unless you already know the plugin name _and_ either 1) search on the docs website or 2) visit the plugin docs link from npm package readme.

In this PR we moved most plugins (except modules / proposals) under the presets page. So users can learn from the sidebar that which plugins are included in which presets and then they can check it out if they want to learn more about the presets.

The "presets" category now displays contents of "config reference / presets" page now that docusaurus 2 supports category link.

Preview link: https://deploy-preview-2739--babel.netlify.app/docs/

### DIFF
--- a/website/sidebars.json
+++ b/website/sidebars.json
@@ -1,50 +1,208 @@
 {
-  "docs": {
-    "Guides": ["index", "usage", "configuration", "learn", "v7-migration"],
-    "Config Reference": [
-      "config-files",
-      "options",
-      "presets",
-      "plugins",
-      "plugins-list",
-      "assumptions"
-    ],
-    "Presets": [
-      "babel-preset-env",
-      "babel-preset-react",
-      "babel-preset-typescript",
-      "babel-preset-flow"
-    ],
-    "Misc": ["roadmap", "caveats", "features-timeline", "faq", "editors"],
-    "Integration Packages": [
-      "babel-cli",
-      "babel-polyfill",
-      "babel-plugin-transform-runtime",
-      "babel-register",
-      "babel-standalone"
-    ],
-    "Tooling Packages": [
-      "babel-parser",
-      "babel-core",
-      "babel-generator",
-      "babel-code-frame",
-      "babel-runtime",
-      "babel-template",
-      "babel-traverse",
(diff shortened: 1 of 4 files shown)


website/sidebars.json
```diff
@@ -1,49 +1,207 @@
...
+      },
+      "items": [
+        {
+          "type": "category",
+          "label": "@babel/preset-env",
+          "link": {
+            "type": "doc",
+            "id": "babel-preset-env"
+          },
+          "items": [
```
JLHwung, line 34: In the future we can add docs of bug fixes plugin here.

liuxingbaoyu:
LGTM.
My personal preference is to move everything from the original.
Include these.
https://deploy-preview-2739--babel.netlify.app/docs/babel-plugin-transform-modules-amd

liuxingbaoyu:
> Previously the plugins have dedicated sidebars, so there are no means to navigate to the plugin pages unless you already know the plugin name _and_ either 1) search on the docs website or 2) visit the plugin docs link from npm package readme.

They are actually accessible through the addons list page, but I think having them in the sidebar would make them significantly easier to access.

JLHwung:
> My personal preference is to move everything from the original.
Include these.
https://deploy-preview-2739--babel.netlify.app/docs/babel-plugin-transform-modules-amd

Yeah we can decide where we should put module transforms and TC39 proposals later. Though `preset-env` technically does include module transforms, I am a bit hesitant about moving them under `preset-env` because they are not part of the spec. They are more like integration plugins with other module formats and may deserve a new top level plugins sidebar item:

```json5
// sidebars.json
{
  "Plugins": [ "Module Transforms", "TC39 proposals" ]
}
```

liuxingbaoyu:
> Though preset-env technically does include module transforms, I am a bit hesitant about moving them under preset-env because they are not part of the spec.

Yes, I also don't think it's necessary to put it under `preset-env`, a new item might be better.

Issue #2735, opened 2023-02-26
Configure redirects from `/en/` to `/`


harish-sethuraman:
Related to https://github.com/reactjs/reactjs.org/issues/5633

Seems like some of the older URLs had `/en/` in the URL. Should we consider redirecting those to `/` to avoid older URLs throwing 404? 

we have `/docs/en/*` for redirects already. Should we handle the repl case separately? 
```
# Docusaurus v1 compat

/docs/en/* /docs/:splat
```

JLHwung:
Yes, we should have redirected `/en/repl` to `/repl`. I didn't know that was even working before.

PR #2742, opened 2023-03-02
chore: add redirect for repl

harish-sethuraman:
This should add redirect to repl? 
fixes https://github.com/babel/website/issues/2735

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -55,8 +55,8 @@ https://babel.netlify.com/* https://babeljs.io/:splat 301!
 /docs/usage/require/                /docs/babel-register
 /docs/usage/options/                /docs/options
 /caveats.html                       /docs/caveats
-/repl.html                          /repl/
-/en/repl                            /repl/
+/repl.html                          /repl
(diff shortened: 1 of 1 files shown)


_redirects
```diff
@@ -56,6 +56,7 @@ https://babel.netlify.com/* https://babeljs.io/:splat 301!
 /docs/usage/options/                /docs/options
 /caveats.html                       /docs/caveats
 /repl.html                          /repl/
+/en/repl                            /repl/
```
JLHwung, line 59:
```suggestion
/repl.html                          /repl
/en/repl                            /repl
```

actually `/repl/` is not working: https://deploy-preview-2742--babel.netlify.app/repl/

harish-sethuraman: aren't trailing slash supposed to work as well? Im confused?

JLHwung: Well the external scripts are sourced as `"./repl/repl.js"`, they won't be loaded if there is trailing slash because it would be `repl/repl/repl.js`.

PR #2746, opened 2023-03-02
Add proposal-regexp-modifiers docs

JLHwung:
Also add supported plugins to the sidebar and applied npm2yarn remarks.

Preview link: https://deploy-preview-2746--babel.netlify.app/docs/babel-plugin-proposal-regexp-modifiers

### DIFF
--- a/docs/plugin-proposal-regexp-modifiers.md
+++ b/docs/plugin-proposal-regexp-modifiers.md
@@ -1,0 +1,81 @@
+---
+id: babel-plugin-proposal-regexp-modifiers
+title: "@babel/plugin-proposal-regexp-modifiers"
+sidebar_label: regexp-modifiers
+---
+
+## Example
+
+### `i` modifier
+```js title="input.js"
+// matches Aa and aa
+const regex = /(?i:a)a/
+```
+will be transformed to
+```js title="output.js"
+const regex = /(?:[Aa])a/
+```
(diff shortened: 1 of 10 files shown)


docs/plugin-proposal-regexp-modifiers.md
```diff
@@ -0,0 +1,81 @@
...
+
+### multiple modifiers
+You can also apply multiple modifiers:
+
+```js
+// matches Aa, aa, A\naa, etc. but not A\na
+const regex = /(?im:^a)a/
+```
+
+This plugin only supports `i`, `m` and `s` modifier.
```
nicolo-ribaudo, line 47:
```suggestion
This proposal only supports `i`, `m` and `s` as inline modifiers.
```

nicolo-ribaudo (approved):

docs/plugin-proposal-regexp-modifiers.md
JLHwung: The modifiers proposal also mentions the `x` flag, though it is covered in another proposal: https://github.com/tc39/proposal-regexp-x-mode. If we ever support that proposal, we should give a note here that to use the `x` modifier you should also install `@babel/plugin-proposal-regexp-x-mode`.

existentialism (approved):

Issue #2551, opened 2021-07-03
lol the demo does nothing


RedGuy12:
![image](https://user-images.githubusercontent.com/61329810/124337411-175d3c00-db68-11eb-9e16-034f690b678b.png)
most demos have nothing changed, like that ^
some do, but most don't

nicolo-ribaudo:
Uh it's because we used old example that now don't need to be compiled anymore in most browsers :grimacing:

RedGuy12:
yeah
i see two possible solutions
1. change your browserslist
2. update the examples (prob better)

PR #2750, opened 2023-03-03
update landing page examples

JLHwung:
Closes https://github.com/babel/website/issues/2551

### DIFF
--- a/js/minirepl.js
+++ b/js/minirepl.js
@@ -3,15 +3,8 @@
 import debounce from "lodash.debounce";
 
 const miniReplExamples = [
-  `@define("my-element")
-class MyElement {}`,
-  `element
-|> jQuery.parseHTML(match[1], %, true)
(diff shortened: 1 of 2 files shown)


liuxingbaoyu:
These few examples don't look very good.
The output of the decorator is very, very long and completely unreadable.
The pipeline-operator is not that famous.
record-and-tuple relies entirely on polyfills.

Unfortunately I didn't think of a good example either.

JLHwung:
I have removed decorators and pipeline examples. The new examples are from ES2025 and stage 3 explicit-resource-management proposal. Please take a look.

liuxingbaoyu (approved):
Thanks!
Non-blocking: Would `proposal-throw-expressions` be better than `proposal-do-expressions`?

JLHwung:
I am fine with either proposal, as long as we showcase something still in early stage. That said, I have updated the examples for throw expressions.

PR #2753, opened 2023-03-08
Update babel config

JLHwung:


babel.config.js
```diff
@@ -1,34 +1,37 @@
...
+      // This is browserslist defaults
+      browsers: ["> 0.5%", "last 2 versions", "Firefox ESR", "not dead"],
+    },
     presets: [
       [
         "@babel/preset-env",
         {
           useBuiltIns: "entry",
-          corejs: 3,
+          corejs: require("core-js/package.json").version,
```
JLHwung, line 15: We could use `babel.config.json` if the `corejs` version can be obtained via a magic string, e.g. `"corejs": "infer-from-node_modules"`

liuxingbaoyu (approved):

PR #2781, opened 2023-05-09
Fix repl CI build

JLHwung:
The REPL CI build is broken because the script `./repl.js` can not be found from `/repl/build/12345`. In this PR we manually inject `/repl/repl.js` so that it can be imported under `/repl/build/12345`

We also remove another unused js import `minirepl` as it has been handled by docusaurus.

Preview URL:

https://deploy-preview-2781--babel.netlify.app/repl/build/54393
https://deploy-preview-2781--babel-next.netlify.app/repl/build/54393

### DIFF
--- a/js/repl/index.html
+++ b/js/repl/index.html
@@ -399,6 +399,7 @@
          rel="stylesheet"
          /> -->
     <script src="/js/codetabs.js"></script>
+    <script defer src="/repl/repl.js"></script>
   </head>
 
   <body>
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -10,7 +10,6 @@ const config = {
   },
   entry: {
     repl: "./js/repl/index.tsx",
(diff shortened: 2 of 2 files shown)


JLHwung:
It is working: https://deploy-preview-2781--babel-next.netlify.app/repl/build/54393/#?browsers=ie%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=PQHQDg3gqgdglgMzgUwCYH0CSrkHsDmATgIZgAWAvsAK5A&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.21.8%2Bpr.15616&externalPlugins=&assumptions=%7B%7D

PR #2783, opened 2023-05-10
Use builtin plugins when available and provide default options for some external plugins

JLHwung:
In this PR we provide builtin plugins available in `@babel/standalone` when the external plugins is an official Babel plugin. In this way we can reduce dependency on the rollup service and improve external plugin loading time.

Then we provide default options for decorators and pipeline plugins. Ideally we should provide an UI to modify such options, but for now at least we can provide a working plugin demo.

[Preview link (decorator transform only)](https://deploy-preview-2783--babel.netlify.app/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=AIMw9mBQDGA2CGBnRACAwig3pFKDEADgNw4qIAu85AltCgLYAUAlFgL6RtA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=react&prettier=false&targets=&version=7.21.8&externalPlugins=%40babel%2Fplugin-proposal-decorators%407.21.0&assumptions=%7B%7D)

### DIFF
--- a/js/repl/Repl.tsx
+++ b/js/repl/Repl.tsx
@@ -102,6 +102,31 @@ function hasOwnProperty(obj, string) {
   return Object.prototype.hasOwnProperty.call(obj, string);
 }
 
+function provideDefaultOptionsForExternalPlugins(pluginName, babelVersion) {
+  switch (pluginName) {
+    case "@babel/plugin-proposal-decorators": {
+      if (compareVersions(babelVersion, "7.21.0") >= 0) {
+        return { version: "2023-01" };
+      } else if (compareVersions(babelVersion, "7.19.0") >= 0) {
+        return { version: "2022-03" };
+      } else if (compareVersions(babelVersion, "7.17.0") >= 0) {
+        return { version: "2021-12" };
+      } else if (compareVersions(babelVersion, "7.0.0") >= 0) {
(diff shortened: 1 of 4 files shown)


js/repl/Repl.tsx
```diff
@@ -102,6 +102,31 @@ function hasOwnProperty(obj, string) {
   return Object.prototype.hasOwnProperty.call(obj, string);
 }
 
+function provideDefaultOptionsForExternalPlugins(pluginName, babelVersion) {
+  switch (pluginName) {
+    case "@babel/plugin-proposal-decorators": {
+      if (compareVersions(babelVersion, "7.21.0") >= 0) {
+        return { version: "2023-01" };
```
nicolo-ribaudo, line 109:
```suggestion
      if (compareVersions(babelVersion, "7.22.0") >= 0) {
        return { version: "2023-05" };
      } else if (compareVersions(babelVersion, "7.21.0") >= 0) {
        return { version: "2023-01" };
```

This will be released soon

nicolo-ribaudo (approved):

PR #2798, opened 2023-06-01
Update polyfill.md

keer-tea:
After my testing, it depends on the version of the `@babel/plugin-transform-regenerator` plugin whether the auxiliary function for the generator will be produced. It is not related to the version of `@babel/core`.

If the `transforme-regenerator` plugin version is lower than 7.18.0, the compiled file will not inline the generator auxiliary function. If the version of the `transform-regenerator` plug-in is higher than or equal to 7.18.0, auxiliary inline function will be generated.

Here is a warehouse address for my testing: https://github.com/keer-tea/babel-regenerator-test

There is one type of warehouse, with versions of both packages less than 7.18.0. In Warehouse 2, the `@babel/core` version is less than 7.18.0, and the `@babel/plugin-transform-regenerator` version is greater than 7.18.0. In Warehouse 3, the `@babel/core` version is greater than 7.18.0, and the `@babel/plugin-transform-reenerator` version is less than 7.18.0.

I hope to correct this mistake.

### DIFF
--- a/docs/polyfill.md
+++ b/docs/polyfill.md
@@ -9,7 +9,7 @@ title: "@babel/polyfill"
 > import "core-js/stable";
 > ```
 >
-> If you are compiling generators or async function to ES5, and you are using a version of `@babel/core` or `@babel/plugin-transform-regenerator` older than `7.18.0`, you must also load the [`regenerator runtime`](https://github.com/facebook/regenerator/tree/main/packages/runtime) package. It is automatically loaded when using `@babel/preset-env`'s `useBuiltIns: "usage"` option or `@babel/plugin-transform-runtime`.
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Thanks for this PR! I agree that the current documentation is not precise, however I think it might be good enough.

The real requirement is "you must have at least version 7.18.0 of `@babel/plugin-transform-regenerator` and `@babel/helpers`". `@babel/helpers` is a dependency of `@babel/core`. `@babel/core` 7.18.0 depends on `@babel/helpers` 7.18.0 so it's guaranteed to get a new enough version. However, `@babel/core` 7.17.12 depends on `@babel/helpers` ^7.17.9, so it _might_ get an older version.

PR #2799, opened 2023-06-08
note on setClassMethods limit

JLHwung:
Resolves https://github.com/babel/babel/issues/15686.

### DIFF
--- a/docs/assumptions.md
+++ b/docs/assumptions.md
@@ -288,7 +288,7 @@ a.b?.();
 
 ## `setClassMethods`
 
-When declaring classes, assume that methods don't shadow getters on the superclass or override any well-known symbol methods and that the program doesn't depend on methods being non-enumerable. Thus, it's safe to assign methods rather than using `Object.defineProperty`.
+When declaring classes, assume that methods don't shadow accessors or non-writable properties on the superclass prototype, and that the program doesn't depend on methods being non-enumerable. Thus, it's safe to assign methods rather than using `Object.defineProperty`.
 
 <div is="assumption-repl" data-assumption="setClassMethods" data-plugins="transform-classes">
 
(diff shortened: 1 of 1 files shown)


docs/assumptions.md
```diff
@@ -288,7 +288,7 @@ a.b?.();
 
 ## `setClassMethods`
 
-When declaring classes, assume that methods don't shadow getters on the superclass and that the program doesn't depend on methods being non-enumerable. Thus, it's safe to assign methods rather than using `Object.defineProperty`.
+When declaring classes, assume that methods don't shadow getters on the superclass or override any well-known symbol methods and that the program doesn't depend on methods being non-enumerable. Thus, it's safe to assign methods rather than using `Object.defineProperty`.
```
nicolo-ribaudo, line 291:
```suggestion
When declaring classes, assume that methods don't shadow accessors or non-writable properties on the superclass prototype, and that the program doesn't depend on methods being non-enumerable. Thus, it's safe to assign methods rather than using `Object.defineProperty`.
```

the problem is not about well-know symbols, but it's that Symbol.hasInstance has `writable: false`.

JLHwung: AFAIK all well known symbol methods are non-writable, so overriding these methods is a sufficient condition for `setClassMethods` assumption to break, yet of course it is not a necessary condition since you can define you own non-writable properties.

existentialism (approved):

docs/assumptions.md
nicolo-ribaudo: Only `@@hasInstance` is non-writable, all the others are!
```js
Object.getOwnPropertyDescriptor(Array.prototype, Symbol.iterator)
> Object { value: values(), writable: true, enumerable: false, configurable: true }
```

JLHwung: Oh my bad, I misread the MDN docs.

PR #2802, opened 2023-06-16
Replace many `>` notes with `:::` docusaurus syntax

nicolo-ribaudo:


docs/plugin-proposal-decorators.md
```diff
@@ -180,10 +180,10 @@ If you are already using `@babel/preset-env`, you can safely remove `@babel/plug
 }
 ```
 
+:::tip
+You can read more about configuring plugin options [here](https://babeljs.io/docs/en/plugins#plugin-options)
 :::
```
JLHwung, line 185:
```suggestion
:::

:::
```

PR #2806, opened 2023-07-20
Update(data,static): name, url, logo, and svg for Uploadcare

tenebricosa:
Hey folks 👋 I am a developer at Uploadcare, a file management and delivery platform, and we would like to add our company details to your wall of users.

As the DevRel of the company, I can ensure you that Babel is a vital part of our technology stack, so here is my shoutout:

"The good thing about Babel is that it has a lot of integrations with different frameworks, and it's really easy to set up. You can just configure it once and forget about it!"


JLHwung (approved):
Thanks.

PR #2808, opened 2023-07-25
learn.md: Remove enumerate from Proxy handler

HolgerJeromin:
This is not mentioned in the standard:

https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#table-proxy-handler-methods


JLHwung (approved):
Thanks.

PR #2810, opened 2023-08-09
update migration docs for decorator versions

JLHwung:
Updated per https://github.com/babel/babel/pull/15676.


docs/v8-migration.md
```diff
@@ -374,21 +374,23 @@ The following syntax plugins are no longer needed, you can safely remove them fr
...
       }]
     ]
   }
   ```
-  The syntax is the same, but you will need to rewrite your decorator functions. The spec repo provides [comparison between the latest version and the `2018-09` version](https://github.com/tc39/proposal-decorators#comparison-with-the-previous-stage-2-decorators-proposal). You can already migrate since Babel 7.22.0, using the `"version": "2023-03"` option of `@babel/plugin-proposal-decorators`.
+  The syntax is the same, but you will need to rewrite your decorator functions. The spec repo provides [comparison between the latest version and the `2018-09` version](https://github.com/tc39/proposal-decorators#comparison-with-the-previous-stage-2-decorators-proposal). You can already migrate since Babel 7.22.0, using the `"version": "2023-05"` option of `@babel/plugin-proposal-decorators`.
+
+  Although Babel 8 still supports the `legacy` version, it is advisable to migrate to the `2023-05` version as well: Both Babel 8 and TypeScript 5.0 support the `2023-05` version, while there are [a few behaviour differences](https://github.com/babel/babel/issues/8864#issuecomment-688535867) in the `legacy` version between Babel and `tsc`'s implementation.
```
nicolo-ribaudo, line 393:
```suggestion
  Although Babel 8 still supports the `legacy` version, it is advisable to migrate to the `2023-05` version regardless: both Babel 8 and TypeScript 5.0 support the `2023-05` version, while there are [a few behaviour differences](https://github.com/babel/babel/issues/8864#issuecomment-688535867) in the `legacy` version between Babel and `tsc`'s implementation.
```

nicolo-ribaudo (approved):

PR #2820, opened 2023-09-09
Update usage.md

X-oss-byte:
@babel-bot run Nextjs on @vercel


houtan-rocky (changes requested):
Hey,
Thanks for the update,
But I think this isn't the best place to put the next.js build ( at the top of usage guide)🤔
Please check https://babeljs.io/docs/presets#other-integrations 🤗

PR #2829, opened 2023-09-23
Fix desc of `useDefineForClassFields` in `transform-typescript`

liuxingbaoyu:


nicolo-ribaudo:
Don't we have the other behaviour with the setClassProperties assumption?

liuxingbaoyu:
Oh yes, I forgot about it!

nicolo-ribaudo (approved):

PR #2830, opened 2023-09-23
Some improvements and fixes

liuxingbaoyu:
I'm trying to fix old version babel support.

By the way, can your repl work normally when `yarn start`? An exception will occur in my repl locally.

### DIFF
--- a/js/repl/PluginConfig.ts
+++ b/js/repl/PluginConfig.ts
@@ -118,7 +118,7 @@ const replDefaults: ReplState = {
   targets: "",
   version: "",
   reactRuntime: "automatic",
-  decoratorsVersion: "2023-01",
+  decoratorsVersion: "2023-05",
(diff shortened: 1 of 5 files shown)


JLHwung:
> By the way, can your repl work normally when yarn start?

It works for me. What error is your console throwing?

liuxingbaoyu:
image
This is weird.🤔

JLHwung:
> image This is weird.🤔

There is a [SO issue](https://stackoverflow.com/questions/71846788/updating-react-scripts-version-from-3-4-0-to-5-0-0-causing-type-error-cannot-re) for another react project, maybe you can try purging node_modules and see if error persists.

JLHwung (approved):

liuxingbaoyu:
Thanks! Unfortunately the problem still exists, which is weird.🤦‍♂️

PR #2837, opened 2023-10-10
compress svg by SVGO

JLHwung:
In this PR we compressed current svg assets using SVGO.

The netlify will terminate the post-process asset optimization on Oct. 17. The html/js/css are already minified by docusaurus and other tools. So we only have to optimize the SVG assets.

There are quite a few `*.png` assets as well. I think we can compress them losslessly into WebP, supported by all major browsers, or do that a few years later.


nicolo-ribaudo (approved):

JLHwung:
> https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset

The MDN also mentions `<picture>` element for offering alternative image types: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture. It seems to me the `srcset` is meant to support different devices with different screen ratios / dpis, etc.

PR #2842, opened 2023-10-23
update broken links

aroyan:
In [learn section](https://babeljs.io/docs/learn) links that refer to ES6 Compatibility table is moved from https://kangax.github.io/compat-table/es6 to https://compat-table.github.io/compat-table/es6


JLHwung (approved):
Thanks!

PR #2848, opened 2023-11-10
Update to Docusaurus 3

JLHwung:
In this PR we update docusaurus to v3 and React to v18. 

1. Docusaurus updates MDX to v3. Impacts to our docs:

- `{` must be escaped `\{` if it is intended to be a literal left brace in the docs
- `<=` is invalid in MDX. Most usages are `IE <= 10`. We replace them with `IE &leq; 10`
- When markdown directives are nested, for example when `:::babe7` directive contains a `:::tip` directive. The outer directive must contain _more_ colons than the inner ones. For example
  ```md
  ::::babel7
  :::tip
  This feature will be overhauled in Babel 8
  :::
  ::::
  ```
  If the outer directive starts with `:::`, the MDX parser will _not_ throw any error, but an extra `:::` text will be visible because the MDX parser ends the outer directives whenever it reads `:::`, even if the inner directive has not been closed yet
  Because of MDX v3, the directive `:::babel[78]` work like any other directives, which means we don't have to add extra linebreaks around the ending `:::`.

2. The customized `:::babel[78]` markdown directives are adjusted to MDX v3 AST. It is now much more straightforward as we can get rid of token reinterpretation
3. The react jsx runtime does not offer umd builds so we have to bundle the runtime with our REPL scripts

### DIFF
--- a/js/repl/ReplOptions.tsx
+++ b/js/repl/ReplOptions.tsx
@@ -436,19 +436,10 @@ class ExpandedContainer extends Component<Props, State> {
                     "reactRuntime",
                     (t) => t.value
                   )}
+                  value={presetsOptions.reactRuntime}
                 >
-                  <option
-                    value="automatic"
-                    selected={presetsOptions.reactRuntime === "automatic"}
-                  >
-                    Automatic
-                  </option>
-                  <option
-                    value="classic"
-                    selected={presetsOptions.reactRuntime === "classic"}
-                  >
-                    Classic
-                  </option>
+                  <option value="automatic">Automatic</option>
+                  <option value="classic">Classic</option>
                 </select>
               </PresetOption>
               <PresetOption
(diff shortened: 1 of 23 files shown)


js/repl/ReplOptions.tsx
```diff
@@ -465,12 +456,10 @@ class ExpandedContainer extends Component<Props, State> {
                     "decoratorsVersion",
                     (t) => t.value
                   )}
+                  value={presetsOptions.decoratorsVersion}
                 >
                   {Object.keys(DECORATOR_PROPOSALS).map((key) => (
-                    <option
-                      value={key}
-                      selected={key === presetsOptions.decoratorsVersion}
-                    >
```
nicolo-ribaudo, line 473: Doesn't removing `selected` change the behavior?

JLHwung: In react 18, passing `selected` to `option` is not supported: https://react.dev/reference/react-dom/components/select#caveats. I didn't check, maybe it was implemented in an earlier React version before.

nicolo-ribaudo (approved):

PR #2861, opened 2024-01-30
docs: update babel types

JLHwung:
Updated `@babel/types` docs for [Babel v7.23.9](https://github.com/babel/babel/releases/tag/v7.23.9).

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -118,7 +118,7 @@ See also `t.isAssignmentExpression(node, opts)` and `t.assertAssignmentExpressio
 
 AST Node `AssignmentExpression` shape:
 - `operator`: `string` (required)
-- `left`: `LVal` (required)
+- `left`: `LVal | OptionalMemberExpression` (required)
 - `right`: `Expression` (required)
 
 Aliases: [`Standardized`](#standardized), [`Expression`](#expression)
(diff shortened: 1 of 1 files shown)


docs/types.md
```diff
@@ -1426,6 +1427,23 @@ Aliases: [`Standardized`](#standardized), [`ModuleSpecifier`](#modulespecifier)
...
+```js title="JavaScript"
+t.importExpression(source, options);
+```
+
+See also `t.isImportExpression(node, opts)` and `t.assertImportExpression(node, opts)`.
+
+AST Node `ImportExpression` shape:
+- `source`: `Expression` (required)
+- `options`: `Expression` (default: `null`)
+- `phase`: `"source" | "defer"` (default: `null`, excluded from builder function)
```
liuxingbaoyu, line 1441: Is this expected?

liuxingbaoyu: Should we add it to the builder parameter? (excluded from builder function)

JLHwung: The source phase import is a stage-3 proposal, we can wait till it gets advanced. Note that it is not uncommon that builder function excludes some AST properties, for example `decorators` are also excluded from `ClassMethod`'s builder. We could add them to the builder when the decorator feature is finished.

liuxingbaoyu (approved):

PR #2862, opened 2024-02-02
Removes Algolia in favor of Orama

micheleriva:
As discussed with @nicolo-ribaudo, I'm opening a PR to add support for Orama search instead of Algolia. This results in faster searches (the search engine runs locally on the browser), with an open-source solution that does not require configuration or maintenance.

Happy to discuss this and make changes if you require so.


Daniel15:
> runs locally on the browser

How much does it increase the weight of the page? How large is the index?

micheleriva:
> How much does it increase the weight of the page? How large is the index?

For this specific website, it's `1.5MB`

Daniel15:
Is the index lazily loaded when you use the search, or is it always loaded even for people that don't ever use the search? 1.5MB is pretty heavy.

Daniel

micheleriva:
Right now, it's always loaded. We could make it lazy, but requires some development.

As an alternative, we can offer Orama Cloud for free for OSS projects, that way you'd have a more Algolia-like experience... but all the sites using Orama OSS are not having big problems with index size.

1.5MB is not lightweight... it's like loading an average-sized image. That may give some perspective I think :)

micheleriva:
Hi, @nicolo-ribaudo any update on this/anything I can do? Also, would you prefer moving to Orama Cloud (of course for free) or keeping the open-source one in this PR? It's a small website, so I personally believe the OSS one is enough for now.

PR #2865, opened 2024-02-26
Add `noUninitializedPrivateFieldAccess` assumption docs

nicolo-ribaudo:
Docs for https://github.com/babel/babel/pull/16267

### DIFF
--- a/docs/assumptions.md
+++ b/docs/assumptions.md
@@ -214,6 +214,30 @@ let getSum = (a, b) => {
 
 </div>
 
+## `noUninitializedPrivateFieldAccess`
+
+Assume that code never attempts to access private fields on classes before that they are initialized. For example:
+```js title="JavaScript"
+class Foo {
+  x = this.#y; // #y is not initialized yet
+  #y = 2;
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/assumptions.md
```diff
@@ -214,6 +214,30 @@ let getSum = (a, b) => {
...
+
+```js title="JavaScript"
+class MyClass {
+  static #id = 123;
+
+  method() {
+    return MyClass.#id;
+  }
+}
+```
```
nicolo-ribaudo, line 237: Without assumption:
```js
class MyClass {
  method() {
    return _id._;
  }
}
var _id = {
  _: 123
};
```
With assumption:
```js
class MyClass {
  method() {
    return _id;
  }
}
var _id = 123;
```

docs/assumptions.md
```diff
@@ -214,6 +214,30 @@ let getSum = (a, b) => {
 
 </div>
 
+## `noUninitializedPrivateFieldAccess`
+
+Assume that code never attempts to access private fields on classes before that they are initialized. For example:
```
existentialism, line 219:
```suggestion
Assume that code never attempts to access private fields on classes before they are initialized. For example:
```

JLHwung (approved):
Can you add a note that this assumption is supported since 7.24.0?

PR #2866, opened 2024-02-26
Add docs for plugin-proposal-json-modules

nicolo-ribaudo:
Docs for https://github.com/babel/babel/pull/16172

Preview: https://deploy-preview-2866--babel-next.netlify.app/docs/babel-plugin-proposal-json-modules


docs/plugin-proposal-json-modules.md
```diff
@@ -0,0 +1,89 @@
...
+<Tabs>
+  <TabItem value="browsers" label="Browsers" default>
+
+  ```js title="output.js"
+  const data = await fetch(import.meta.resolve("./data.json")).then(r => r.json());
+  ```
+  </TabItem>
+  <TabItem value="node-esm" label="Node.js (ESM)">
+
+  ```js title="output.js"
```
JLHwung, line 38: Nit:
```suggestion
  ```js title="output.mjs"
```

docs/plugin-proposal-json-modules.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Example
+```js title="input.js"
+import data from "./data.json" with { type: "json" };
+```
+
+will be transformed to
+
+<Tabs>
+  <TabItem value="browsers" label="Browsers" default>
```
JLHwung, line 30: Since the preface mentions that "whether the generated code should be compatible with Node.js, browsers, or both", can you add a tab here to show the output when targeting both platforms?

JLHwung (approved):
Can you add sidebar items to

https://github.com/babel/website/blob/773314d652753952c54dc7eaa33d2bba6043e06e/website/sidebars.js#L241 

We can split the group into "stage 3" and "early" as well.

PR #2872, opened 2024-02-27
Docs updates for `transform-runtime` cleanup

nicolo-ribaudo:
Docs for https://github.com/babel/babel/pull/16311

### DIFF
--- a/docs/v8-migration.md
+++ b/docs/v8-migration.md
@@ -403,11 +403,21 @@ The following syntax plugins are no longer needed, you can safely remove them fr
 
 ### `@babel/plugin-transform-runtime`
 
+![medium](https://img.shields.io/badge/risk%20of%20breakage%3F-medium-yellow.svg)
+
+- The `corejs` option has been removed ([#16311](https://github.com/babel/babel/pull/16311))
+
+  **Migration**: To inject polyfills, you can use [`babel-plugin-polyfill-corejs3`](https://github.com/babel/babel-polyfills/blob/main/packages/babel-plugin-polyfill-corejs3/README.md) or [babel-plugin-polyfill-corejs2](https://github.com/babel/babel-polyfills/blob/main/packages/babel-plugin-polyfill-corejs2/README.md) directly.
+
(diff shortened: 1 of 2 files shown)


docs/v8-migration.md
```diff
@@ -403,12 +403,22 @@ The following syntax plugins are no longer needed, you can safely remove them fr
 
 ### `@babel/plugin-transform-runtime`
 
+![medium](https://img.shields.io/badge/risk%20of%20breakage%3F-medium-yellow.svg)
+
+- The `corejs` option has been removed ([#16311](https://github.com/babel/babel/pull/16311))
+
+  **Migration**: To inject polyfills, you can use [`babel-plugin-polyfill-corejs3`](https://github.com/babel/babel-polyfills/blob/main/packages/babel-plugin-polyfill-corejs3/README.md) or [babel-plugin-polyfill-corejs2](https://github.com/babel/babel-polyfills/blob/main/packages/babel-plugin-polyfill-corejs2/README.md) directly.
```
zloirock, line 410: I'm confused. Sure, moving this functional to direct usage `core-js-pure` / `babel-plugin-polyfill-corejsx` is good. But many Babel helpers require modern standard library features. How will they work in the sandboxed (`core-js-pure`) case?

zloirock: Never mind, I found that it checks used `babel-plugin-polyfill-corejsx` transform.

nicolo-ribaudo: Yes, we'll keep publishing `@babel/runtime-corejs3` and it will be automatically used with this config:
```json
{
  "plugins": [
    "@babel/plugin-transform-runtime",
    "babel-plugin-polyfill-corejs3"
  ]
}
```

Note that this already happens now in the recent Babel 7 releases.

For Babel 8, I think we should stop publishing `@babel/runtime-corejs2` since at this point nobody should be using core-js 2 anymore.

zloirock: > For Babel 8, I think we should stop publishing `@babel/runtime-corejs2` since at this point nobody should be using `core-js` 2 anymore.

Definitely. Now this is already a dangerous legacy.

PR #2873, opened 2024-02-28
Support decorators proposal 2023-11

liuxingbaoyu:


nicolo-ribaudo:
Already done in https://github.com/babel/website/pull/2860 :)

PR #2879, opened 2024-03-10
Fix page not found error

juliojordan:
Hey there. If I'm right, the link is broken. This change aims to fix it.
> For full documentation on the Babel CLI see the [usage docs](https://babel.dev/docs/usage/cli/).


JLHwung (approved):
Thanks!

PR #2884, opened 2024-03-29
Fix usage on plugin-proposal-optional-chaining-assign

SharakPL:
Fixes https://github.com/babel/babel/issues/16395


liuxingbaoyu (approved):
Thanks!

PR #2889, opened 2024-05-11
docs: fix the `jsxPragma` option of the TypeScript plugin to have the correct default value.

Dunqing:


nicolo-ribaudo (approved):
Thank you!

Issue #2855, opened 2023-12-06
[Bug]: "Get browser-compatible JavaScript out" Scrollbar Not working 


harsitagarwalla187:
### 💻

- [ ] Would you like to work on a fix?

### Description
While i was trying to get the browser compatible javascript output, I found out that i am unable to scroll the output that you can see in the below image. 

### Current Behavior

Go babel [website](https://babeljs.io/), write some code in `Put in next-gen JavaScript` such that the output width should be greater than the `Get browser-compatible JavaScript out
display width`

![image](https://github.com/babel/babel/assets/84774840/dac1f91a-b81d-4f39-847c-4d27349b07aa)

### Expected behavior

One should be able to scroll in x-direction

prajyot004:
Can I work on this issue ?

liuxingbaoyu:
Feel free to open a PR! Thanks!

PR #2890, opened 2024-05-13
fix : "Get browser-compatible JavaScript out" Scrollbar Not working

prajyot004:
### Description
Fixed the issue of horizontal scrolling in homepage mini editor. Horizontal scroll was not working. I have linked the both videos of after and before fix.

Fixes : #2855

Before fix 

https://github.com/babel/website/assets/75519731/f3f602ca-b84e-4aff-8d54-4b60f42c5f81

After fix

https://github.com/babel/website/assets/75519731/65cabac9-6864-4612-96ab-7a1a57cf1e0d

### DIFF
--- a/js/minirepl.js
+++ b/js/minirepl.js
@@ -4,6 +4,7 @@ import debounce from "lodash.debounce";
 
 const miniReplExamples = [
   "element.index ?? -1;",
+  "const a = <h1>hello world</h1>;",
   "const styles = {\n" + "  ...defaults,\n" + '  color: "#f5da55",\n' + "};",
   "const city = address?.city",
   'var name = "Guy Fieri";\nvar place = "Flavortown";\n\n`Hello ${name}, ready for ${place}?`;',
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo (commented):
Thank you! Would it be possible to fix this in a way that makes the scrollbar visible only when it's actually needed?

prajyot004:
I'll try to do that

prajyot004:
@nicolo-ribaudo I looked for solution but there is only two ways either we need to keep the output window selectable not editable OR always keep the horizontalscrollbars visible. And on top of it we are keeping the error message window there only we are just change its opacity to 1 when error occurs the pace is  still occupied by error div  so it is making conflicts between scrollbar and error div.
As of now "ace editor" does not provide the property to focus on the scroll bar so there are only two possible solutions are there.

JLHwung:
Thank you for your PR.

This issue has been fixed in https://github.com/babel/website/pull/2750. We enabled the line wrapping for the landing page REPL so there won't be a horizontal scroll bar at all.

PR #2892, opened 2024-05-31
docs: update Babel/estree ast difference

sxzz:
`import * as ns from ''` differs between Babel AST and ESTree.


JLHwung (approved):
Thanks.

PR #2898, opened 2024-06-07
docs(options.md): clarify `--source-maps` CLI option

jedwards1211:
it's confusing.  Honestly it would be less confusing and IMO better if having `sourceMaps: true` in config files would make `@babel/cli` write `.map` files to disk.


JLHwung (approved):
Thank you.

liuxingbaoyu:
https://github.com/babel/babel/pull/14950
I remember that we have support for `if having sourceMaps: true in config files would make @babel/cli write .map files to disk`.
I tested it with this and the `.map` files were generated correctly.
```
echo '{"sourceMaps": true}' > .babelrc
mkdir src
echo 'export default 1' > src/index.js
babel src --out-dir lib
```

PR #2905, opened 2024-07-06
fix(docsearch): better text contrast for active hit in dark mode

Den-dp:
before:
image

after:
image

for light mode it stays without changes:
image


JLHwung (approved):
Thanks.

PR #2908, opened 2024-07-09
fix(options): drop visual noise

Armanio:
Hi!
I notice rubbish on [options](https://babeljs.io/docs/options#no-targets) page. My PR drop unused text.

Снимок экрана 2024-07-09 в 16 53 19

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -365,7 +365,7 @@ If a minor version is not specified, Babel will interpret it as `MAJOR.0`. For e
 
 #### No targets
 
-:::babel7
+::::babel7
 
 When no targets are specified: Babel will assume you are targeting the oldest browsers possible. For example, `@babel/preset-env` will transform all ES2015-ES2020 code to be ES5 compatible.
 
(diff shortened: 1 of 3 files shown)


JLHwung:
Thank you for your PR. Yes the `:::` is not meant to be in the final html, but removing it is not better as it is supposed to close the opening `:::babel8`, a custom markdown directive that we use to build Babel 8 docs website from the same markdown files.

Do you want to take a look at the remark plugin?
https://github.com/babel/website/blob/449f038a14f8fbb37d1590f745ed3055bb4dd136/website/docusaurus.config.js#L87-L119

Here we should remove everything inclusively between `:::babel8` and `:::` when the environment variable `BABEL_8_BREAKING` is not enabled.

JLHwung:
> Oh of course! `tip` directive is children `babel7`. Okay, what is correct behaviour? Its incorrect mark up or incorrect directive processing and we need support directive inside directive?

I think we should already support nested directives. I didn't investigate this yet, but my gut is that the remark AST is not correct, it seems like the `:::` is not recognized as the end of the `:::babel7` directive, but rather a paragraph. You can inspect the remark AST via the `root` variable.

JLHwung:
I just pushed the new fix, this is a leftover from https://github.com/babel/website/pull/2848#issue-1988368133. Thanks for reporting this issue.

PR #2946, opened 2024-09-04
Add Igalia as a base support sponsor

nicolo-ribaudo:
Igalia is now paying for me to work on Babel, so that I don't rely on OpenCollective anymore.


nicolo-ribaudo:
It's not showing up in the preview, but neither other sponsors from that file that are on the live website. Lets see what happens

PR #2965, opened 2024-09-16
Add migration docs for flow `enums` option

nicolo-ribaudo:
Docs for https://github.com/babel/babel/pull/16792


liuxingbaoyu:
The same content appears to be copied once?

JLHwung (approved):

nicolo-ribaudo:
> The same content appears to be copied once?

Once for the preset, once for the plugin. It's the same migration, but at least it's easier for people to find it based on how they are using flow.

PR #2967, opened 2024-09-21
Note section for Symbol.metadata in the plugin-proposal-decorators doc

naman1608:
In reference to the issue raised in the babel repository: https://github.com/babel/babel/issues/16838

### DIFF
--- a/docs/plugin-proposal-decorators.md
+++ b/docs/plugin-proposal-decorators.md
@@ -76,15 +76,6 @@ require("@babel/core").transformSync("code", {
   ]
 });
 ```
-When using decorators which either access or modify the metadata in the decorator context, you may need to use `Symbol.metadata`. When `Symbol.metadata` is not available, Babel defaults to `Symbol.for("Symbol.metadata")`: this may be incompatible with other packages that use a different fallback.
-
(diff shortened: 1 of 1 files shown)


docs/plugin-proposal-decorators.md
```diff
@@ -76,6 +76,26 @@ require("@babel/core").transformSync("code", {
...
+npm install core-js
+```
+
+Import the Metadata Polyfill
+
+```js title="Javascript"
+import 'core-js/proposals/decorator-metadata-v2.js';
+```
+
+Note: This import must come before any code that uses decorators or accesses `Symbol.metadata`.
```
nicolo-ribaudo, line 98:
````suggestion
When using decorators which either access or modify the metadata in the decorator context, you may need to use `Symbol.metadata`. When `Symbol.metadata` is not available, Babel defaults to `Symbol.for("Symbol.metadata")`: this may be incompatible with other packages that use a different fallback.

To ensure that `Symbol.metadata` is available globally and matches the symbol used by the Babel decorators plugin during transpilation, you will need to either include a polyfill that defines it, or define it yourself:

```js title="symbol-metadata-polyfill.js"
Symbol.metadata = Symbol.for("Symbol.metadata");
```

You can also use a third-party polyfill, such as `core-js/proposals/decorator-metadata-v2.js`. Make sure that the polyfill is executed before any code that uses decorators or accesses `Symbol.metadata`.
````

JLHwung (approved):
Thanks.

PR #2968, opened 2024-09-24
Update facebook/jsx link

skyclouds2001:
the default branch of [facebook/jsx](https://github.com/facebook/jsx/) has changed to `main`, the link is better to be updated, too


JLHwung (approved):
Thank you.

Issue #2969, opened 2024-09-26
Babel 8 repl does not work


nicolo-ribaudo:
When trying to use preset-env, it throws
> [BABEL] /repl.ts: @babel/preset-env: The 'loose' and 'spec' options have been removed, and you should configure granular compiler assumptions instead. See https://babeljs.io/assumptions for more information. (While processing: "base$0")

PR #2970, opened 2024-09-26
fix: REPL with `preset-env` in Babel 8

liuxingbaoyu:
Fixes: #2969

### DIFF
--- a/js/repl/PluginConfig.ts
+++ b/js/repl/PluginConfig.ts
@@ -66,7 +66,7 @@ const envPresetDefaults = {
     default: "3.21",
   },
   modules: {
-    default: false,
+    default: "false",
   },
 } as const;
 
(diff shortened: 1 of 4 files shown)


JLHwung (approved):

js/repl/PluginConfig.ts
```diff
@@ -66,7 +66,7 @@ const envPresetDefaults = {
     default: "3.21",
   },
   modules: {
-    default: false,
+    default: "false",
```
nicolo-ribaudo, line 69: Q: Why is this a string?

liuxingbaoyu: image
`select.value` does not accept `false`.

Thanks for your reminder, I checked the code and it handles the `false` of string.
image

liuxingbaoyu: No, it seems like it should be done another way. :)

JLHwung (approved):

PR #3013, opened 2024-11-12
Fix unused markdown directives warning

JLHwung:
Current main:

https://babel.dev/docs/babel-parser

image

This PR:

image

This error is caught by https://github.com/facebook/docusaurus/pull/9394. 

@slorber Can docusaurus throw on unused MDX directives? I can't find such an option so if I missed anything please let me know, thank you.

### DIFF
--- a/docs/parser.md
+++ b/docs/parser.md
@@ -348,18 +348,18 @@ When a plugin is specified multiple times, only the first options are considered
 
   - `decoratorsBeforeExport` (`boolean`)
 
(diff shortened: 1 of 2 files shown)


slorber:
We haven't made it configurable unfortunately.

If you think it would be useful we could add that. 

Please open an issue with the motivation so that we don't forget. I think it's reasonable to add `siteConfig.markdown.onUnusedDirective: 'throw'` or a similar option

PR #3019, opened 2024-11-29
Using unpkg instead of googleapis

liuxingbaoyu:
googleapis cannot be accessed in China. :)
Fixes https://github.com/babel/babel/discussions/16985

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -272,7 +272,7 @@ const siteConfig = {
 
   scripts: [
     {
-      src: "https://cdnjs.cloudflare.com/ajax/libs/jquery/1.9.1/jquery.min.js",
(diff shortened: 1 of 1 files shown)


website/docusaurus.config.js
```diff
@@ -272,7 +272,7 @@ const siteConfig = {
 
   scripts: [
     {
-      src: "https://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js",
+      src: "https://cdnjs.cloudflare.com/ajax/libs/jquery/1.9.1/jquery.min.js",
```
JLHwung, line 275: Can we import the `jquery` from `unpkg.com`, like we handle other resources? I don't have strong preference over specific CDN, either cloudfare or unpkg look good to me. However it would be great to stick with one consistently: so that we don't have two external dependencies that could effectively break the website.

PR #3022, opened 2024-12-06
[Babel 8] doc: Migration for `TSClassImplements` and `TSInterfaceHeritage`

liuxingbaoyu:
https://github.com/babel/babel/pull/16731

### DIFF
--- a/docs/v8-migration-api.md
+++ b/docs/v8-migration-api.md
@@ -110,8 +110,9 @@ Check out the [v8-migration guide](v8-migration.md) for other user-level changes
 
   __Migration__: If you have a customized plugin accessing `typeParameter` of a `TSMappedType` node, use `node.key` and `node.constraint` in Babel 8.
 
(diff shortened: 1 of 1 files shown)


existentialism (approved):

docs/v8-migration-api.md
```diff
@@ -110,6 +110,74 @@ Check out the [v8-migration guide](v8-migration.md) for other user-level changes
...
+  class C implements X<T> {}
+  interface I extends X<T> {}
+
+  // AST in Babel 7
+  {
+    type: "ClassDeclaration",
+    id: Identifier("C"),
+    implements: [
+      {
+        type: "TsExpressionWithTypeArguments",
```
JLHwung, line 130: Since we have also removed the `t.tsExpressionWithTypeArguments` builder in Babel 8, could you also add a migration note within the `@babel/types` section?

liuxingbaoyu: Maybe just mention in the AST changes that the corresponding builders and validators are removed as well? Since we also removed the validators in `types` and `traverse`.

PR #3023, opened 2024-12-06
add TSTypeReference migration note

JLHwung:
Docs PR for https://github.com/babel/babel/pull/17008

### DIFF
--- a/docs/v8-migration-api.md
+++ b/docs/v8-migration-api.md
@@ -77,6 +77,38 @@ Check out the [v8-migration guide](v8-migration.md) for other user-level changes
     - For `node.parameters` in Babel 7, use `node.params` in Babel 8
     - For `node.typeAnnotation` in Babel 7, use `node.returnType` in Babel 8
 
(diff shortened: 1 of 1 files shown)


docs/v8-migration-api.md
```diff
@@ -77,6 +77,38 @@ Check out the [v8-migration guide](v8-migration.md) for other user-level changes
...
+    typeArguments: {
+      type: "TSTypeParameterInstantiation",
+      params: [{
+        type: "TSStringKeyword"
+      }]
+    }
+  }
+  ```
+
+  **Migration**: If you have a customized plugin accessing `typeParameters`, use `typeArguments` in Babel 8
```
nicolo-ribaudo, line 110: Sometimes the migration seems so obvious that probably it doesn't make sense to write it explicitly 😅 (not just here, but in many other AST changes).

If we already say "we renamed `foo` to `bar`", it's obvious that the migration is to rename `foo` to `bar`.

JLHwung, line 110:
```suggestion
```
Okay, removed this note.

nicolo-ribaudo (approved):

PR #3032, opened 2024-12-16
add docs for typeArguments renaming

JLHwung:
Also grouped other typeArguments renaming into a single note.

Docs for https://github.com/babel/babel/pull/17020


docs/v8-migration-api.md
```diff
@@ -77,65 +77,250 @@ Check out the [v8-migration guide](v8-migration.md) for other user-level changes
...
+        params: [{
+          type: "TSStringKeyword"
+        }]
+      }
     }
-  }
 
-  // AST in Babel 8
-  {
-    type: "TSTypeReference",
```
JLHwung, line 99: Also fixed a docs error, the type here should be `TSTypeQuery`, it is overlooked during copy and paste.

nicolo-ribaudo (approved):

Issue #1931, opened 2018-12-27
Unable to run project on windows


saranshkataria:
When I try and run the setup locally, I keep getting errors. A couple of initial ones were that NODE-ENV cannot be set directly in windows, it is used as SET NODE_ENV=production. Similarly, cp does not exist in windows. It should be -cmd copy and so on. Not sure if a third party package is used for windows checks or how it should work ideally, would be happy to send in PR's if get some help in pointing towards the right direction.

xtuc:
Where did you see `NODE-ENV`? it should be `NODE_ENV `.

I agree that we should use something portable, Node based. Do you have a suggestion?

nicolo-ribaudo:
Would something like `cygwin` or `msys2` work to setup babel on windows?

saranshkataria:
WSL and cygwin would be workarounds, not solving the exact solution. I agree with @xtuc that it should be something node based. [Cross Env](https://www.npmjs.com/package/cross-env) is one possible solution which I have found that seems to work across platforms. 

For copying, webpack-copy-plugin should do the job I think?

PR #3048, opened 2025-02-02
Update setup guide and make commands compatible with Windows

Lehoczky:
Hi there 👋

I wanted to contribute to the project, but ran into some issues setting up the project. This PR updates the README to be up to-date, and modifies the yarn commands to be runnable on Windows. See my comments below.

Related: https://github.com/babel/website/issues/1931


package.json
```diff
@@ -8,8 +8,8 @@
     "bootstrap": "run-s bootstrap:sponsors build",
     "bootstrap:sponsors": "node ./scripts/download-sponsors.js",
     "build": "run-s build:codemirror6 build:repl build:docusaurus build:redirects",
-    "build:docusaurus": "cd website && yarn && yarn build && mv build babel && mkdir build && mv babel build/babel",
```
Lehoczky, line 11: `mv` does not exist for Windows, and there is an [--out-dir](https://docusaurus.io/docs/cli#docusaurus-build-sitedir) option for Docusaurus, so we don't have to move files around manually

package.json
```diff
@@ -8,8 +8,8 @@
...
-    "build:redirects": "cp ./_redirects ./website/build/babel",
```
Lehoczky, line 12: `cp` does not exist on Windows, but we can use node's [fs.cpSync() function](https://nodejs.org/api/fs.html#fscpsyncsrc-dest-options), which is available since v16.7.0.

This is how removing and copying files are done in the SvelteKit repo too: [link to source](https://github.com/sveltejs/kit/blob/main/packages/adapter-netlify/package.json#L36)

README.md
```diff
@@ -7,13 +7,12 @@ This is the source for the [babeljs.io](https://babeljs.io) website; feel free t
 
 ### Setup
 
-Node: Check that Node is installed with version 10.19.0 and up. You can check this with node -v.
+Node: Check that Node is installed with version 20.5.0 and up. You can check this with `node -v`.
 
-Yarn: Make sure that Yarn 1 is installed with version >= 1.19.0.
+Yarn: Make sure that Yarn 4 is installed with version >= 4.0.1.
```
nicolo-ribaudo, line 12: Yarn 1.19 should also be fine, since it will automatically use the Yarn version installed in this repository. Anything higher than 1.19 is ok.

nicolo-ribaudo (commented):
Thank you!

PR #3049, opened 2025-02-02
Page speed improvements

Lehoczky:
This PR aims to speed up the documentation site by:
- removing jquery in favour of native JavaScript methods (42.2kb)
- deleting old, unused scripts, like `repl-page-hacks.js` and `code-blocks-buttons.js` (1.3kb + 4.6kb because clipboard.js can be dropped in this case too)
- only loading `ace` on the home page, where it is used (123kb!)


JLHwung (approved):
This PR looks good to me, thank you.

PR #3050, opened 2025-02-07
Improve generator docs and use named export

vovkasm:
Named export was introduced in babel/babel#17100 and allows to import generator from node.js esm scripts from cjs built generator.

### DIFF
--- a/docs/generator.md
+++ b/docs/generator.md
@@ -59,11 +59,8 @@ Babel generator contains exactly one function `generate`. It can be imported in 
 ```js title="JavaScript"
 const generate = require("@babel/generator");
 const { generate } = require("@babel/generator");
-import generate from "@babel/generator";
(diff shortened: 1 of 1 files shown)


docs/generator.md
```diff
@@ -52,6 +52,19 @@ const output = generate(
...
+Babel generator contains exactly one function `generate`. It can be imported in different ways.
+
+```js title="JavaScript"
+const generate = require("@babel/generator");
+const { generate } = require("@babel/generator");
+import generate from "@babel/generator";
+import { generate } from "@babel/generator";
+```
+
+The last option works for node.js in esm modules.
```
nicolo-ribaudo, line 66:
````suggestion
```js title="JavaScript"
const generate = require("@babel/generator");
const { generate } = require("@babel/generator");
import { generate } from "@babel/generator";
```
````

Issue #1986, opened 2019-03-14
The MatchPattern docs should clarify that Babel's glob implementation is extremely basic


loganfsmyth:
Ideally https://babeljs.io/docs/en/options#matchpattern should have more examples of what does and does not work as a glob match and that it is not simply passing through to the `glob` module.

mbtools:
It's a terrible trap. It took me (and all sort of AIs) a full day to realize that the matching is *NOT* a glob.

Here's the corresponding coding:

https://github.com/babel/babel/blob/33110bf6a6e47ee084a79050a20ea67d9317af20/packages/babel-core/src/config/pattern-to-regex.ts#L18-L53

PR #3066, opened 2025-03-14
Expand docs about `MatchPattern`

mbtools:
Countless devs have fallen into the trap of trying to use full `glob` patterns for matching. 

This PR...

- Clarifies the string matching logic 
- Notes that it's *not* a glob
- Provides examples on how matching works
 
Closes https://github.com/babel/website/issues/1986


JLHwung (approved):
Thank you!

PR #3068, opened 2025-03-15
Make ReplOptions wider

fisker:
image
image

I can make `usebuiltins` option use less space, but feel use `minWidth` more flexible.


JLHwung (approved):
Thank you.

PR #3076, opened 2025-04-03
feat: Use monaco editor

liuxingbaoyu:
This was tested manually, so let me see if everything works fine after deployment.


js/repl/Monaco.tsx
```diff
@@ -0,0 +1,139 @@
...
+    React.useState<monaco.editor.IStandaloneCodeEditor | null>(null);
+
+  useEffect(() => {
+    shikiToMonaco(highlighter, monaco);
+    setEditor(
+      (editor = monaco.editor.create(container.current, {
+        padding: {
+          top: 2,
+        },
+        theme: "dark-plus",
```
JLHwung, line 45: Can we set the light theme by default? Ideally it should adapt to the user colour scheme preference.

liuxingbaoyu: Since our tabs and options are dark, I prefer to always use a dark theme. :)

JLHwung: The current REPL supports user colour scheme preference. If you prefer dark scheme, the REPL will adapt to the system settings (we should have also migrated the top-right colour scheme switch button to REPL page):

Light mode:
image

Dark mode:
image

The tabs/options are for the contrast with the editor, it does not mean we should use dark unconditionally: Similar to the landing page: the upper part is always dark

Light mode:
image

Dark mode:
image

liuxingbaoyu: L~M2AIG3)XNQKTKZPLH@~HJ
M$W`5RQVW`PCHEU81LEEPO
To be honest, our current light mode doesn't seem very ideal.
As an example, typescript-eslint is more like full light mode.
Of course I am in favor of introducing a button so that users can quickly switch.

JLHwung: I am all for changing the setting panel accordingly. But to force users with light mode to use dark mode editor is another thing. The current behaviour is that, if the system is in light mode, the editor will be in light mode as well, so as to the dark mode.

liuxingbaoyu:
Yes I'm trying.
Do you remember why we split the REPL?

JLHwung:
> Yes I'm trying. Do you remember why we split the REPL?

It was separated when we are migrating to docusuarus 2 so that we can speed up the migration process: https://github.com/babel/website/pull/2622#issuecomment-1036208436

Now that we are using docusaurus v3 and planing a REPL overhaul, maybe we can try to see if we can move it back and unify the navbar. @nicolo-ribaudo WDYT?

JLHwung:
The new build looks much better. Now that the monaco editor will capture the command/ctrl key so cmd/ctrl+K will not work, can we hide the `DocSearch-Button-Keys`? 

image

js/repl/Monaco.tsx
```diff
@@ -0,0 +1,228 @@
...
+          top: 2,
+        },
+        // https://github.com/microsoft/monaco-editor/issues/4311
+        // automaticLayout: true,
+        model: fastMode
+          ? monaco.editor.createModel(
+              "",
+              "javascript",
+              monaco.Uri.file("output/output.tsx")
+            )
```
liuxingbaoyu, line 86: This is a somewhat bad hack.

The syntax highlighter registered by shiki will cause performance degradation, which is obvious when using `timeTravel` on large files.
And monaco does not support using different syntax highlighters for the same language.
So we set the language of the output code to `js`, and notice that it will not warn even if we output ts, but there is no syntax highlighting for ts keywords.

What do you think about using this solution, or not using it (because it is not very slow, just a little slow), or continuing to use codemirror for output code?

JLHwung: The current main is not significantly faster ([repl](https://babel.dev/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=IYZwngdgxgBAZgV2gFwJYHsI2ACgJQwDeAUDGTMQL5A&forceAllTransforms=true&modules=commonjs&shippedProposals=false&evaluate=false&fileSize=false&timeTravel=true&sourceType=script&lineWrap=true&presets=env&prettier=false&targets=&version=7.27.0&externalPlugins=&assumptions=%7B%7D)). I am good with the current status. 

To improve the time travel performance, maybe we can lazy highlight the immediate step output?

liuxingbaoyu: Do you mean not highlighting when the user switches tabs, but highlighting after a few hundred milliseconds without new switching?

liuxingbaoyu: I recorded some gifs to show the difference visually.
Note the difference between the highlighted themes in normal and fast mode.

main
7f68d733-63ac-44df-9905-2ad31c53ed18
normal
d53cf2fa-92eb-441c-a0bf-6dbcedc65e64
fastMode
3660f24b-24cf-4b21-bab7-2536ddb6619a

JLHwung: Oh I mean the page load time: the time spent between opening the URL until the highlighted code is rendered.

The time spent switching tab does seem a bit slower, but I think it is still acceptable.

Aside: Now comparing this branch vs main, I do notice that the editor font size of monaco (12px) is smaller in than codemirror (0.875rem), which is also the font-size of other GUI text. Could you update the font-size of monaco?

liuxingbaoyu: Unfortunately monaco only supports px, I tried modifying the css properties but that resulted in an exception.

fisker:
The loading page is no longer fullscreen 

image

js/repl/Monaco.tsx
JLHwung: OK, in that case we should set it to 14px then.

PR #3087, opened 2025-05-14
Another pass through v8 migration docs

nicolo-ribaudo:
This PR is mostly moving things around to find them easier to use:
- rather than splitting the breaking changes in "config" and "compilation", that are highly related to each other anyway, it just splits them based on which packages are affected
- it sorts the packages by:
  - core, parser, generator (which are used by everybody)
  - preset-env
  - preset-react and JSX packages
  - preset-typescript and TS packages
  - preset-flow and dlow packages
  - other -transform- plugins
  - other -proposal- plugins
  - other packages (eslint-parser and node)

### DIFF
--- a/docs/v8-migration.md
+++ b/docs/v8-migration.md
@@ -3,15 +3,17 @@ title: "Upgrade to Babel 8"
 id: v8-migration
 ---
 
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Merging because it will cause conflicts with everything otherwise, but we can iterate through new PRs :)

I'll also go through the docs for integrations now.

PR #3092, opened 2025-05-23
Add v8 migration summary for steps to take during Babel 7

nicolo-ribaudo:
Hopefully this will simplify the migration, as it will help people incrementally prepare their codebase for Babel 8.

### DIFF
--- a/docs/v8-migration.md
+++ b/docs/v8-migration.md
@@ -11,6 +11,21 @@ If you are working directly with Babel's API (because, for example, you maintain
 
 > If you are upgrading from Babel 6, please check the [Babel 7 migration guide](./v7-migration.md) first.
 
+## Getting ready {#getting-ready}
+
+Before migrating to Babel 8, we recommend that you update your Babel 7 configuration to align it with Babel 8's new defaults. Once your build process is working again with Babel 7, you can then upgrade to Babel 8.
+
+You should read this full document to understand what options you need to change and what plugins you will need to remove, but here are the most important steps that you should take to modernize your config:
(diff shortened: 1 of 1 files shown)


docs/v8-migration.md
```diff
@@ -11,6 +11,21 @@ If you are working directly with Babel's API (because, for example, you maintain
...
+- If you are using the TypeScript or Flow presets, replace the `isTSX` and `allExtensions` options with [`ignoreExtensions`](#babel-preset-typescript).
+- If you are transforming TypeScript or Flow, set the `allowDeclareFields` option to true (https://babeljs.io/docs/babel-preset-typescript#allowdeclarefields).
+- If you are using [`@babel/plugin-transform-runtime`'s `corejs`](https://babeljs.io/docs/babel-plugin-transform-runtime#corejsu) or [`@babel/preset-env`'s `useBuiltIns`](https://babeljs.io/docs/babel-preset-env#usebuiltins) options, remove it and use [`babel-plugin-polyfill-corejs3`](https://github.com/babel/babel-polyfills/tree/main/packages/babel-plugin-polyfill-corejs3) instead.
+- If you are using `@babel/plugin-proposal-decorators`, set its [`version`](https://next.babeljs.io/docs/babel-plugin-proposal-decorators/#version) option to `legacy` or `2023-11`.
```
JLHwung, line 26: Can you use the relative markdown links? In the future if `babeljs.io` points to the Babel 8 docs, these links should still work for `v7.babeljs.io`.

nicolo-ribaudo: Actually, maybe these links should just point to v7? Since they are about what to do _before_ upgrading to v8.

JLHwung: The reason I prefer relative markdown links, such as `./babel-plugin-transform-runtime.md#corejsu` is that docusaurus can check whether this link is still valid when we change the markdown structure. Since the links here are mostly option documents, how about a relative link to current docs and then fallback to the v7 doc if such option has been removed?

JLHwung (approved):

existentialism (approved):

PR #3093, opened 2025-05-29
rename proposal-explicit-resource-management to transform-*

JLHwung:
In this PR we rename the proposal-explicit-resource-management docs and update the example output.

The REPL link in the plugin docs is also updated.


JLHwung:
This PR should be merged after https://github.com/babel/babel/pull/17340 is released, otherwise the landing page REPL will break.

PR #3105, opened 2025-06-21
Improve REPL build process

liuxingbaoyu:
Merge the REPL build process into the website.


JLHwung:
The loading transition page in this PR becomes text-only (There is only "Loading" on the top-left with a white background), while the current main features the the loading icon in the middle of the dark background. Could you take a look?

website/src/pages/repl.js
```diff
@@ -1,22 +0,0 @@
...
-    const script = document.createElement("script");
-    script.src = "/repl/index.js";
-    document.body.appendChild(script);
-    return () => {
-      document.body.removeChild(script);
-    };
-  }, []);
-  return (
-    <div>
-      <Layout noFooter={true}>
```
JLHwung, line 15: The `noFooter={true}` should be added to the `repl.tsx`

JLHwung: The Footer is visible if one scrolls to the bottom from the left toolbar, hence we used the `noFooter` property.

image

website/src/components/repl/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+export default () => {
+  return (
+    <BrowserOnly>
+      {() => {
+        return (
+          <React.Suspense
+            fallback={
+              <div className={loadingStyles.loader}>
+                <div className={loadingStyles.loaderContent}>
+                  "Loading Babel..."
```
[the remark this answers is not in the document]
JLHwung: Nit: Could we extract the fallback into a new component?

JLHwung (approved):

PR #3110, opened 2025-06-26
Enable prettier and update docusaurus

liuxingbaoyu:


website/docusaurus.config.ts
```diff
@@ -144,6 +144,9 @@ const siteConfig: Config = {
   future: {
     // See https://docusaurus.io/blog/releases/3.6
     experimental_faster: true,
+    v4: {
+      removeLegacyPostBuildHeadAttribute: true,
```
liuxingbaoyu, line 148: I made some attempts to enable `useCssCascadeLayers` but failed.🤦‍♂️

JLHwung (approved):

PR #3118, opened 2025-07-15
Allows manual editing of configuration

liuxingbaoyu:
This PR does not integrate manual editing with visual editing, which will be done in a subsequent PR.
The first commit moves some files around for better readability.

### DIFF
--- a/website/src/components/repl/Tabs.tsx
+++ b/website/src/components/repl/Tabs.tsx
@@ -9,7 +9,7 @@ export type TabsOptions = {
 
 export default function Tabs({ current, labels, onClick }: TabsOptions) {
   return (
-    <ul className={`pills padding--xs ${css({ "--ifm-list-margin": 0 })}`}>
+    <ul className={`pills padding--sm ${css({ "--ifm-list-margin": 0 })}`}>
       {labels.map((label, index) => (
         <li
           key={index}


website/src/components/repl/Tabs.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export type TabsOptions = {
+  current: string;
+  labels: string[];
+  onClick: (current: string) => void;
+};
+
+export default function Tabs({ current, labels, onClick }: TabsOptions) {
+  return (
+    <div className={`padding--xs ${styles.container}`}>
```
JLHwung, line 12: Could we use the pills component here?

https://infima.dev/docs/components/pills

website/src/components/repl/lib/uriUtils.ts
```diff
@@ -75,28 +76,36 @@ const parseQuery = () => {
...
 const updateQuery = (state: ReplState) => {
   const query = URL_KEYS.map((key) => {
-    if (state[key] == null) {
+    const value = state[key];
+    if (
+      value == null ||
+      value == "" ||
+      (typeof value === "object" && Object.keys(value).length === 0)
+    ) {
       return null;
```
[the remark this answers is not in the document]
JLHwung: I think it is fine to keep empty object as-is. They are more like an empty array but it is not nullish.

website/webpack.plugin.js
```diff
@@ -5,6 +5,9 @@ module.exports = function () {
     name: "docusaurus-webpack-plugin",
     configureWebpack() {
       return {
+        // There seems to be a rspack bug.
+        cache: false,
```
JLHwung, line 9: Has this issue been reported to the docusaurus or rspace repo?

website/src/components/repl/lib/uriUtils.ts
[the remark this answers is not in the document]
liuxingbaoyu: Ok, let's keep it as it is for now.
I plan to only add user-changed items to the url in a subsequent PR.

website/src/components/repl/Tabs.tsx
liuxingbaoyu: image
image
Both look good to me, the pills might need a border.

website/webpack.plugin.js
liuxingbaoyu: Not yet, the error message is pretty irrelevant.
I can't reproduce it now, so let's leave it as is.

JLHwung (approved):

PR #3122, opened 2025-07-28
Integrate configuration files with the UI

liuxingbaoyu:
Only the last commit is new, I will rebase after https://github.com/babel/website/pull/3118 is merged.

### DIFF
--- a/website/src/components/repl/ExternalPlugins.tsx
+++ b/website/src/components/repl/ExternalPlugins.tsx
@@ -76,8 +76,9 @@ export default class ExternalPlugins extends React.Component<Props, State> {
         {plugins.map((p) => (
           <li key={p.name}>
             <span className={currentStyles.pluginName}>
-              {p.name}{" "}
-              <span className={currentStyles.pluginVersion}>v{p.version}</span>
+              {p.name}
+              {/* {" "}
+              <span className={currentStyles.pluginVersion}>v{p.version}</span> */}
             </span>
             <button
               className={currentStyles.pluginActions}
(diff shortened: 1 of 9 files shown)


website/src/components/repl/ExternalPlugins.tsx
```diff
@@ -76,8 +76,9 @@ export default class ExternalPlugins extends React.Component<Props, State> {
         {plugins.map((p) => (
           <li key={p.name}>
             <span className={currentStyles.pluginName}>
-              {p.name}{" "}
-              <span className={currentStyles.pluginVersion}>v{p.version}</span>
+              {p.name}
+              {/* {" "}
+              <span className={currentStyles.pluginVersion}>v{p.version}</span> */}
```
liuxingbaoyu, line 81: Since we are using babel's built-in plugin, the version is meaningless.

website/src/components/repl/Repl.tsx
```diff
@@ -437,18 +596,14 @@ class Repl extends React.Component<Props, State> {
...
+    const shorthandName = this._getBundledPluginName(plugin);
+    if (shorthandName) {
+      return this._workerApi.registerPluginAlias(plugin.name, shorthandName);
     }
-    const bundledUrl = [
-      "https://bundle.run",
-      "https://packd.liuxingbaoyu.xyz",
-    ].map((url) => `${url}/${plugin.name}@${plugin.version}`);
+
+    const bundledUrl = ["https://packd.liuxingbaoyu.xyz"].map(
```
JLHwung, line 604: Aside: we should probably setup Babel's own `packd` instance in the future.

Issue #3044, opened 2025-01-22
Docs for "[babel 8] Remove bugfixes option"


nicolo-ribaudo:
https://github.com/babel/babel/pull/17078

Just so that I don't forget

PR #3129, opened 2025-09-11
update docs for bugfixes removal

JLHwung:
Docs PR for https://github.com/babel/babel/pull/17078.

Closes https://github.com/babel/website/issues/3044

Closes https://github.com/babel/website/issues/3132

### DIFF
--- a/docs/v8-migration.md
+++ b/docs/v8-migration.md
@@ -364,7 +364,7 @@ The following syntax plugins are no longer needed, you can safely remove them fr
 
   **Migration**: You will probably be fine with the new behaviour as Babel now tries to compile the broken syntax to the closest _non-broken modern syntax_ supported by your target browsers.
 
-  If anyhow you want to restore the Babel 7 behaviour for debug purpose, you can specify the [`exclude` option](./preset-env.md#exclude): `{ exclude: [/bugfix/] }` to explicitly exclude the bugfixes plugins. Note that doing so will bloat the output as Babel has to compile features to older targets.
+  If anyhow you want to restore the Babel 7 behaviour for debug purpose, you can specify the [`exclude` option](./preset-env.md#exclude): `{ exclude: [/bugfix/] }` to explicitly exclude the bugfixes plugins. Note that doing so will bloat the output as Babel has to compile features that, if not of some specific bug, would be generally supported by your targets.
 
 - Removed syntax plugins can not be used in `include` and `exclude` ([#15810](https://github.com/babel/babel/pull/15810))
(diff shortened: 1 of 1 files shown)


docs/v8-migration.md
```diff
@@ -342,31 +342,33 @@ The following syntax plugins are no longer needed, you can safely remove them fr
...
 
-- Enable the [`bugfixes`](./preset-env.md#bugfixes) option by default ([#13866](https://github.com/babel/babel/pull/13866))
+- The `bugfixes` options has been removed ([#13866](https://github.com/babel/babel/pull/13866), [#17078](https://github.com/babel/babel/pull/17078))
 
-  **Migration**: You will probably be fine with the new behaviour as Babel now tries to compile the broken syntax to the closest _non-broken modern syntax_ supported by your target browsers. If anyhow you want to restore the Babel 7 behaviour, you can specify `bugfixes: false`.
+  **Migration**: You will probably be fine with the new behaviour as Babel now tries to compile the broken syntax to the closest _non-broken modern syntax_ supported by your target browsers.
 
-- Removed syntax plugins can not be used in `includes` and `excludes` ([#15810](https://github.com/babel/babel/pull/15810))
+  If anyhow you want to restore the Babel 7 behaviour for debug purpose, you can specify the [`exclude` option](./preset-env.md#exclude): `{ exclude: [/bugfix/] }` to explicitly exclude the bugfixes plugins. Note that doing so will bloat the output as Babel has to compile features to older targets.
```
liuxingbaoyu, line 367: Will disabling the bugfix cause the output to become larger?

JLHwung: Yes, for example, given the input
```js
var a;
a?.b
```
with target `chrome: 90`,
the output with `{ bugfixes: true }` is
```js
var a;
a?.b;
```
whereas the output with `{ bugfixes: false }` is
```js
var a;
a === null || a === void 0 ? void 0 : a.b;
```

liuxingbaoyu (approved):

docs/v8-migration.md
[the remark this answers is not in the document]
nicolo-ribaudo:
```suggestion
  If anyhow you want to restore the Babel 7 behaviour for debug purpose, you can specify the [`exclude` option](./preset-env.md#exclude): `{ exclude: [/bugfix/] }` to explicitly exclude the bugfixes plugins. Note that doing so will bloat the output as Babel has to compile features that, if not of some specific bug, would be generally supported by your targets.
```

PR #3131, opened 2025-09-19
doc: Update `preset-react`

liuxingbaoyu:
Both plugins appear to be deprecated and are no longer included in `preset-react`.


liuxingbaoyu:
The build failure seems to be an opencollective issue, which has been reported.
https://github.com/opencollective/opencollective/issues/8196

PR #3135, opened 2025-09-30
Fix download-sponsors from Open Collective API

znarf:
- add User-Agent (instead of fetch default one that depends on environment)
- lower graphqlPageSize to 500 for members query (minimize potential timeouts)
- add OC_DEBUG to troubleshoot, track rate limiting and execution time

Address an issue reported here https://github.com/opencollective/opencollective/issues/8196


JLHwung (approved):
Thank you very much for a quick response.

The download-sponsor script is mostly copied from the webpack project. I guess they might be affected by the same issue. Could you also take a look if their recent website builds are normal?

liuxingbaoyu:
Thanks!

typescript-eslint appears to use the same API.
https://github.com/typescript-eslint/typescript-eslint/blob/f7e672eb3f57dcc5929116ae473e9de021197152/tools/scripts/generate-sponsors.mts#L47

PR #3139, opened 2025-10-28
feat: Add AST viewer

liuxingbaoyu:
Many thanks to typescript-eslint, the AST viewer is entirely based on theirs.

### DIFF
--- a/website/src/components/repl/Repl.tsx
+++ b/website/src/components/repl/Repl.tsx
@@ -171,7 +171,7 @@ class Repl extends React.Component<Props, State> {
       transitions: [],
       currentTransition: {},
       leftTab: "code",
-      rightTab: "code",
+      rightTab: "output",
     };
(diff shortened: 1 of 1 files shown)


fisker:
Stringify regular expression?

image

fisker:
Ah. not caused by this PR, the babeljs.io version also have this.

website/src/components/repl/Repl.tsx
```diff
@@ -398,19 +281,31 @@ class Repl extends React.Component<Props, State> {
             <div className={styles.codePanel}>
               <Tabs
                 current={state.rightTab}
-                labels={["code"]}
+                labels={["code", "input ast"]}
```
JLHwung:
```suggestion
                labels={["output", "input AST"]}
```
The "AST" should be uppercase only as it is an acronym.

PR #3211, opened 2026-06-14
[babel 8] Add `modules: "auto"` default change docs

nicolo-ribaudo:
Docs for https://github.com/babel/babel/pull/18069

### DIFF
--- a/docs/preset-env.md
+++ b/docs/preset-env.md
@@ -150,6 +150,12 @@ Setting this to `false` will preserve ES modules. Use this only if you intend to
 
 By default `@babel/preset-env` uses [`caller`](options.md#caller) data to determine whether ES modules and module features (e.g. `import()`) should be transformed. Generally `caller` data will be specified in the bundler plugins (e.g. `babel-loader`, `@rollup/plugin-babel`) and thus it is not recommended to pass `caller` data yourself -- The passed `caller` may overwrite the one from bundler plugins and in the future you may get suboptimal results if bundlers supports new module features.
 
+:::babel8
+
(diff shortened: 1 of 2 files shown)


docs/preset-env.md
```diff
@@ -150,6 +150,12 @@ Setting this to `false` will preserve ES modules. Use this only if you intend to
 
 By default `@babel/preset-env` uses [`caller`](options.md#caller) data to determine whether ES modules and module features (e.g. `import()`) should be transformed. Generally `caller` data will be specified in the bundler plugins (e.g. `babel-loader`, `@rollup/plugin-babel`) and thus it is not recommended to pass `caller` data yourself -- The passed `caller` may overwrite the one from bundler plugins and in the future you may get suboptimal results if bundlers supports new module features.
 
+:::babel8
+
+Unless the caller specifies otherwise, Babel will assume that it does support ES modules.
```
JLHwung, line 155:
```suggestion
Unless the caller specifies otherwise, Babel will assume that it does support ES modules syntax including `import`, `import()` and `import * as ns`.
```

Looking back we should have added a new caller flag for import attributes, in the future we may need more caller flags if other module proposals such as `import.defer` reaches stage 4.

nicolo-ribaudo: It depends on how transpilable are these things. If the _target_ supports import attributes but the _caller_ does, it's likely that transpiling them to a fetch call it's going to make it impossible for the bundler anyway.

nicolo-ribaudo:
I wanted to apply the suggestion above and instead merged 😅 Doing it in a separate commit.

PR #3213, opened 2026-06-17
Add Node.js 22 to supported versions of Babel 8.0.0

remcohaszing:
The blog post stated that Babel 8.0.0 supports Node.js 24 or greater. But based on the changelog, it looks like Node.js 22 is also supported. This aligns with the fact that Node.js 20 reached EOL this year, not Node.js 22.


nicolo-ribaudo (approved):
Thanks :)

PR #3214, opened 2026-06-17
Clarify `ignoreFunctionLength` description

jessealama:
Clarify that `ignoreFunctionLength` refers to non-default parameters not followed by default parameters.

### DIFF
--- a/docs/assumptions.md
+++ b/docs/assumptions.md
@@ -99,7 +99,7 @@ export const number = 2;
 
 ## `ignoreFunctionLength`
 
-Functions have a `.length` property that reflect the number of parameters up to the last non-default parameter. When this option is enabled, assume that the compiled code does not rely on this `.length` property.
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Thank you!

docs/assumptions.md
```diff
@@ -99,7 +99,7 @@ export const number = 2;
 
 ## `ignoreFunctionLength`
 
-Functions have a `.length` property that reflect the number of parameters up to the last non-default parameter. When this option is enabled, assume that the compiled code does not rely on this `.length` property.
+Functions have a `.length` property that reflect the number of parameters up to the last non-default parameter that is not followed by a default parameter. When this option is enabled, assume that the compiled code does not rely on this `.length` property.
```
nicolo-ribaudo:
```suggestion
Functions have a `.length` property that reflect the number of parameters up to the last non-default parameter that is not preceded by a default parameter. When this option is enabled, assume that the compiled code does not rely on this `.length` property.
```

PR #3220, opened 2026-07-02
Normalize url

JLHwung:
In this PR we normalize URL to the slashless form, this helps agolia crawling the page and build the correct index.

We also add the last two versions to the blog rewrite.


JLHwung:
I will open a new PR for the Babel v8 REPL changes.

PR #3223, opened 2026-07-03
fix: update algolia docsearch index

JLHwung:


JLHwung:
The v6 index now works as expected:

https://deploy-preview-3223--babel-old.netlify.app/

image

liuxingbaoyu (approved):

PR #3234, opened 2026-07-31
Revert "Migrate to css cascade layers"

JLHwung:
Reverts babel/website#3229

Currently the CSS-in-JS solution `emotion` does not support `@layer` natively, see https://github.com/emotion-js/emotion/issues/3134 This solution sees wide usage in the REPL page, so the there are unexpected style changes introduced in #3229.

I suggest we migrate from `emotion` to CSS modules, which just works with `@layer` as they are handled both CSS preprocessors. Once we agree on my suggestion, I can work on migration to CSS modules and reland #3229 in the future.


liuxingbaoyu (approved):
CSS modules sounds good to me.
We're already using it. :)
https://github.com/babel/website/blob/main/website/src/components/ast/ASTViewer.module.css

JLHwung:
> CSS modules sounds good to me. We're already using it. :) https://github.com/babel/website/blob/main/website/src/components/ast/ASTViewer.module.css

Yes, that's one of the reasons I suggest migration but I forgot to mention it.

PR #3239, opened 2026-08-07
Add docs for bytes and text modules plugins

nicolo-ribaudo:
Docs for https://github.com/babel/babel/pull/17817

### DIFF
--- a/docs/plugin-proposal-bytes-modules.md
+++ b/docs/plugin-proposal-bytes-modules.md
@@ -1,0 +1,113 @@
+---
+id: babel-plugin-proposal-bytes-modules
+title: "@babel/plugin-proposal-bytes-modules"
+sidebar_label: bytes-modules
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+Transforms `import ... with { type: "bytes" }` declarations to platform-specific API to read the imported file as an immutable `Uint8Array`.
+
+:::caution
+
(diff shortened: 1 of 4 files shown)


docs/plugin-proposal-bytes-modules.md
```diff
@@ -0,0 +1,113 @@
...
+title: "@babel/plugin-proposal-bytes-modules"
+sidebar_label: bytes-modules
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+Transforms `import ... with { type: "bytes" }` declarations to platform-specific API to read the imported file as an immutable `Uint8Array`.
+
+:::caution
```
JLHwung, line 12: Nit: There are quite a few caution tip in this docs: https://deploy-preview-3239--babel-v8.netlify.app/docs/babel-plugin-proposal-bytes-modules

I think it is a bit overloaded, Can we group them into a list within one caution tip?

docs/plugin-proposal-bytes-modules.md
```diff
@@ -0,0 +1,113 @@
...
+
+The transformation applied by this plugin depends on your top-level [`targets`](./options.md#targets) to detect whether the generated code should be compatible with Node.js, browsers, or both. When targeting Node.js, the generated code will also change depending on whether you are compiling modules to CommonJS or not.
+
+:::caution
+This plugin cannot be used when compiling modules to AMD, SystemJS, or UMD.
+:::
+
+:::caution
+This plugin only transforms import declarations and not dynamic `import()` calls.
+:::
```
JLHwung, line 26: IMO these two can be combined into something like "The plugin only supports CommonJS and transforms static import declarations". We can fairly assume people using AMD/SystemJS/UMD know what is CommonJS.

PR #3244, opened 2026-09-13
Remove `deprecatedImportAssert` plugin

fisker:
It has been removed in v8. https://babel.dev/docs/v8-migration#babel-parser


nicolo-ribaudo (approved):
Thanks :)

PR #3246, opened 2026-09-15
fix "Require a bigint as the only argument of t.bigIntLiteral" section of Babel V8 API migration docs

samualtnorman:
currently looks like this
image


nicolo-ribaudo:
Thanks :)

